PR #9, opened 2017-07-11
without <scala.version>

wolfstudy:
be short of <scala.version>2.11</scala.version> on pom.xml


ilovesoup:
spark drops the support for scala 2.10 and it'd be save to support only 2.11 for now.

PR #36, opened 2017-08-15
Implementate Spark SQL JDBC and SQL scripting SQL interface

ilovesoup:
Make spark-sql and thrift server working
This is indeed a hack. Only works for Spark 2.11 and not compatible even for 2.10.
Need to create profile at least make it compatible for 2.10 for now.

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/tispark/36)

### DIFF
--- a/src/main/scala/org/apache/spark/sql/TiSparkSession.scala
+++ b/src/main/scala/org/apache/spark/sql/TiSparkSession.scala
@@ -17,26 +17,26 @@ package org.apache.spark.sql
 
 import java.util.concurrent.atomic.AtomicReference
 
-import org.apache.spark.{SparkConf, SparkContext}
 import org.apache.spark.annotation.InterfaceStability
 import org.apache.spark.internal.Logging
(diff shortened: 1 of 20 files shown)


src/main/scala/org/apache/spark/sql/TiSparkSession.scala
```diff
@@ -0,0 +1,38 @@
+package org.apache.spark.sql
```
shenli: miss license

src/main/scala/org/apache/spark/sql/hive/thriftserver/TiSparkSQLDriver.scala
```diff
@@ -0,0 +1,99 @@
+/*
```
shenli: Should we use PingCAP's license?

ilovesoup: This is almost a copy of original spark with minor changes.
I suppose we keep original copyright?

PR #38, opened 2017-09-05
Some minor styling cleanup and simplification

liancheng:
This PR makes some minor cleanup and simplifications to `TiStrategy.scala` to make the code more readable and idiomatic.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/TiStrategy.scala
+++ b/src/main/scala/org/apache/spark/sql/TiStrategy.scala
@@ -15,7 +15,6 @@
 
 package org.apache.spark.sql
 
-
 import com.pingcap.tikv.expression.{TiByItem, TiColumnRef, TiExpr}
 import com.pingcap.tikv.meta.{TiIndexInfo, TiSelectRequest}
 import com.pingcap.tikv.predicates.ScanBuilder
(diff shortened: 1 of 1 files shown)


src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -74,7 +65,7 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
                                    filters: Seq[Expression],
                                    source: TiDBRelation,
                                    selectRequest: TiSelectRequest = new TiSelectRequest): TiSelectRequest = {
-    val tiFilters = filters.map(expr => expr match { case BasicExpression(expr) => expr })
```
liancheng: `map` calls with a single `case` branch is dangerous since it may throw `MatchError` at runtime if this `case` branch doesn't cover all the cases.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -90,43 +81,35 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
...
-
-            case Max(BasicExpression(arg)) =>
-              selectRequest.addAggregate(new TiMax(arg),
-                fromSparkType(aggExpr.aggregateFunction.dataType))
-
-            case First(BasicExpression(arg), _) =>
-              selectRequest.addAggregate(new TiFirst(arg),
-                fromSparkType(aggExpr.aggregateFunction.dataType))
-
-            case _ => None
```
liancheng: Returning a `None` in a `foreach` closure call looks weird since the closure returns `Unit`. The original `None` is simply discarded.

zhexuany:
Just pulled ur changes and tested it locally. merging it.

PR #45, opened 2017-09-23
Eliminates these repeated computation in multi aggregations query

RayeRen:
When more than one aggregation are requested in one query, tispark will push down some aggregations repeatedly, like `count(number#0L)` in the following example.

```scala
scala> spark.sql ("select count(number),avg(number) from person").explain
== Physical Plan ==
*HashAggregate(keys=[], functions=[sum(count(number#0L)#42L), sum(sum(number#0L)#43L), sum(count(number#0L)#45L)])
+- Exchange SinglePartition
   +- *HashAggregate(keys=[], functions=[partial_sum(count(number#0L)#42L), partial_sum(sum(number#0L)#43L), partial_sum(count(number#0L)#45L)])
      +- Scan CoprocessorRDD[count(number#0L)#42L,sum(number#0L)#43L,count(number#0L)#45L]
```

This PR eliminates these repeated computation:

```scala
scala> spark.sql ("select count(number),avg(number) from person").explain()
== Physical Plan ==
*HashAggregate(keys=[], functions=[sum(count(number#0L)#20L), sum(sum(number#0L)#21L), sum(count(number#0L)#20L)])
+- Exchange SinglePartition
   +- *HashAggregate(keys=[], functions=[partial_sum(count(number#0L)#20L), partial_sum(sum(number#0L)#21L), partial_sum(count(number#0L)#20L)])
      +- Scan CoprocessorRDD[count(number#0L)#20L,sum(number#0L)#21L]
```

### DIFF
--- a/src/main/scala/org/apache/spark/sql/TiStrategy.scala
+++ b/src/main/scala/org/apache/spark/sql/TiStrategy.scala
@@ -180,11 +180,18 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
   // and then we don't return (don't planLater) and plan the remaining all at once
   private def doPlan(source: TiDBRelation, plan: LogicalPlan): Seq[SparkPlan] = {
 
-    val aliasMap = mutable.HashMap[Expression, Alias]()
+    val aliasMap = mutable.HashMap[(Boolean,Expression), Alias]()
     val avgPushdownRewriteMap = mutable.HashMap[ExprId, List[AggregateExpression]]()
     val avgFinalRewriteMap = mutable.HashMap[ExprId, List[AggregateExpression]]()
 
-    def toAlias(expr: Expression) = aliasMap.getOrElseUpdate(expr, Alias(expr, expr.toString)())
+    def toAlias(expr: AggregateExpression) =
+      if (!expr.deterministic) {
+        Alias(expr, expr.toString())()
+      } else {
+        aliasMap.getOrElseUpdate(
+          (expr.deterministic, expr.canonicalized), Alias(expr, expr.toString)()
+        )
+      }
 
     def newAggregate(aggFunc: AggregateFunction,
                      originalAggExpr: AggregateExpression) =
@@ -266,7 +273,7 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
           aggExpr =>
             avgPushdownRewriteMap
               .getOrElse(aggExpr.resultId, List(aggExpr))
-        }
+        }.distinct
 
         selReq = aggregationToSelectRequest(groupingExpressions,
           pushdownAggregates,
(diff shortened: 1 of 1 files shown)


src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -180,11 +180,11 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
   // and then we don't return (don't planLater) and plan the remaining all at once
   private def doPlan(source: TiDBRelation, plan: LogicalPlan): Seq[SparkPlan] = {
 
-    val aliasMap = mutable.HashMap[Expression, Alias]()
+    val aliasMap = mutable.HashMap[(String,Attribute), Alias]()
     val avgPushdownRewriteMap = mutable.HashMap[ExprId, List[AggregateExpression]]()
     val avgFinalRewriteMap = mutable.HashMap[ExprId, List[AggregateExpression]]()
 
-    def toAlias(expr: Expression) = aliasMap.getOrElseUpdate(expr, Alias(expr, expr.toString)())
+    def toAlias(expr: AggregateExpression) = aliasMap.getOrElseUpdate((expr.aggregateFunction.prettyName,expr.references.head), Alias(expr, expr.toString)())
```
Novemser: Suppose we execute this query "select count(*) from customer", the AttributeSet of this query will be empty and throw a java.util.NoSuchElementException.

RayeRen:
Change `references.head` to `expr.children.head` so that it shouldn't be null anymore.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -180,11 +180,11 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
   // and then we don't return (don't planLater) and plan the remaining all at once
   private def doPlan(source: TiDBRelation, plan: LogicalPlan): Seq[SparkPlan] = {
 
-    val aliasMap = mutable.HashMap[Expression, Alias]()
+    val aliasMap = mutable.HashMap[(String,Expression), Alias]()
     val avgPushdownRewriteMap = mutable.HashMap[ExprId, List[AggregateExpression]]()
     val avgFinalRewriteMap = mutable.HashMap[ExprId, List[AggregateExpression]]()
 
-    def toAlias(expr: Expression) = aliasMap.getOrElseUpdate(expr, Alias(expr, expr.toString)())
+    def toAlias(expr: AggregateExpression) = aliasMap.getOrElseUpdate((expr.aggregateFunction.prettyName,expr.children.head), Alias(expr, expr.toString)())
```
Novemser: There's a function ```semanticEquals``` defined in [[org.apache.spark.sql.catalyst.expressions.Expression]], which can be used to determine whether two expressions are the same, and it has been broadly used in the Spark framework. I think the logic behind this function can help us with our expression deduplication.

Suppose we execute this query ```select count(c_custkey + 1),count(1 + c_custkey) from customer```, if we use ```expr.children.head``` to distinguish child expressions, we may not regard ```c_custkey + 1``` and ```1 + c_custkey``` as the same(but actually they are deterministic and can be canonicalized to a same expression), resulting in duplicated push downs. But if we switch to ```semanticEquals```'s method, the problem can be solved.

Below is my solution : )
```
    val aliasMap = mutable.HashMap[(Boolean, Expression), Alias]()

    def toAlias(expr: AggregateExpression) = {
      if (!expr.deterministic) {
        Alias(expr, expr.toString())()
      } else {
        aliasMap.getOrElseUpdate(
          (expr.deterministic, expr.canonicalized), Alias(expr, expr.toString)()
        )
      }
    }
```

zhexuany:
@RayeRen friendly ping. any updates?c

RayeRen:
@zhexuany  Yeah, I accepted @Novemser 's suggestion. Thanks a lot!

zhexuany:
@ilovesoup PTAL.

PR #47, opened 2017-10-09
[update]Modify integration test wrapper

Novemser:
Minor update for undergoing integration test correctly.


integtest/README.md
```diff
@@ -38,12 +39,12 @@ test.basepath=/Users/whoever/workspace/pingcap/tispark/integtest/testcases
 
 | Key           | Desc          |
 | ------------- |:-------------:|
-| pd.addrs        | Placement Driver Address separated by "," |
+| spark.tispark.pd.addresses | PD Cluster Addresses, split by comma(Used for spark) |
```
ilovesoup: Would you please check if it is the pd address in config.properties effective or the one in spark-default.conf effective?
ticontext will only read spark's conf settings instead of config.properties. I guess spark.tispark.pd.addresses=127.0.0.1:2379 here is useless.
If so, please modify the readme.

PR #53, opened 2017-10-11
rename and prepare for index logic merge

ilovesoup:


ilovesoup:
@birdstorm @Novemser @zhexuany PTAL

zhexuany:
@birdstorm PTAL.

PR #57, opened 2017-10-24
fix inconsistent with tikv-lib-java

zhexuany:
This change is [Reviewable](https://reviewable.io/reviews/pingcap/tispark/57)

### DIFF
--- a/src/main/scala/com/pingcap/tispark/TiRDD.scala
+++ b/src/main/scala/com/pingcap/tispark/TiRDD.scala
@@ -15,6 +15,8 @@
 
 package com.pingcap.tispark
 
+import java.util
+
 import com.pingcap.tikv._
 import com.pingcap.tikv.meta.{TiSelectRequest, TiTimestamp}
 import com.pingcap.tikv.operation.SchemaInfer
@@ -38,7 +40,7 @@ class TiRDD(val selectReq: TiSelectRequest,
   type TiRow = com.pingcap.tikv.row.Row
 
   @transient lazy val session: TiSession = TiSession.create(tiConf)
-  @transient lazy val (fieldsType: List[DataType], rowTransformer: RowTransformer) = initializeSchema
+  @transient lazy val (fieldsType: List[DataType], rowTransformer: RowTransformer) = initializeSchema()
   @transient lazy val snapshot: Snapshot = session.createSnapshot(ts)
 
   def initializeSchema(): (List[DataType], RowTransformer) = {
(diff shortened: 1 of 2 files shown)


src/main/scala/com/pingcap/tispark/TiRDD.scala
```diff
@@ -48,11 +50,11 @@ class TiRDD(val selectReq: TiSelectRequest,
...
 
   override def compute(split: Partition, context: TaskContext): Iterator[Row] = new Iterator[Row] {
-    selectReq.bind
+    selectReq.resolve()
     // bypass, sum return a long type
-    val tiPartition = split.asInstanceOf[TiPartition]
-    val iterator = snapshot.select(selectReq, split.asInstanceOf[TiPartition].task)
-    val finalTypes = rowTransformer.getTypes.toList
+    val tiPartition: TiPartition = split.asInstanceOf[TiPartition]
+    val iterator: util.Iterator[TiRow] = snapshot.tableRead(selectReq, split.asInstanceOf[TiPartition].task)
```
birdstorm: I don't think `Iterator` here stands for that under `java.util` package...

src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -57,7 +57,7 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
     selectRequest.setTableInfo(table)
     // In case count(*) and nothing pushed, pick the first column
     if (selectRequest.getFields.size == 0) {
-      selectRequest.addField(TiColumnRef.create(table.getColumns.get(0).getName))
+      selectRequest.addRequiredColumn(TiColumnRef.create(table.getColumns.get(0).getName))
```
ilovesoup: this change is not merged to master yet. its not a inconsistency

src/main/scala/com/pingcap/tispark/TiRDD.scala
```diff
@@ -48,11 +50,11 @@ class TiRDD(val selectReq: TiSelectRequest,
   }
 
   override def compute(split: Partition, context: TaskContext): Iterator[Row] = new Iterator[Row] {
-    selectReq.bind
+    selectReq.resolve()
```
ilovesoup: this is not merged to master yet. not a inconsistency

src/main/scala/com/pingcap/tispark/TiRDD.scala
zhexuany: It actually is. You can check signature of tableRead. It returns a `SelectIterator` which implements `util.Iterator`.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
zhexuany: Yeah. You are right. Just wait `index_read` merging into master. If there are still some inconsistency, it should be address. Otherwise, we can just close this PR.

ilovesoup:
Is this still problem? I will close this PR if not.

PR #60, opened 2017-10-27
add more conf items

zhexuany:


ilovesoup:
plz pr against index read branch since they are to be merged in pair.

PR #65, opened 2017-11-01
Add R language support

Novemser:
Add a shabby thin layer to support `TiSpark` in R language.

### DIFF
--- a/TiSparkR/DESCRIPTION
+++ b/TiSparkR/DESCRIPTION
@@ -2,9 +2,10 @@ Package: TiSparkR
 Type: Package
 Title: TiSpark for R
 Version: 0.1.0
-Author: Novemser
-Maintainer: The package maintainer <novemser@gmail.com>
-Description: Add a shabby thin layer to support TiSpark in R language.
+Author: PingCAP
+Maintainer: Novemser <novemser@gmail.com>
+Description: A shabby thin layer to support TiSpark in R language.
 License: Apache 2.0
+Copyright: 2017 PingCAP, Inc.
 Encoding: UTF-8
 LazyData: true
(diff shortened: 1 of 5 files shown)


TiSparkR/R/tisparkR.R
```diff
@@ -0,0 +1,13 @@
+# Title     : TiSparkR
```
ilovesoup: Please add copyright info

TiSparkR/Readme.md
```diff
@@ -0,0 +1,42 @@
+## TiSparkR
```
ilovesoup: Looks like merging to root/README.md is better

TiSparkR/R/tisparkR.R
```diff
@@ -0,0 +1,13 @@
...
+# Objective : TiSpark entry for R
+# Created by: novemser
+# Created on: 17-11-1
+
+createTiContext <- function(session) {
```
ilovesoup: how error handling behaves here in R?

TiSparkR/DESCRIPTION
```diff
@@ -0,0 +1,10 @@
+Package: TiSparkR
+Type: Package
+Title: TiSpark for R
+Version: 0.1.0
+Author: Novemser
+Maintainer: The package maintainer <novemser@gmail.com>
+Description: Add a shabby thin layer to support TiSpark in R language.
+License: Apache 2.0
+Encoding: UTF-8
+LazyData: true
```
ilovesoup: what is lazydata?

Novemser: If the `DESCRIPTION` contains `LazyData: true`, then datasets in R will be lazily loaded. This means that they won’t occupy any memory until we use them. It is recommended that we always include `LazyData: true` in the DESCRIPTION.

TiSparkR/R/tisparkR.R
Novemser: Done.

PR #67, opened 2017-11-02
integration test

birdstorm:
this pr tries to merge integration test into master


integtest/src/main/scala/com/pingcap/spark/JDBCWrapper.scala
```diff
@@ -113,45 +119,44 @@ class JDBCWrapper(prop: Properties) extends LazyLogging {
...
 
-  private def insertRow(row: List[Any], schema: List[String], table: String): Unit = {
+  private def insertRow(row: List[Any], schema: List[String], table: String) = {
     logger.info("Insert into : " + table)
     val placeholders = List.fill(row.size)("?").mkString(",")
-    val stat = s"insert into ${table} values (${placeholders})"
+    val stat = s"insert into $table values ($placeholders)"
     val ps = connection.prepareStatement(stat)
-    row.zipWithIndex.map { case (value, index) => {
+    row.zipWithIndex.map { case (value, index) =>
```
ilovesoup: maybe change it to foreach since we not really using return value.

birdstorm:
All done. PTAL @ilovesoup thx~

PR #69, opened 2017-11-03
Fix Time type

birdstorm:
sync with same titled branch in tikv client


birdstorm:
Reminder: this version of tikv client need changes after its corresponding branch is merged because we cannot find a correct submodule beforehand.

birdstorm:
closing this pr because changes to adapt corresponding branch are not needed.

PR #72, opened 2017-11-09
add Type tests.

birdstorm:
Amongst 28 types tested, binary types (`blob`, `text`) are not comparable, and `enum` and `set` are not supported.


birdstorm:
@ilovesoup @zhexuany PTAL

PR #83, opened 2017-11-16
[Fix]Don't let bit type be pushed down

Novemser:
TiKV doesn't allow Bit type operation to be pushed down, so we skip this type in TiSpark.


src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -200,10 +203,46 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
         originalAggExpr.isDistinct,
         NamedExpression.newExprId)
 
+    /**
+      * Is expression allowed to be pushed down
+      * @param expression the expression to examine
+      * @return whether expression can be pushed down
+      */
+    def allowExpPushDown(expression: Expression): Boolean = {
```
ilovesoup: We have test utils called isSupportedFilter
Please move it there.

PR #86, opened 2017-11-20
fix pom

birdstorm:
delete duplicate dependencies

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,18 +88,6 @@
             <version>${junit.version}</version>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.powermock</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -94,12 +94,6 @@
             <version>1.6.6</version>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.powermock</groupId>
-            <artifactId>powermock-module-junit4</artifactId>
-            <version>1.6.6</version>
-            <scope>test</scope>
-        </dependency>
         <dependency>
```
ilovesoup: I believe we don't use powermock-api-mockito either. If to clean things up, please remove that one either.

birdstorm: done. thx for pointing out.

PR #87, opened 2017-11-21
[Fix]No Binding NPE

Novemser:
Originally in `isSupportedBasicExpression`, some expression had not been resolved to a table correctly, causing a NPE problem which prevent the expression from push down. Fixed this by adding a `resolve` to all the expression.


Novemser:
@ilovesoup PTAL

PR #92, opened 2017-11-23
fix integration test framework load error

birdstorm:
cherry-pick integration test framework fix to master


birdstorm:
tikv-client has conflicts with this pr, PTAL @ilovesoup @zhexuany

Issue #98, opened 2017-11-29
(equation/inequation with null) in where clause fails


birdstorm:
Problem occurs when we have some equation concerning null types

e.g.,
```
select x from a where x = null;
```
where `x` is of any type

Following exceptions are thrown:
```
ResultStage 84 (collect at SparkWrapper.scala:53) failed in 0.022 s due to Job aborted due to stage failure: Task 0 in stage 84.0 failed 1 times, most recent failure: Lost task 0.0 in stage 84.0 (TID 84, localhost, executor driver): com.pingcap.tikv.expression.TiExpressionException: NULL constant has no type
	at com.pingcap.tikv.expression.TiConstant.getType(TiConstant.java:90)
	at com.pingcap.tikv.expression.scalar.And.validateArguments(And.java:52)
	at com.pingcap.tikv.expression.TiFunctionExpression.<init>(TiFunctionExpression.java:34)
	at com.pingcap.tikv.expression.TiBinaryFunctionExpression.<init>(TiBinaryFunctionExpression.java:23)
	at com.pingcap.tikv.expression.scalar.And.<init>(And.java:28)
	at com.pingcap.tikv.predicates.PredicateUtils.mergeCNFExpressions(PredicateUtils.java:34)
	at com.pingcap.tikv.meta.TiSelectRequest.buildTableScan(TiSelectRequest.java:160)
	at com.pingcap.tikv.meta.TiSelectRequest.buildScan(TiSelectRequest.java:90)
	at com.pingcap.tikv.operation.SelectIterator.getRowIterator(SelectIterator.java:61)
	at com.pingcap.tikv.Snapshot.tableRead(Snapshot.java:115)
	at org.apache.spark.sql.tispark.TiRDD$$anon$2.<init>(TiRDD.scala:61)
	at org.apache.spark.sql.tispark.TiRDD.compute(TiRDD.scala:53)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:323)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:287)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:38)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:323)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:287)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:38)
[...]

PR #101, opened 2017-11-30
[Trivial]Don't let null constant push down

Novemser:
`TiConstant` don't support `null` type currently, we don't allow null constant push down as a trivial fix before a satisfactory solution.


birdstorm:
should fix #98

Novemser:
Since any SQL condition match `col != null` or `col == null` will cause the problem, label it as `p0`.

Novemser:
PTAL @ilovesoup

ilovesoup:
@Novemser Please resolve conflict.

ilovesoup:
@Novemser please resolve conflict.

Issue #85, opened 2017-11-17
Add version support on TiContext.version() or add udf tispark_version()


ilovesoup:
Both Client and TiSpark should be covered independently before we merge them into one project.
For now troubleshooting is a headache. We might as well make it a auto-build with different version artifact. But due to our development pace, it's hard to catch up.

PR #104, opened 2017-12-01
Add Print version

ilovesoup:
Version has two parts: TiSpark version and TiKV Client version.
Both are represented by git describe --tags
It is tag + commit ahead of tag - current SHA

Resolve https://github.com/pingcap/tispark/issues/85

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/tispark/104)

### DIFF
--- a/scripts/version.sh
+++ b/scripts/version.sh
@@ -1,4 +1,19 @@
 #!/bin/sh
+#
+#   Copyright 2017 PingCAP, Inc.
+#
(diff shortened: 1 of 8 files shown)


ilovesoup:
PTAL @zhexuany @Novemser @birdstorm

scripts/version.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
```
Novemser: Copyright

PR #110, opened 2017-12-04
Top n push down

Novemser:
Support for `Limit` and `order by` clause.


Novemser:
Refer to client PR: [#178](https://github.com/pingcap/tikv-client-lib-java/pull/178)

src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -146,6 +147,70 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
...
+    // If sortOrder is not null, limit must be greater than 0
+    if (limit < 0 || (sortOrder == null && limit == 0)) {
+      return Nil
+    }
+
+    child match {
+      case PhysicalOperation(projectList, filters, LogicalRelation(source: TiDBRelation, _, _))
+          if filters.forall(TiUtils.isSupportedFilter(_, source)) =>
+        pruneTopNFilterProject(limit, projectList, filters, source) :: Nil
+      case TiAggregation(
```
ilovesoup: Would you please confirm if topN and aggregates can be pushdown independently. 
For example, if you have group by name and count
you have 
region 1: A, 100, B, 1
region 2: A, 100, B 1000
and you take Top 1. In above case, B will be ignored in each region request but actually B is the group to take.
I wonder if above case works in your implementation.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -225,8 +266,126 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
...
+
+    aggregate.AggUtils.planAggregateWithoutDistinct(
+      groupingExpressions,
+      residualAggregateExpressions,
+      rewrittenResultExpression,
+      toCoprocessorRDD(source, output, dagReq)
+    )
+  }
+
+  def validGroupAggregate(groupingExpressions: Seq[NamedExpression],
```
ilovesoup: Change it to isValidAggregates. Aggregates include aggexpr and groupby exprs.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -248,89 +407,15 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
...
-        val output = (pushdownAggregates.map(x => toAlias(x)) ++ groupingExpressions)
-          .map(_.toAttribute)
-
-        val projectSeq: Seq[Attribute] = projects.asInstanceOf[Seq[Attribute]]
-        projectSeq.foreach(attr => dagReq.addRequiredColumn(TiColumnRef.create(attr.name)))
-
-        aggregate.AggUtils.planAggregateWithoutDistinct(
+          ) if validGroupAggregate(groupingExpressions, aggregateExpressions, source) =>
+        val dagReq: TiDAGRequest = filterToDAGRequest(filters, source)
+        groupAggregateProjection(
```
ilovesoup: same here

src/main/scala/org/apache/spark/sql/TiStrategy.scala
Novemser: Changed.

src/main/scala/org/apache/spark/sql/TiStrategy.scala

Novemser:
@ilovesoup Sorry for the inconvenience I bought, this PR only handles TopN logic but not aggregation with TopN logic.

Novemser:
Integration test result:
Tests run: 7211  Tests succeeded: 3735  Tests failed: 21  Tests skipped: 3455
Result is as expected.

PR #114, opened 2017-12-06
Fix cast AIOB error

Novemser:
Related PR and issue in kv-client: [#183](https://github.com/pingcap/tikv-client-lib-java/pull/183)

Change master ref.


Novemser:
Useless since we will adapt to pom parent child project mode.

PR #120, opened 2017-12-08
Fix integration scripts logic

birdstorm:
* Delete all template files
* Leave key information inside properties file
* Rename test log output so ./test_all.sh won't clear previous files
* Change base comparison for TPCH tests to Spark JDBC since TiDB may fail for large table joins
* Modify README (in progress)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,4 +47,3 @@ hs_err_pid*
 metastore_db/
 tpch
 .java-version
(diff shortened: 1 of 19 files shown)


Novemser:
@birdstorm Maybe we should remove 'tp_bit' type test cases since they produce too many failed tests, but these failed cases only has error in decoding ASCII code 49 to `1`.

PR #121, opened 2017-12-08
Fix accumulator register twice bug

Novemser:
Fix accumulator register twice bug.


Novemser:
Refer to https://github.com/pingcap/tispark/pull/123

PR #125, opened 2017-12-11
don't push down datetime and timestamp

### LINKED ISSUE #124: Inconsistent result when comparing DateTime with Timestamp
Known affected branch:
```
master
```

Schema:
```sql
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                      |
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| tb_n  | CREATE TABLE `tb_n` (
  `dt` datetime DEFAULT CURRENT_TIMESTAMP,
  `ts` timestamp DEFAULT CURRENT_TIMESTAMP
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin |
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

Data:
```sql
+---------------------+---------------------+
| dt                  | ts                  |
+---------------------+---------------------+
| 2017-12-11 11:38:36 | 2017-12-11 11:38:36 |
+---------------------+---------------------+
1 row in set (0.01 sec)

```

SQL:
```
select * from tb_n where dt = ts
```

TiSpark:
Nothing selected.

Spark with JDBC:
TiDB:
```sql
+---------------------+---------------------+
| dt                  | ts                  |
+---------------------+---------------------+
| 2017-12-11 11:38:36 | 2017-12-11 11:38:36 |
+---------------------+---------------------+
```

TiSpark plan:
```bash
== Physical Plan ==
TiDB CoprocessorRDD{[table: tb_n] , Ranges: Start:[-9223372036854775808], End: [9223372036854775807], Columns: [dt], [ts], Filter: Not(IsNull([dt])), Not(IsNull([ts])), Equal([dt], [ts])}
```

Seems we pushed down `Equal([dt], [ts])` and TiKV filtered all rows.

birdstorm:
fix #124 

When we push datetime and timestamp down to coprocessor, it will be confusing since they have different timezones in storage. One of the best solution is not to push them down at all, just like what spark does.

### DIFF
--- a/src/main/scala/com/pingcap/tispark/TiUtils.scala
+++ b/src/main/scala/com/pingcap/tispark/TiUtils.scala
@@ -22,10 +22,10 @@ import com.pingcap.tikv.expression.{ExpressionBlacklist, TiExpr}
 import com.pingcap.tikv.kvproto.Kvrpcpb.{CommandPri, IsolationLevel}
 import com.pingcap.tikv.meta.{TiColumnInfo, TiTableInfo}
 import com.pingcap.tikv.types._
-import org.apache.spark.sql.{SparkSession, TiStrategy}
 import org.apache.spark.sql.catalyst.expressions.aggregate._
 import org.apache.spark.sql.catalyst.expressions.{AttributeReference, Expression, Literal, NamedExpression}
 import org.apache.spark.sql.types.{DataType, DataTypes, MetadataBuilder, StructField, StructType}
+import org.apache.spark.sql.{SparkSession, TiStrategy}
 import org.apache.spark.{SparkConf, sql}
 
 import scala.collection.JavaConversions._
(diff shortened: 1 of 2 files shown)


src/main/scala/com/pingcap/tispark/TiUtils.scala
```diff
@@ -80,7 +80,10 @@ object TiUtils {
         // bit/duration type is not allowed to be pushed down
         case attr: AttributeReference if nameTypeMap.contains(attr.name) =>
           val head = nameTypeMap.get(attr.name).head
-          return !head.isInstanceOf[BitType] && head.getTypeCode != Types.TYPE_DURATION
+          return !head.isInstanceOf[BitType] &&
+            head.getTypeCode != Types.TYPE_DURATION &&
+            head.getTypeCode != Types.TYPE_TIMESTAMP &&
```
ilovesoup: This means TS and Datetime will not be pushed? Why is that?

birdstorm: because afaik, spark does not push down timestamp and datetime in calculation.

birdstorm:
close this pr since we now need to push down timestamp, or it will be too slow.

PR #127, opened 2017-12-11
Switch integration test to ScalaTest framework

Novemser:
Migrant original integration test to ScalaTest framework.


Novemser:
PTAL @ilovesoup @birdstorm

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "core/src/test/resources/tispark-test"]
+	path = core/src/test/resources/tispark-test
+	url = https://github.com/Novemser/tispark_test_data.git
```
Novemser: Test data are managed in a separate repository.

README.md
```diff
@@ -84,15 +84,16 @@ Read the [Quick Start](./docs/userguide.md).
 ## How to build
 To build all TiSpark modules :
 ```
-mvn clean package
+mvn clean install -Dmaven.test.skip=true
```
Novemser: use arg `-Dmaven.test.skip=true` to skip unit test and integration tests.

core/src/test/scala/org/apache/spark/sql/expression/Aggregate0Suite.scala
```diff
@@ -0,0 +1,161 @@
...
+
+package org.apache.spark.sql.expression
+
+import org.apache.spark.sql.BaseTiSparkSuite
+import org.apache.spark.sql.test.SharedSQLContext
+
+class Aggregate0Suite extends BaseTiSparkSuite with SharedSQLContext {
+
+  test("select tp_double from full_data_type_table  group by (tp_double)  order by tp_double ") {
+    runTest(
```
birdstorm: can we wrap runTest() into a single function, e.g., runTestAll() ?

Novemser: Something like https://github.com/pingcap/tispark/pull/127/files#diff-604c956654a282408905e45528650602 ArithmeticTest2Suite.scala ?

birdstorm: yeah, we can wrap `runTest(query, query.replace("full_data_type_table", "full_data_type_table_j"))` into a single function.
Maybe that would be too much work? LOL

.gitmodules
birdstorm: Brilliant! So we are not concerning Chinese characters any more, are we?

core/src/test/scala/org/apache/spark/sql/test/TestConstants.scala
```diff
@@ -0,0 +1,31 @@
...
+package org.apache.spark.sql.test
+
+object TestConstants {
+  val KeyTiDBAddress = "tidb.addr"
+  val KeyTiDBPort = "tidb.port"
+  val KeyTiDBUser = "tidb.user"
+  val KeyTestDB = "test.db"
+  val KeyTPCHDB = "tpch.db"
+  val KeyUseRawSparkMySql = "spark.use_raw_mysql"
+  val KeyMysqlAddress = "mysql.addr"
```
birdstorm: It was my fault that in fact these config attributes with leading 'mysql' are used for possible future tests against raw mysql. So they are not used this moment, maybe we can add a comment here just in case?

birdstorm: Or we can simply delete their usage if we are sure they won't be used anyhow.

core/src/test/Readme.md
```diff
@@ -0,0 +1,24 @@
+## Integration Test
+### Usage
+1. Prepare your Spark and TiDB cluster environment, it is recommended to use our [docker-compose](../../../docker-compose.yaml) script to deploy a TiDB cluster locally, or you may prefer other ways to set up an appropriate test environment.
+2. You can modify test configuration placed in [tidb_config.properties.template](./resources/tidb_config.properties.template), copy this template as tidb_config.properties and config this file according to your environment.
+3. Run `mvn clean install` to run all tests.
+
+### Configuration
+You can change your test configuration in `tidb_config.properties`
+```bash
+// TiDB address
```
birdstorm: Just in case users simply copy and paste these lines, we may better use '#' instead of '//' here.

core/src/test/scala/org/apache/spark/sql/QueryTest.scala
```diff
@@ -0,0 +1,352 @@
...
+    require(
+      actualAnswer.length == expectedAnswer.length,
+      s"actual answer length ${actualAnswer.length} != " +
+        s"expected answer length ${expectedAnswer.length}"
+    )
+
+    // TODO: support other numeric types besides Double
+    // TODO: support struct types?
+    actualAnswer.toSeq.zip(expectedAnswer.toSeq).foreach {
+      case (actual: Double, expected: Double) =>
```
birdstorm: is it possible that we get different types for `actual` and `expected`? e.g., `Double` and `BigDecimal`?

.gitmodules
Novemser: No, we don't need to consider Chinese characters ordering in `blob` representation any more. :laughing:

core/src/test/scala/org/apache/spark/sql/expression/Aggregate0Suite.scala
Novemser: Emmmm some work will do.....these generated code could be improved.

core/src/test/scala/org/apache/spark/sql/test/TestConstants.scala
Novemser: Agree.

core/src/test/Readme.md
Novemser: Nice catch.

core/src/test/scala/org/apache/spark/sql/QueryTest.scala
Novemser: My fault, this function has not been used, a complete refactor is needed....

Novemser: But we might use this later in Spark plan tests.

core/src/test/scala/org/apache/spark/sql/expression/Aggregate0Suite.scala
Novemser: Done.

core/src/test/scala/org/apache/spark/sql/test/TestConstants.scala
Novemser: Done.

core/src/test/Readme.md
Novemser: Done.

PR #136, opened 2017-12-13
Add debug utils

ilovesoup:
For now it print region distribution for a specific table.

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/tispark/136)


src/main/scala/org/apache/spark/sql/TiContext.scala
```diff
@@ -15,22 +15,31 @@
...
 import org.apache.spark.internal.Logging
+import scala.collection.JavaConverters._
 
 class TiContext(val session: SparkSession) extends Serializable with Logging {
   val sqlContext: SQLContext = session.sqlContext
   val conf: SparkConf = session.sparkContext.conf
   val tiConf: TiConfiguration = TiUtils.sparkConfToTiConf(conf)
   val tiSession: TiSession = TiSession.create(tiConf)
   val meta: MetaManager = new MetaManager(tiSession.getCatalog)
+  val debug: DebugTool = new DebugTool
```
Novemser: Lazy initialization may be better here since user may call `TiContext.debug` several times.

src/main/scala/org/apache/spark/sql/TiContext.scala
```diff
@@ -15,22 +15,31 @@
...
 
   TiUtils.sessionInitialize(session)
 
   final val version: String = TiSparkVersion.version
 
+  class DebugTool {
+    def getRegionDistribution(dbName: String, tableName: String): Map[String, Integer] = {
```
Novemser: Maybe some brief comment here will be better.

src/main/scala/org/apache/spark/sql/TiContext.scala
```diff
@@ -15,21 +15,35 @@
...
+import scala.collection.JavaConverters._
 
 class TiContext(val session: SparkSession) extends Serializable with Logging {
   val sqlContext: SQLContext = session.sqlContext
   val conf: SparkConf = session.sparkContext.conf
   val tiConf: TiConfiguration = TiUtils.sparkConfToTiConf(conf)
   val tiSession: TiSession = TiSession.create(tiConf)
   val meta: MetaManager = new MetaManager(tiSession.getCatalog)
+
+  val debug: DebugTool = new DebugTool
```
Novemser: Lazy initialization may be better here since user may call `TiContext.debug` several times.

src/main/scala/org/apache/spark/sql/TiContext.scala
```diff
@@ -15,21 +15,35 @@
...
+
   TiUtils.sessionInitialize(session, tiSession)
 
+
   final val version: String = TiSparkVersion.version
 
+  class DebugTool {
+    def getRegionDistribution(dbName: String, tableName: String): Map[String, Integer] = {
```
Novemser: Maybe adding some brief comment on `getRegionDistribution` here will be clearer for user to use.

PR #137, opened 2017-12-14
hide bit type in tests, fix potential compare error for double type

birdstorm:


integtest/src/main/scala/com/pingcap/spark/TestCase.scala
```diff
@@ -629,9 +632,11 @@ class TestCase(val prop: Properties) extends LazyLogging {
       new SimpleDateFormat("yy-MM-dd HH:mm:ss").format(value)
     }
 
-    def compValue(lhs: Any, rhs: Any): Boolean = rhs match {
+    def compValue(lhs: Any, rhs: Any): Boolean = lhs match {
```
Novemser: Doing address and `toString` equality comparison at first shall avoid redundant `toString` logic below.
```scala
def compValue(lhs: Any, rhs: Any): Boolean =
  if (lhs == rhs || lhs.toString.eq(rhs.toString)) {
    true
  } else rhs match {
    case _: Double | _: Float | _: BigDecimal | _: java.math.BigDecimal =>
      val l = toDouble(lhs)
      val r = toDouble(rhs)
      Math.abs(l - r) < eps || Math.abs(r) > eps && Math.abs((l - r) / r) < eps
    case _: Number | _: BigInt | _: java.math.BigInteger =>
      toInteger(lhs) == toInteger(rhs)
    case _: Timestamp =>
      toString(lhs) == toString(rhs)
    case _ => false
  }
```

PR #141, opened 2017-12-18
Add test framework for each issue

birdstorm:
A test framework consisting of integration tests for each fixed issue so that we can locate the problem faster when refactoring.

### DIFF
--- a/integtest/src/main/scala/com/pingcap/spark/IssueTestCase.scala
+++ b/integtest/src/main/scala/com/pingcap/spark/IssueTestCase.scala
@@ -1,0 +1,52 @@
+package com.pingcap.spark
+
+import java.util.Properties
+
+import scala.collection.mutable.ArrayBuffer
+
+/**
+  * Created by birdstorm on 2017/12/15.
+  */
+class IssueTestCase(prop: Properties) extends TestCase(prop) {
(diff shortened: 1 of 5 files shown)


integtest/src/main/scala/com/pingcap/spark/IssueTestCase.scala
```diff
@@ -0,0 +1,55 @@
...
+  /**
+    * An example of Tests for issues
+    * Use "Test"+"Client"/"Spark"+<IssueNumber> for test name.
+    * Common test includes building test table from TiDB
+    * and retrieving result from TiSpark.
+    * Remember to call refresh() after test data is built or renewed.
+    *
+    * @param dbName Name of database, the default is issue_test
+    */
+  private def TestClient0198(dbName: String): Unit = {
```
ilovesoup: What does 0198 means?

Novemser: Issue No.198 https://github.com/pingcap/tikv-client-lib-java/issues/198

birdstorm: as described in comment above

PR #143, opened 2017-12-18
Optimize aggregation push down column logic

Novemser:
Relevant to https://github.com/pingcap/tikv-client-lib-java/pull/203

TPCH:
All passed.

DAG:
Result: Tests run: 6744  Tests succeeded: 3397  Tests failed: 16  Tests skipped: 3331

PR #153, opened 2017-12-22
Simplify aggregation push-down

liancheng:
This PR tries to simplify/refactor the aggregation push-down code path in `TiStrategy` to improve readability and maintainability by:

- [x] Use more Scala idiomatic coding style 

  In some cases, it may introduce extra expression traversals. It shouldn't be an issue since the number of expressions is quite limited but feel free to disagree.

- [x] Unify `Average` push-down

  Instead of special-casing `Average` in `groupAggregateProjection`, we may want to convert `Average`s into `Count`s and `Sum`s earlier in the form of Catalyst expressions so that we can remove the clumsy code paths involving `avgPushdownRewriteMap`, `avgFinalRewriteMap`, and building partial/final aggregate expressions.

Opening this PR early for discussion and to make sure it makes Travis happy. If people would like to split this work into a few separate PRs for easier code review and future Git blaming, I'd be happy to do that.

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/tispark/153)


src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -271,59 +271,51 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
...
     val avgFinalRewriteMap = mutable.HashMap[ExprId, List[AggregateExpression]]()
 
-    def newAggregate(aggFunc: AggregateFunction, originalAggExpr: AggregateExpression) =
-      AggregateExpression(
-        aggFunc,
-        originalAggExpr.mode,
-        originalAggExpr.isDistinct,
-        originalAggExpr.resultId
-      )
-
```
liancheng: This local function is replaced by inlined `.copy()` calls.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -271,59 +271,51 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
     source: TiDBRelation,
     dagReq: TiDAGRequest
   ): Seq[SparkPlan] = {
-    val aliasMap = mutable.HashMap[(Boolean, Expression), Alias]()
```
liancheng: Instead of building the alias map on the fly, I build it with an extra traversal over `aggregateExpressions` using an immutable map inside `toAlias`.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -271,59 +271,51 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
...
-      } else {
-        aliasMap.getOrElseUpdate(
-          (expr.deterministic, expr.canonicalized),
-          Alias(expr, expr.toString)()
-        )
-      }
+    val toAlias: AggregateExpression => Alias = {
+      lazy val deterministicAggAliases = aggregateExpressions.collect {
+        case e if e.deterministic => e -> Alias(e.canonicalized, e.toString())()
+      }.toMap
```
liancheng: Hmm, no need to be `lazy`, updating...

src/main/scala/org/apache/spark/sql/TiContext.scala
```diff
@@ -35,7 +34,6 @@ class TiContext(val session: SparkSession) extends Serializable with Logging {
 
   TiUtils.sessionInitialize(session, tiSession)
 
-
```
liancheng: Accidental whitespace changes introduced by scalafmt.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -162,40 +165,37 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
...
-            TiByItem.create(
-              BasicExpression.convertToTiExpr(order.child).get,
-              order.direction.sql.equalsIgnoreCase("DESC")
-            )
+  private def addSortOrder(request: TiDAGRequest, sortOrder: Seq[SortOrder]): Unit =
+    sortOrder.foreach { order: SortOrder =>
+      request.addOrderByItem(
+        TiByItem.create(
+          BasicExpression.convertToTiExpr(order.child).get,
+          order.direction.sql.equalsIgnoreCase("DESC")
```
liancheng: Ensured below that `sortOrder` will never be null.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -459,9 +401,42 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
...
+      val rewrittenResultExpressions = resultExpressions.map {
+        _.transform(rewrite).asInstanceOf[NamedExpression]
+      }
+
+      Some(
+        groupingExpressions,
+        (aggregateExpressions ++ extraAggregateExpressions).distinct,
+        rewrittenResultExpressions,
+        child
+      )
```
liancheng: Made `TiAggregation` always return results w/o any occurrences of `Average`. I think this is reasonable since the purpose of `TiAggregation` is to extract aggregate functions that can be pushed down to TiKV, while `Average` is not directly pushable.

Extracting the `Average` rewriting code path into a separate method also LGTM.

liancheng:
@Novemser This is reviewable now.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -141,68 +142,67 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
...
+  private def takeOrderedAndProject(
     limit: Int,
     sortOrder: Seq[SortOrder],
     child: LogicalPlan,
     project: Seq[NamedExpression]
   ): SparkPlan = {
-    // If sortOrder is not null, limit must be greater than 0
-    if (limit < 0 || (sortOrder == null && limit == 0)) {
+    // If sortOrder is empty, limit must be greater than 0
+    if (limit < 0 || (sortOrder.isEmpty && limit == 0)) {
```
liancheng: Did I miss something here? I don't think `SortOrder` can be null in Spark.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -263,129 +263,69 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
...
-          avgPushdownRewriteMap(aggExpr.resultId) = List(sumToPush, countToPush)
-          avgFinalRewriteMap(aggExpr.resultId) = List(sumFinal, countFinal)
-          List(sumFinal, countFinal)
-        case _ => aggExpr :: Nil
+        case e: Max     => aggExpr.copy(aggregateFunction = e.copy(child = partialResultRef))
+        case e: Min     => aggExpr.copy(aggregateFunction = e.copy(child = partialResultRef))
+        case e: Sum     => aggExpr.copy(aggregateFunction = e.copy(child = partialResultRef))
+        case e: First   => aggExpr.copy(aggregateFunction = e.copy(child = partialResultRef))
+        case _: Count   => aggExpr.copy(aggregateFunction = Sum(partialResultRef))
+        case _: Average => throw new IllegalStateException("All AVGs should have been rewritten.")
```
Novemser: It seems we should rewrite `Average` to `sum`/`count` as the original logic rather than throw an exception here.

Such querys like:
```sql
select avg(any_col) from any_table
```
will trigger `IllegalStateException`.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
Novemser: Returning the original `aggregateExpressions` which may still contain an `average` expression, passing it directly to method `groupAggregateProjection`(line 388) will cause an `IllegalStateException`.

I suggest remove the `average` expression in the final `aggregateExpressions`

liancheng: My bad, nice catch!

liancheng:
@Novemser Updated according to the comments. Thanks!

src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -263,129 +263,72 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
...
-        Alias(expr, expr.toString())()
-      } else {
-        aliasMap.getOrElseUpdate(
-          (expr.deterministic, expr.canonicalized),
-          Alias(expr, expr.toString)()
-        )
-      }
+    def aliasPushedPartialResult(e: AggregateExpression): Alias = {
+      deterministicAggAliases.getOrElse(e, Alias(e, e.toString())())
+    }
```
liancheng: @Novemser @ilovesoup The only non-deterministic one handled by `TiAggregate` is `First`, which doesn't seem to be necessary to be special-cased here. Any counter examples?

src/main/scala/org/apache/spark/sql/TiStrategy.scala
Novemser: Sorry it's my fault, originally `takeOrderedAndProject` is called in `collectLimit`, and an explicit null sortOrder will be passed, I removed that call in `collectLimit` but forgot to remove these checks here, thanks for your dedicated review!

src/main/scala/org/apache/spark/sql/TiStrategy.scala
Novemser: Seems true, currently I cannot find any counter example.

liancheng: Do we want to remove this logic? Was trying to add comments to explain why determinism is a concern here and then realized that it's not...

src/main/scala/org/apache/spark/sql/TiStrategy.scala
liancheng: I think it's fixed now (but I cannot test it...).

src/main/scala/org/apache/spark/sql/TiStrategy.scala
Novemser: There seems to be some issue with `deterministicAggAliases`, originally `aliasMap` was designed to  eliminate duplicated aggregations, e.g. 
SQL:
```sql
select count(col+1),count(1+col) from any_table
```
`count(col + 1)` and `count(1 + col)` will be reduced to only one canonicalized expression like `count(col + 1)`, and push this canonicalized expression down to TiKV, but current implementation doesn't serve the same as the original design(could be verified by executing the above sql).
Similar issue has been discussed here https://github.com/pingcap/tispark/pull/45
We may need some further discussion on this topic. : )

src/main/scala/org/apache/spark/sql/TiStrategy.scala
Novemser: Sorry for the inconvenience, I shall do tests for you.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
liancheng: Ah, my bad, it should be:

```scala
deterministicAggAliases.getOrElse(e.canonicalized, Alias(e, e.toString())())
```

liancheng:
@Novemser Fixed the canonicalization issue and rebased to the most recent master.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
liancheng: On the other hand, I'm happy to code w/o writing any tests, LOL

Novemser:
Brilliant, I shall run tests for this PR.

liancheng:
@Novemser BTW, I think it's possible to add webhooks to trigger PingCAP internal private tests for opened/updated PRs. Presenting test results to contributors would still be a problem (kinda like an OJ) but still better than nothing.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -263,129 +263,72 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
...
+        case _: Count   => aggExpr.copy(aggregateFunction = Sum(partialResultRef))
+        case _: Average => throw new IllegalStateException("All AVGs should have been rewritten.")
+        case _          => aggExpr
       }
     }
 
-    val pushdownAggregates = aggregateExpressions.flatMap { aggExpr =>
-      avgPushdownRewriteMap
-        .getOrElse(aggExpr.resultId, List(aggExpr))
-    }.distinct
```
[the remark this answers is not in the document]
Novemser: Yes, I think so.

Novemser:
@liancheng Pleased to receive your advice! We will implement that ASAP.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
Novemser: Very pleased to receive your response so fast, BTW I think determinism logic could be simplified as you suggested : )

src/main/scala/org/apache/spark/sql/TiStrategy.scala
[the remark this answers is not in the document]
Novemser: This seems a problem of mine...You can leave this to me. :rofl:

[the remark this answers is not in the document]
liancheng: Updated.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
liancheng: I can open a separate PR for this change. This one is already too big to review and track.

src/main/scala/org/apache/spark/sql/TiStrategy.scala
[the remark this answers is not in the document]
Novemser: @liancheng Sorry to notify you that after some tests, `canonicalized aggregateExpressions` may not be able to be pushed down directly, since some expressions like `count(col_1)` will canonicalized to `count(none)` and throw unexpected exception elsewhere.

I suggest restore original non-canonicalized logic and leave this for me to fix in further PRs.

Thanks a lot!

Novemser:
**SQL**
```sql
select count(any_col) from any_table
```

Result
Exception caught during process:
```java
Caused by: java.lang.RuntimeException: Couldn't find count(tp_int#100L)#362L in [sum#364L,count(tp_int#100L)#363L]
  at scala.sys.package$.error(package.scala:27)
  at org.apache.spark.sql.catalyst.expressions.BindReferences$$anonfun$bindReference$1$$anonfun$applyOrElse$1.apply(BoundAttribute.scala:94)
  at org.apache.spark.sql.catalyst.expressions.BindReferences$$anonfun$bindReference$1$$anonfun$applyOrElse$1.apply(BoundAttribute.scala:88)
  at org.apache.spark.sql.catalyst.errors.package$.attachTree(package.scala:52)
  ... 141 more
```


Seems some minor issue occurred in `residualAggregateExpressions`/`output`, or other parts.
@liancheng PTAL

src/main/scala/org/apache/spark/sql/TiStrategy.scala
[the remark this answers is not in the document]
liancheng: Thanks for the information. Reverted this change.

liancheng:
@Novemser Will look into the test failure later. I added both you and @ilovesoup to [my personal TiSpark fork][1] as collaborators. Feel free to push to this PR branch to fix any issues you find if you think this PR useful. I think this is easier since you have access to the full test results.

[1]: https://github.com/liancheng/tispark

liancheng:
@Novemser Could you please provide the full exception stack trace?

Novemser:
@liancheng Very glad to receive your invitation, will look into this problem.

Novemser:
Stack trace information updated. Seems `output` and `residualAggregateExpressions` has inconsistent expressions which violated aggregation function rule.

zhexuany:
@Novemser Could you paste Stack trace information here?

Novemser:
@zhexuany Full stack trace has been updated in the comment above : )

Novemser:
@liancheng The above issue has been fix by the last commit, `deterministicAggAliases` logic changed to  `e.canonicalized -> Alias(e, e.toString())()` and works GTM.

Novemser:
Test report:
TPCH
Tests run: 21  Tests succeeded: 21  Tests failed: 0  Tests skipped: 0

DAG
Tests run: 6841  Tests succeeded: 3766  Tests failed: 8  Tests skipped: 3067

Novemser:
This PR is LGTM.

@zhexuany @ilovesoup PTAL

src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -300,125 +301,73 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
...
-            case other => other
-          }
-          .asInstanceOf[NamedExpression]
-    )
+    aggregationToDAGRequest(groupingExpressions, aggregateExpressions.distinct, source, dagReq)
+
+    projects
+      .map { _.toAttribute.name }
+      .map { TiColumnRef.create }
+      .foreach { dagReq.addRequiredColumn }
```
liancheng: I see, missed the `addRequiredColumn` call while cleaning up the original code path, thanks for fixing it!

liancheng:
@Novemser I made a minor refactoring in my [last commit][1], which is arguably better (in the sense that eliminating mutable collections), depending on the coding style PingCAP prefers.

In Spark (or more specifically, Catalyst), we tend not to use mutable states (var or mutable collections) whenever possible for non-critical code paths to minimize side effects. Part of the reason why I neglected the `addRequiredColumns()` call was that `TiDAGRequest` is mutable and it's hard to track what side effects had happened.

If `TiDAGRequest` were immutable and always returned a new instance when more query components are added, you'll have to provide a new variable (and hence a new name) to the newly created instance (`dagReqWithRequiredColumns`, `dagReqWithFilters`, etc.) and then it would be pretty hard to neglect part of the logic while doing refactoring.

Again, this is quite subjective and feel free to disagree and revert :)

[1]: https://github.com/pingcap/tispark/pull/153/commits/ab9247b749ba4ce4c8dc57a2d44afab4c0be25f3

src/main/scala/org/apache/spark/sql/TiStrategy.scala
Novemser: Not your problem, this code snippet was just introduced into `master` branch few hours ago via this PR https://github.com/pingcap/tispark/pull/143, and I merged that change into this branch.

Novemser:
@liancheng I totally agree with you on the thought of not to use mutable states here and pretty appreciate your elegant solution of using immutable states. As for `TiDAGRequest`,  I'm quite in favor of your idea personally, feel free to open another PR to do such refactor if you like : )

Thanks again for your dedicated work!

Novemser:
Test report:
TPCH
Tests run: 21  Tests succeeded: 21  Tests failed: 0  Tests skipped: 0

DAG
Tests run: 6841  Tests succeeded: 3766  Tests failed: 8  Tests skipped: 3067
Result is as expected

Novemser:
BTW it's interesting and glad to serve like a manual only judge here, LOL

liancheng:
Please let me know anything that I'm expected to do to get this moving forward. One of the things I can think of is to split this single refactoring PR into smaller ones for easier tracking and reducing potential conflicts with other important outstanding PRs.

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -163,81 +165,75 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
...
+  }
 
-  def filterToDAGRequest(
+  private def filterToDAGRequest(
     filters: Seq[Expression],
     source: TiDBRelation,
     dagRequest: TiDAGRequest = new TiDAGRequest(pushDownType(), timeZoneOffset())
   ): TiDAGRequest = {
-    val tiFilters: Seq[TiExpr] = filters.collect { case BasicExpression(expr) => expr }
+    val tiFilters = filters.collect { case BasicExpression(expr) => expr }.asJava
```
zhexuany: please add type information back.

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -163,81 +165,75 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
...
-    if (sortOrder != null) {
-      sortOrder.foreach(
-        (order: SortOrder) =>
-          request.addOrderByItem(
-            TiByItem.create(
-              BasicExpression.convertToTiExpr(order.child).get,
-              order.direction.sql.equalsIgnoreCase("DESC")
-            )
+  private def addSortOrder(request: TiDAGRequest, sortOrder: Seq[SortOrder]): Unit =
+    sortOrder.foreach { order: SortOrder =>
```
zhexuany: why do we delete `null` value check here? If sortOrder is `null`, `foreach` could throw an NPE here.

zhexuany: I just saw you change null to nil when you call this function.

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
liancheng: Addressed. Thanks!

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
liancheng: Yea, please see [this comment thread][1] for more details.

[1]: https://github.com/pingcap/tispark/pull/153#discussion_r158456121

zhexuany: Yes. change to nil means empty list. Null value check is unnecessary.

liancheng:
@Novemser @zhexuany Thanks for the detailed review!

Novemser:
@liancheng We appreciate for your excellent work! : )

Issue #158, opened 2017-12-24
Speed up integration test with multi thread


birdstorm:
Now a full integration test is too slow.

PR #161, opened 2017-12-27
Add multi thread support for tests

birdstorm:
Add multi-thread for faster integration tests


birdstorm:
refer to #158 

PTAL @Novemser

Novemser:
@birdstorm Is this PR ready for review?

birdstorm:
@Novemser yes, PTAL

integtest/src/main/scala/com/pingcap/spark/TestCase.scala
```diff
@@ -57,11 +57,17 @@ class TestCase(val prop: Properties) extends LazyLogging {
 
   private val eps = 1.0e-2
 
-  protected var testsFailed = 0
-  protected var testsExecuted = 0
-  protected var testsSkipped = 0
-  protected var inlineSQLNumber = 0
-  protected var ignoredTest = 0
+  @volatile protected var testsFailed = 0
```
ilovesoup: Atomic operation like a += 1 cannot be protected by volatile only.
a += 1 -> a = a + 1 involve both load and store 2 operations. When load, add and store, volatile does not ensure the original value you load is not changed by other threads.
Use AtomicInteger instead.

integtest/src/main/scala/com/pingcap/spark/DAGTestCase.scala
```diff
@@ -106,21 +109,23 @@ class DAGTestCase(prop: Properties) extends TestCase(prop) {
...
+        testsExecuted += 1
+        inlineSQLNumber += 1
+        pool.submit(new SQLProcessorRunnable(sql, inlineSQLNumber, dbName))
+        logger.info("Running num: " + count + " sql")
       } catch {
         case _: Throwable => logger.error("result: Run SQL " + sql + " Failed!")
       }
     }
-    result = !result
+    pool.shutdown()
```
ilovesoup: How can you ensure tests done at this moment?

Novemser:
Maybe we can first switch to scalatest, and then see if we need multi-threading to improve performance.

birdstorm:
Closing this PR since we are switching to Scala Test Framework.

Issue #162, opened 2017-12-27
select count(something + constant) reveals NPE on master branch


birdstorm:
```
scala> spark.sql("select count(tp_int + 1) from full_data_type_table").show
java.lang.NullPointerException
  at com.pingcap.tikv.expression.TiColumnRef.getType(TiColumnRef.java:84)
  at com.pingcap.tikv.expression.scalar.TiScalarFunction.getArgType(TiScalarFunction.java:61)
  at com.pingcap.tikv.expression.scalar.TiScalarFunction.getArgTypeCode(TiScalarFunction.java:72)
  at com.pingcap.tikv.expression.scalar.TiScalarFunction.getSignature(TiScalarFunction.java:44)
  at com.pingcap.tikv.expression.scalar.TiScalarFunction.getName(TiScalarFunction.java:51)
  at com.pingcap.tikv.expression.TiFunctionExpression.toString(TiFunctionExpression.java:123)
  at com.google.guava4pingcap.base.Joiner.toString(Joiner.java:448)
  at com.google.guava4pingcap.base.Joiner$2.appendTo(Joiner.java:257)
  at com.google.guava4pingcap.base.Joiner.appendTo(Joiner.java:154)
  at com.google.guava4pingcap.base.Joiner.join(Joiner.java:197)
  at com.google.guava4pingcap.base.Joiner.join(Joiner.java:186)
  at com.pingcap.tikv.expression.TiFunctionExpression.toString(TiFunctionExpression.java:124)
  at com.google.guava4pingcap.base.Joiner.toString(Joiner.java:448)
  at com.google.guava4pingcap.base.Joiner$2.appendTo(Joiner.java:257)
  at com.google.guava4pingcap.base.Joiner.appendTo(Joiner.java:154)
  at com.google.guava4pingcap.base.Joiner.join(Joiner.java:197)
  at com.google.guava4pingcap.base.Joiner.join(Joiner.java:186)
  at com.pingcap.tikv.meta.TiDAGRequest.toString(TiDAGRequest.java:606)
  at org.apache.spark.sql.execution.CoprocessorRDD.verboseString(CoprocessorRDD.scala:50)
  at org.apache.spark.sql.catalyst.trees.TreeNode.generateTreeString(TreeNode.scala:562)
  at org.apache.spark.sql.execution.InputAdapter.generateTreeString(WholeStageCodegenExec.scala:258)
[...]

PR #164, opened 2017-12-27
Fix Aggregation push down NPE issue

Novemser:
We need to resolve the `TiColumnRef` after creating them from `pushdownAggregates`.
Should fix https://github.com/pingcap/tispark/issues/162


Novemser:
@birdstorm PTAL

src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -342,6 +342,9 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
     }.distinct
 
     aggregationToDAGRequest(groupingExpressions, pushdownAggregates, source, dagReq)
+    // Need to resolve column info after add aggregation push downs
+    dagReq.setTableInfo(source.table)
+    dagReq.resolve()
```
birdstorm: may move to coprocessorRDD()

Novemser:
@birdstorm Updated according to comment.

PR #197, opened 2018-01-04
Refactor expression and type system

ilovesoup:
1. Refactor type and decouple encoding/decoding utility to Codec classes.
2. Organize encoding / decoding of Key and values into specific classes
3. Refactor expressions and move expression processing logic to visitor classes and fixes type infer on the fly.
4. Refactor predicates processing logic according to new visitor and Key utilities.
5. Fix index encoding

This change is [Reviewable](https://reviewable.io/reviews/pingcap/tispark/197)


core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -137,47 +139,54 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
         throw new IllegalArgumentException("Should never be here")
 
       case f @ Sum(BasicExpression(arg)) =>
-        dagRequest.addAggregate(new TiSum(arg), fromSparkType(f.dataType))
+        dagRequest
+          .addAggregate(AggregateFunction.newCall(FunctionType.Sum, arg), fromSparkType(f.dataType))
 
-      case f @ Count(args) =>
+      case f @ Count(args) if args.length == 1 =>
```
Novemser: Why we always have `args.length == 1` here?

Novemser: Seems `count(any_thing)` has been rewritten to `count(1)` in `lineitem`, but why is this happen?

tikv-client/src/main/java/com/pingcap/tikv/expression/AggregateFunction.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.List;
+import java.util.Objects;
+
+public class AggregateFunction implements Expression {
+  public enum FunctionType {
+    Sum,
+    Count,
+    Min,
+    Max,
+    First
```
Novemser: What about `last`?

tikv-client/src/main/java/com/pingcap/tikv/expression/ComparisonBinaryExpression.java
```diff
@@ -0,0 +1,199 @@
...
+    if (normalizedPredicate != null) {
+      return normalizedPredicate.orElseGet(() -> null);
+    }
+    if (getLeft() instanceof Constant && getRight() instanceof ColumnRef) {
+      Constant left = (Constant) getLeft();
+      ColumnRef right = (ColumnRef) getRight();
+      Type newType;
+      switch (getComparisonType()) {
+        case EQUAL:
+          newType = EQUAL;
```
Novemser: `Type.`

tikv-client/src/main/java/com/pingcap/tikv/expression/ComparisonBinaryExpression.java
```diff
@@ -0,0 +1,199 @@
...
+          newType = Type.GREATER_THAN;
+          break;
+        case GREATER_EQUAL:
+          newType = Type.LESS_EQUAL;
+          break;
+        case GREATER_THAN:
+          newType = Type.LESS_THAN;
+          break;
+        case NOT_EQUAL:
+          newType = NOT_EQUAL;
```
Novemser: `Type.`, maybe stay consistent with other calls' is better.

tikv-client/src/main/java/com/pingcap/tikv/expression/Constant.java
```diff
@@ -0,0 +1,134 @@
...
+  public Constant(Object value, DataType type) {
+    this.value = value;
+    this.type = type == null ? getDefaultType(value) : type;
+  }
+
+  protected static boolean isIntegerType(Object value) {
+    return value instanceof Long
+        || value instanceof Integer
+        || value instanceof Short
+        || value instanceof Byte;
```
Novemser: Is it necessary to consider `BigInteger` here?

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/ProtoConverter.java
```diff
@@ -0,0 +1,282 @@
...
+
+  @Override
+  protected Expr visit(ColumnRef node, Void context) {
+    Expr.Builder builder = Expr.newBuilder();
+    builder.setTp(ExprType.ColumnRef);
+    CodecDataOutput cdo = new CodecDataOutput();
+    // After switching to DAG request mode, expression value
+    // should be the index of table columns we provided in
+    // the first executor of a DAG request.
+    IntegerCodec.writeLong(cdo, node.getColumnInfo().getOffset());
```
Novemser: There's one PR's logic omitted here. More details can be found here https://github.com/pingcap/tikv-client-lib-java/pull/203/files#diff-24cc1d9198b06e4fdc4ab8ca532df54aR82

tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
```diff
@@ -218,26 +242,21 @@ private DAGRequest buildTableScan() {
     // DO NOT EDIT EXPRESSION CONSTRUCTION ORDER
     // Or make sure the construction order is below:
     // TableScan/IndexScan > Selection > Aggregation > TopN/Limit
-    TiExpr filterExpr = mergeCNFExpressions(
-        getFilter().stream().peek(this::setColumnOffsets).collect(Collectors.toList())
-    );
-
```
Novemser: There's one PR's logic omitted here, please refer to https://github.com/pingcap/tikv-client-lib-java/pull/203

tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
```diff
@@ -281,31 +296,6 @@ private DAGRequest buildTableScan() {
     return request;
   }
 
-  private void setColumnOffsets(TiExpr expr) {
```
Novemser: Same as above, please refer to https://github.com/pingcap/tikv-client-lib-java/pull/203

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
ilovesoup: Only if use with count(distinct) you can have more than one columns here, but such cases cannot be pushed down. So we do only push count(column)
count(1) transform only happen when column is not null. In that case, count(1) equals to count(column)

tikv-client/src/main/java/com/pingcap/tikv/expression/AggregateFunction.java
ilovesoup: MySQL does not have first/last but TiDB has a built-in firstrow for aggregate-less group by.

tikv-client/src/main/java/com/pingcap/tikv/expression/ComparisonBinaryExpression.java
ilovesoup: Unified.

tikv-client/src/main/java/com/pingcap/tikv/expression/ComparisonBinaryExpression.java
ilovesoup: Unified.

tikv-client/src/main/java/com/pingcap/tikv/expression/Constant.java
ilovesoup: Seems not necessary for me since it looks like no column matches to BigInteger for now. If you need a big number you can use BigDecimal.

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
ilovesoup: Checked one more time it seems multiple columns without distinct means + 1 if any column is not null.

PR #199, opened 2018-01-05
Add potential missing test cases

Novemser:
Add potential missing cases

Originally we put `sum($col),avg($col),min($col),max($col)` together in one sql, not all columns could be aggregated by `sum`, `avg`(e.g. `timestamp`, `datetime` etc.), and original case will result in a `compile error` thus ignored by our test framework, however, columns do not support `sum`, `avg` could support the rest agg functions listed in the original sql, which we escaped.

### DIFF
--- a/integtest/src/main/scala/com/pingcap/spark/DAGTestCase.scala
+++ b/integtest/src/main/scala/com/pingcap/spark/DAGTestCase.scala
@@ -9,12 +9,6 @@ import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
 
 class DAGTestCase(prop: Properties) extends TestCase(prop) {
-  private val SPARK_DEFAULT_TO_IGNORE = Set[String](
-    "select tp_datetime,tp_date from full_data_type_table  where tp_datetime = tp_date order by id_dt  limit 20", // we cannot get anything from spark, but can get some data from TiDB
-    "select tp_date,tp_datetime from full_data_type_table  where tp_date = tp_datetime order by id_dt  limit 20",
-    "select tp_date,tp_datetime from full_data_type_table  where tp_date < tp_datetime order by id_dt  limit 20"
-  )
-
   protected val ARITHMETIC_CONSTANT: List[String] = List[String](
     java.lang.Long.MAX_VALUE.toString,
     java.lang.Long.MIN_VALUE.toString,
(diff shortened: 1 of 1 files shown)


integtest/src/main/scala/com/pingcap/spark/DAGTestCase.scala
```diff
@@ -145,9 +140,20 @@ class DAGTestCase(prop: Properties) extends TestCase(prop) {
     s"select tp_bigint%1000 a, count(*) from $TABLE_NAME group by (tp_bigint%1000) having sum(tp_bigint%1000) < 100 order by a"
   )
 
-  def createArithmeticAgg(): List[String] = colSet.map((col: String) => s"select sum($col),avg($col),min($col),max($col) from $TABLE_NAME group by $col ${orderBy(col)}").toList
+  def createSimpleSelect(): List[String] = colSet.map((col: String) => select(col) + orderBy(col) + limit()).toList
+
+  def createArithmeticAgg(): List[String] = colSet.flatMap((col: String) =>
+    Seq(s"select sum($col) from $TABLE_NAME",
```
birdstorm: We may just use `List()` instead of `Seq().toList`

integtest/src/main/scala/com/pingcap/spark/DAGTestCase.scala
```diff
@@ -145,9 +140,20 @@ class DAGTestCase(prop: Properties) extends TestCase(prop) {
...
+      s"select avg($col) from $TABLE_NAME",
+      s"select min($col) from $TABLE_NAME",
+      s"select abs($col) from $TABLE_NAME",
+      s"select max($col) from $TABLE_NAME"))
+    .toList
 
-  def createFirstLast(): List[String] = colSet.map((col: String) => s"select first($col), last($col) from $TABLE_NAME group by $col order by $col").toList
+  def createFirstLast(): List[String] = colSet.flatMap((col: String) =>
+    Seq(s"select first($col) from $TABLE_NAME ${groupBy("tp_nvarchar")} ${orderBy("tp_nvarchar")}",
```
birdstorm: same here

integtest/src/main/scala/com/pingcap/spark/DAGTestCase.scala
birdstorm: You may just use List() instead of Seq()

Novemser: Why?

Novemser: This is a flatMap, even using List(), result has to call `.toList`

birdstorm: my bad, I forgot colSet is a mutable set.

PR #206, opened 2018-01-11
use docker compose to provide tidb cluster service

zhexuany:
@ilovesoup @birdstorm @Novemser PTAL


README.md
```diff
@@ -77,11 +77,12 @@ To build all TiSpark modules :
...
 ```
 
-If you want to build single moudle, change directory to the one you want to: 
+## How to test
+We use docker to provide tidb cluster service which allow you can test across different platform. You have to install docker in order to test locally. CI is still under inverstigating and developing. A lot of work need to be done in our current integration test.
 
-```
-mvn clean package
-```
+Suppose you alrady install docker, you can use `docker-compose up -d` to launch tidb cluster service. If you want to see tidb cluster's log you can launch via `docker-compose up`. You can use `docker-compose down` to shutdown entire tidb cluster service. All data are stored in `data` directory at the root level of this project. Feel free to change it, but I suggest you keep it as same becuase that is the default value.
```
ilovesoup: typo already.

docker-compose.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - --name=pd0
+      - --client-urls=http://0.0.0.0:2379
+      - --peer-urls=http://0.0.0.0:2380
+      - --advertise-client-urls=http://pd0:2379
+      - --advertise-peer-urls=http://pd0:2380
+      - --initial-cluster=pd0=http://pd0:2380
+      - --data-dir=/data/pd0
+      - --config=/pd.toml
+    restart: on-failure
+  tikv0:
```
ilovesoup: How do you solve startup sequence problem?
According to the compose file, it seems tikv will not wait for pd "ready to serve" and might fail to start.
That's what I encounter before using docker-compose with TiDB.

zhexuany: There is a section `depends_on` which solve the issue you just mentioned.

ilovesoup: So now we use restart on failure to do "wait" logic, is that right?

birdstorm: I think `depends_on` on line 34 is what does "wait" logic here.

Issue #200, opened 2018-01-05
Error casting tp_text, tp_tinytext, tp_longtext etc. in aggregation functions


Novemser:
SQL
`select max(tp_text) from full_data_type_table`

Exception
```java
Job aborted due to stage failure: Task 0 in stage 1047.0 failed 1 times, most recent failure: Lost task 0.0 in stage 1047.0 (TID 97737, localhost, executor driver): java.lang.ClassCastException: java.lang.String cannot be cast to [B
	at org.apache.spark.sql.catalyst.expressions.BaseGenericInternalRow$class.getBinary(rows.scala:47)
	at org.apache.spark.sql.catalyst.expressions.GenericInternalRow.getBinary(rows.scala:194)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection.apply(Unknown Source)
	at org.apache.spark.sql.execution.CoprocessorRDD$$anonfun$doExecute$1$$anonfun$apply$2.apply(CoprocessorRDD.scala:65)
	at org.apache.spark.sql.execution.CoprocessorRDD$$anonfun$doExecute$1$$anonfun$apply$2.apply(CoprocessorRDD.scala:63)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
	at org.apache.spark.sql.execution.aggregate.SortBasedAggregationIterator.initialize(SortBasedAggregationIterator.scala:103)
	at org.apache.spark.sql.execution.aggregate.SortBasedAggregationIterator.<init>(SortBasedAggregationIterator.scala:113)
	at org.apache.spark.sql.execution.aggregate.SortAggregateExec$$anonfun$doExecute$1$$anonfun$3.apply(SortAggregateExec.scala:86)
	at org.apache.spark.sql.execution.aggregate.SortAggregateExec$$anonfun$doExecute$1$$anonfun$3.apply(SortAggregateExec.scala:77)
	at org.apache.spark.rdd.RDD$$anonfun$mapPartitionsInternal$1$$anonfun$apply$25.apply(RDD.scala:827)
	at org.apache.spark.rdd.RDD$$anonfun$mapPartitionsInternal$1$$anonfun$apply$25.apply(RDD.scala:827)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:38)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:323)
[...]

PR #211, opened 2018-01-15
Fix text cast error

Novemser:
Should fix https://github.com/pingcap/tispark/issues/200


Novemser:
DAG
Failed 6(Other error).

TPCH
All passed.

Our DAGTestCase has covered this issue, no need to add other test cases.

Novemser:
PTAL @birdstorm @ilovesoup

PR #213, opened 2018-01-16
fix getting next handle when encounter Long.MAX_VALUE

birdstorm:
Coprocessor accepts CLOSED_OPEN ranges. So when handle is Long.MAX_VALUE, the next handle should be acknowledged as well rather than throwing exception.

Also, other ranges we passed to coprocessor should be closed_open.

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/key/RowKey.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/key/RowKey.java
@@ -48,7 +48,7 @@ public class RowKey extends Key {
    * Initializes an imaginary globally MAXIMUM rowKey with tableId.
    */
   private RowKey(long tableId) {
-    super(encodeMaxHandle(tableId));
+    super(encodeBeyondMaxHandle(tableId));
     this.tableId = tableId;
     this.handle = Long.MAX_VALUE;
     this.maxHandleFlag = true;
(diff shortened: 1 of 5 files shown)


birdstorm:
DAG Test: fail 12 (other errors)
TPCH Test cannot be done because of sum/avg decode problems

birdstorm:
@ilovesoup PTAL

tikv-client/src/main/java/com/pingcap/tikv/util/KeyRangeUtils.java
```diff
@@ -96,10 +97,23 @@
...
 
+  /**
+   * Build a Coprocessor Range
+   *
+   * @param range Range with Comparable endpoints
+   * @return a CLOSED_OPEN range for coprocessor
+   */
   public static KeyRange makeCoprocRange(Range<Key> range) {
     if (!range.hasLowerBound() || !range.hasUpperBound()) {
       throw new TiClientInternalException("range is not closed");
```
ilovesoup: Change closed to bounded.

tikv-client/src/main/java/com/pingcap/tikv/key/RowKey.java
```diff
@@ -51,12 +66,16 @@ public static RowKey toRowKey(long tableId, TypedKey handle) {
...
 
+  public static RowKey toMaxRowKey(long tableId) {
+    return new RowKey(tableId);
+  }
+
   public static RowKey createMin(long tableId) {
     return toRowKey(tableId, Long.MIN_VALUE);
   }
 
   public static RowKey createMax(long tableId) {
```
ilovesoup: Should be BeyondMax since Max is long.max

birdstorm:
Test in progress

birdstorm:
This PR is not yet ready for merge because some problem occurs in index tests.

PR #214, opened 2018-01-16
Change CodecDataInput to use synchronized ByteArrayInputStream

Novemser:
Since some logic in `CodecDataInput.java` will only be invoked by one thread at the same time, we don't need to use a synchronized ByteArrayInputStream.


Novemser:
Optimized:
q1 -x5 elapsed time: 126018455141ns

Original master:
q1 -x5 elapsed time: 133661851691ns

q1(Turn off agg push down):
```sql
select min(tp_dt),min(tp_int) from tb_downgrade
```

PR #215, opened 2018-01-16
fix float/double/decimal decode in sum/avg logic

birdstorm:
Fixes type logic for TiKV version v1.0.5 and later.

Sum aggregation implements the following rules according to MySQL/TiKV logic:

* `sum(double)` and `sum(real)` will be of DOUBLE type.
* All other types in sum aggregation will be of NEW_DECIMAL type.
* support older version of TiKV by letting RealType able to decode NEW_DECIMAL.


birdstorm:
PTAL @ilovesoup

birdstorm:
DAG Test: 6 failed (with 4 text order TiSpark issue and 2 sum overflow TiKV issue)
TPCH Test: passed

PR #216, opened 2018-01-16
add tikv, pd, and tidb's config files

zhexuany:


.gitignore
```diff
@@ -61,4 +61,4 @@ tikv-client/proto
 
 # ignore tikv-client bazel build
 tikv-client/bazel-*
-
+data
```
birdstorm: Is this line necessary?

zhexuany: Yes. data will be created by docker. By ignoring this directory, we can avoid `git add` this directory.

PR #217, opened 2018-01-17
 Add requestOrder flag to indicate whether the output should be compared in order

birdstorm:
Since TiSpark does not guarantee retrieved data in order in most cases.
@ilovesoup PTAL


birdstorm:
DAG Tests: 6 failed (2 of 6: sum decode, 4 or 6: text decode)
Index Tests: 3 failed (3 of 3: handle overflow longmax)
TPCH Tests: passed

PR #218, opened 2018-01-18
fix unsupported types

birdstorm:
since all types have been refactored.


tikv-client/src/main/java/com/pingcap/tikv/expression/TypeBlacklist.java
```diff
@@ -17,13 +17,47 @@
 
 package com.pingcap.tikv.expression;
 
+import java.util.HashMap;
+
 public class TypeBlacklist extends Blacklist {
+  private static final HashMap<String, String> typeToMySQLMap = initialTypeMap();
```
Novemser: May use generic `Map` instead of `HashMap` here

tikv-client/src/main/java/com/pingcap/tikv/expression/TypeBlacklist.java
```diff
@@ -17,13 +17,48 @@
...
 package com.pingcap.tikv.expression;
 
+import java.util.HashMap;
+import java.util.Map;
+
 public class TypeBlacklist extends Blacklist {
+  private static final Map<String, String> typeToMySQLMap = initialTypeMap();
+
+  private static HashMap<String, String> initialTypeMap() {
+    HashMap<String, String> map = new HashMap<>();
```
ilovesoup: Source is enum why use string for map search here?

birdstorm:
INDEX Test: 1 failed.
DAG Test: 6 failed.
TPCH Test: passed.

birdstorm:
May have to dig into the one failure in index tests later, this pr is ready to merge for now. PTAL @ilovesoup

PR #222, opened 2018-01-25
Fix between test

birdstorm:
fix design problems in between tests


birdstorm:
This PR also resolves concern in #218

birdstorm:
I will add back the problematic test since it may be an possible issue described in #223 .

PR #231, opened 2018-02-01
Fix Average on bigint overflow

Novemser:
Rewrite original `Sum(LongType)` expression returning `LongType` which may cause Long overflow in Average cases to another `PromotedSum` expression.

### DIFF
--- a/core/src/test/scala/org/apache/spark/sql/expression/ArithmeticAgg0Suite.scala
+++ b/core/src/test/scala/org/apache/spark/sql/expression/ArithmeticAgg0Suite.scala
@@ -110,8 +110,6 @@ class ArithmeticAgg0Suite extends BaseTiSparkSuite {
   allCases foreach { query =>
     {
       test(query) {
-        if (query.equals("select avg(tp_bigint) from full_data_type_table"))
-          cancel("Average on bigint is pending to be fixed")
         runTest(query, query.replace("full_data_type_table", "full_data_type_table_j"))
       }
     }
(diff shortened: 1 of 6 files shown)


core/src/test/scala/org/apache/spark/sql/expression/ArithmeticAgg0Suite.scala
```diff
@@ -110,8 +110,6 @@ class ArithmeticAgg0Suite extends BaseTiSparkSuite {
   allCases foreach { query =>
     {
       test(query) {
-        if (query.equals("select avg(tp_bigint) from full_data_type_table"))
-          cancel("Average on bigint is pending to be fixed")
```
Novemser: Now this test case can pass.

core/src/test/scala/org/apache/spark/sql/expression/index/Between0Suite.scala
```diff
@@ -35,7 +35,9 @@ class Between0Suite extends BaseTiSparkSuite {
   allCases foreach { query =>
     {
       test(query) {
-        if (query.equals("select tp_real from full_data_type_table_idx  where tp_real between 4.44 and 0.5194052764001038 order by id_dt")) cancel("This case is pending to be solved in other PRs")
+        if (query.equals(
+              "select tp_real from full_data_type_table_idx  where tp_real between 4.44 and 0.5194052764001038 order by id_dt"
+            )) cancel("This case is pending to be solved in other PRs")
```
Novemser: Introduced by scalafmt.

Novemser:
PTAL @ilovesoup @birdstorm

PR #235, opened 2018-02-05
Fix charset to utf8

ilovesoup:
Since tidb use only utf-8 and we need to fix charset to utf-8 instead of jvm default.

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/tispark/235)


Novemser:
LGTM

PR #236, opened 2018-02-06
fix reset filter

birdstorm:
fix broken reset filter.


birdstorm:
PTAL @ilovesoup @Novemser

Issue #232, opened 2018-02-01
Error in complex group by query plan


Novemser:
```sql
scala> spark.sql("select tp_int + 1 from full_data_type_table  group by (tp_int + 1)").show
org.apache.spark.sql.catalyst.errors.package$TreeNodeException: execute, tree:
Exchange hashpartitioning((tp_int#46L + 1)#297L, 200)
+- *HashAggregate(keys=[(tp_int#46L + 1) AS (tp_int#46L + 1)#297L], functions=[], output=[(tp_int#46L + 1)#297L])
   +- TiDB CoprocessorRDD{[table: full_data_type_table] , Columns: [tp_int], Group By: [[[tp_int] PLUS 1] ASC]}

  at org.apache.spark.sql.catalyst.errors.package$.attachTree(package.scala:56)
  at org.apache.spark.sql.execution.exchange.ShuffleExchange.doExecute(ShuffleExchange.scala:112)
  at org.apache.spark.sql.execution.SparkPlan$$anonfun$execute$1.apply(SparkPlan.scala:114)
  at org.apache.spark.sql.execution.SparkPlan$$anonfun$execute$1.apply(SparkPlan.scala:114)
  at org.apache.spark.sql.execution.SparkPlan$$anonfun$executeQuery$1.apply(SparkPlan.scala:135)
  at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:151)
  at org.apache.spark.sql.execution.SparkPlan.executeQuery(SparkPlan.scala:132)
  at org.apache.spark.sql.execution.SparkPlan.execute(SparkPlan.scala:113)
  at org.apache.spark.sql.execution.InputAdapter.inputRDDs(WholeStageCodegenExec.scala:235)
  at org.apache.spark.sql.execution.aggregate.HashAggregateExec.inputRDDs(HashAggregateExec.scala:141)
  at org.apache.spark.sql.execution.WholeStageCodegenExec.doExecute(WholeStageCodegenExec.scala:368)
  at org.apache.spark.sql.execution.SparkPlan$$anonfun$execute$1.apply(SparkPlan.scala:114)
  at org.apache.spark.sql.execution.SparkPlan$$anonfun$execute$1.apply(SparkPlan.scala:114)
  at org.apache.spark.sql.execution.SparkPlan$$anonfun$executeQuery$1.apply(SparkPlan.scala:135)
[...]

PR #237, opened 2018-02-07
Fix group by without aggregates

ilovesoup:
When a group by issued without aggregates, Spark has contract that in the result expressions, there will be attributes referring back directly to input schema before aggregation. Those attributes are not dealt with correctly by replacing with TiKV output attributes.
Refer to https://github.com/pingcap/tispark/issues/232

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/tispark/237)

### DIFF
--- a/core/src/test/scala/org/apache/spark/sql/expression/ArithmeticAgg0Suite.scala
+++ b/core/src/test/scala/org/apache/spark/sql/expression/ArithmeticAgg0Suite.scala
@@ -35,8 +35,8 @@ class ArithmeticAgg0Suite extends BaseTiSparkSuite {
     // Below two cases will causes overflow if no predicates applied
     // While Spark does not check overflow for types other than decimal
     // TiKV will do the check and throw exception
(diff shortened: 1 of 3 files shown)


core/src/test/scala/org/apache/spark/sql/expression/ArithmeticAgg0Suite.scala
```diff
@@ -32,8 +32,11 @@ class ArithmeticAgg0Suite extends BaseTiSparkSuite {
     "select sum(tp_smallint) from full_data_type_table",
     "select sum(tp_float) from full_data_type_table",
     "select avg(tp_char) from full_data_type_table",
-    "select avg(tp_bigint) from full_data_type_table",
-    "select sum(tp_bigint) from full_data_type_table",
+    // Below two cases will causes overflow if no predicates applied
+    // While Spark does not check overflow for types other than decimal
+    // TiKV will do the check and throw exception
```
Novemser: `sum(tp_bigint)` overflow problem has been fixed in the latest master branch of TiKV, and TiKV will not throw overflow exception any more. I suggest removing the predicates and run these cases with the latest TiKV.

PR #254, opened 2018-03-05
Reduce jar size

Novemser:
Remove scala dependency.


birdstorm:
No need to run all tests for this pr

PR #263, opened 2018-03-11
Update R/Python/core/tikv-client modules to RC1

Novemser:
1. Fix statistics parameter for `Python`, `R` module.
2. Update TiSpark version to `1.0-RC1`

### DIFF
--- a/core/src/main/scala/com/pingcap/tispark/statistics/StatisticsManager.scala
+++ b/core/src/main/scala/com/pingcap/tispark/statistics/StatisticsManager.scala
@@ -49,15 +49,15 @@ private[statistics] case class StatisticsResult(histId: Long,
 }
 
 /**
-  * Manager class for maintaining table statistics information cache.
-  *
-  * Statistics information is useful for index selection and broadcast join support in TiSpark currently,
-  * and these are arranged follows:
-  * `statisticsMap` contains `tableId`->[[TableStatistics]] data, each table(id) will have a [[TableStatistics]]
-  * if you have loaded statistics information successfully.
-  *
-  * @param tiSession TiSession used for selecting statistics information from TiKV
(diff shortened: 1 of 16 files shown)


core/src/main/scala/com/pingcap/tispark/statistics/StatisticsManager.scala
```diff
@@ -49,15 +49,15 @@ private[statistics] case class StatisticsResult(histId: Long,
...
-  * Manager class for maintaining table statistics information cache.
-  *
-  * Statistics information is useful for index selection and broadcast join support in TiSpark currently,
-  * and these are arranged follows:
-  * `statisticsMap` contains `tableId`->[[TableStatistics]] data, each table(id) will have a [[TableStatistics]]
-  * if you have loaded statistics information successfully.
-  *
-  * @param tiSession TiSession used for selecting statistics information from TiKV
-  */
+ * Manager class for maintaining table statistics information cache.
```
Novemser: Accidentally introduced by scala formatter.

Novemser:
PTAL @ilovesoup @birdstorm

python/setup.py
```diff
@@ -2,7 +2,7 @@
 setup(
     name = 'pytispark',
     packages = ['pytispark'],
-    version = '0.1.3',
+    version = '1.0.0-RC1',
```
birdstorm: should this be `1.0.0-RC1` or `1.0-RC1`?

Novemser: I saw both on [Pypi](https://pypi.python.org)

Novemser: The former could do more fined grained version control.

Novemser: But we generally do not modify this module...

PR #269, opened 2018-03-12
Convert blob of string to stringType

ilovesoup:
Blob type might contains String content. In such cases, charset are not "binary" but values like "utf8" instead. Mapping them into StringType instead of BytesType.

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/tispark/269)


Novemser:
/run-all-tests tikv=release-1.1 tidb=release-1.1 pd=release-1.1

PR #273, opened 2018-03-13
Move null time out of test data

Novemser:
Since null `datetime`,`date`,`timestamp` will impact default JDBC behavior, which in our case, some invalid time like `000-00-00` will be converted to null(specified in JDBC url as`zeroDateTimeBehavior=convertToNull`,see more about this [here](https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-reference-configuration-properties.html)`) in expression evaluation, but actually they should not appear in the result record.

As we need to make sure https://github.com/pingcap/tispark/pull/265 and https://github.com/pingcap/tispark/pull/261 both work, along with minimal impact to other integration test cases, I removed invalid date data out of test data and add single test cases for such issue.


Novemser:
PTAL @birdstorm @ilovesoup

PR #274, opened 2018-03-13
Minor modification of TiHandle RDD print info

Novemser:
1. Remove useless filter for double read cases.
2. Improve `TiDAGRequest`'s toString.


Novemser:
No need to do this modification, original filter will still be used in some scenarios.

PR #276, opened 2018-03-13
Change downgrade index scan range merge logic

Novemser:
Merge originally potential discrete ranges into one complete big range in downgraded scan. Original merge logic cannot deal well with highly discrete ranges, and result in an expensive multi-range point scan on TiKV, which may cause too much pressure on TiKV.

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/util/KeyRangeUtils.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/util/KeyRangeUtils.java
@@ -15,7 +15,9 @@
 
 package com.pingcap.tikv.util;
 
-import com.google.common.collect.*;
+import com.google.common.collect.BoundType;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Range;
 import com.google.protobuf.ByteString;
 import com.pingcap.tikv.codec.CodecDataInput;
 import com.pingcap.tikv.codec.CodecDataOutput;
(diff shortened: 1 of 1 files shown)


tikv-client/src/main/java/com/pingcap/tikv/util/KeyRangeUtils.java
```diff
@@ -127,18 +127,17 @@ public static KeyRange makeCoprocRange(Range<Key> range) {
...
     }
 
-    RangeSet<Key> rangeSet = TreeRangeSet.create();
-    for (KeyRange keyRange : ranges) {
-      Range<Key> range = makeRange(keyRange.getStart(), keyRange.getEnd());
-      rangeSet.add(range);
-    }
+    // Merge discrete ranges into one complete big range
+    ByteString start = ranges.get(0).getStart();
+    ByteString end = ranges.get(ranges.size() - 1).getEnd();
```
Novemser: It is guaranteed `ranges` list contains at least 2 elements, see logic at the beginning of this function.

birdstorm: maybe better to add it into comment

birdstorm: I think this strategy has worst performance concerning a couple of random small ranges

ilovesoup: It's fine to leave it a big big range for now. Fine tuning needs a lot of consideration and might be in vain since we do not know the distance of handle from each other and detail cost balance of reading.

PR #277, opened 2018-03-13
Limit recursive call depth

Novemser:
In order to avoid stack overflow in some extreme cases, like we constructed the wrong DAGRequest/Range, TiKV may throw other error and cause `GrpcNeedRegionRefreshException` in `DAGIterator`, which recursively calls `process` again and again and forever won't recover. 

Fix this by limit the depth of such call.


Novemser:
Need to verify usage of `OtherError`, since TiDB seems doesn't handle this kind of error.

Issue #284, opened 2018-03-15
TiDBMapDatabase fails when Decimal length exceeding 38


birdstorm:
Concerning table:
```
CREATE TABLE `high_decimal_precision` (
  `a` decimal(50) DEFAULT NULL,
  `b` decimal(22,10) DEFAULT NULL,
  `c` int(11) DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin
```

```
scala> ti.tidbMapDatabase("test")
org.apache.spark.sql.AnalysisException: DecimalType can only support precision up to 38;
  at org.apache.spark.sql.types.DecimalType.<init>(DecimalType.scala:50)
  at org.apache.spark.sql.types.DecimalType$.apply(DecimalType.scala:42)
  at org.apache.spark.sql.types.DataTypes.createDecimalType(DataTypes.java:123)
  at com.pingcap.tispark.TiUtils$.toSparkDataType(TiUtils.scala:124)
  at com.pingcap.tispark.TiUtils$$anonfun$getSchemaFromTable$1.apply$mcVI$sp(TiUtils.scala:159)
  at scala.collection.immutable.Range.foreach$mVc$sp(Range.scala:160)
  at com.pingcap.tispark.TiUtils$.getSchemaFromTable(TiUtils.scala:151)
  at com.pingcap.tispark.TiDBRelation.schema$lzycompute(TiDBRelation.scala:38)
  at com.pingcap.tispark.TiDBRelation.schema(TiDBRelation.scala:38)
  at org.apache.spark.sql.execution.datasources.LogicalRelation.<init>(LogicalRelation.scala:40)
  at org.apache.spark.sql.SparkSession.baseRelationToDataFrame(SparkSession.scala:389)
  at org.apache.spark.sql.SQLContext.baseRelationToDataFrame(SQLContext.scala:307)
  at org.apache.spark.sql.TiContext$$anonfun$tidbMapDatabase$1$$anonfun$apply$3.apply(TiContext.scala:166)
  at org.apache.spark.sql.TiContext$$anonfun$tidbMapDatabase$1$$anonfun$apply$3.apply(TiContext.scala:146)
  at scala.collection.immutable.List.foreach(List.scala:381)
  at org.apache.spark.sql.TiContext$$anonfun$tidbMapDatabase$1.apply(TiContext.scala:146)
[...]

PR #285, opened 2018-03-15
TiDBMapDatabase fails when Decimal length exceeding 38

birdstorm:
Partial fix for #284

### DIFF
--- a/core/src/main/scala/com/pingcap/tispark/TiUtils.scala
+++ b/core/src/main/scala/com/pingcap/tispark/TiUtils.scala
@@ -125,7 +125,7 @@ object TiUtils {
       case _: DecimalType =>
         var len = tp.getLength
         if (len > MAX_PRECISION) {
-          println("Decimal precision exceeding 38, value will be truncated")
+          logger.warning("Decimal precision exceeding 38, value will be truncated")
           len = MAX_PRECISION
         }
         DataTypes.createDecimalType(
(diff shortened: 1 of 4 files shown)


birdstorm:
current behavior for table:
```
CREATE TABLE `high_decimal_precision` (
  `a` decimal(50) DEFAULT NULL,
  `b` decimal(22,10) DEFAULT NULL,
  `c` int(11) DEFAULT NULL

+--------------------------------------------------+-------------------------+------+
| a                                                | b                       | c    |
+--------------------------------------------------+-------------------------+------+
| 123456789012345678901234567890123456789012345678 |           12.3141100000 |    4 |
| 223456789012345678901234567890123456789012345678 | 123131414141.3143131100 |    6 |
+--------------------------------------------------+-------------------------+------+
```

```
> select b from high_decimal_precision
+-----------------------+
|b                      |
+-----------------------+
|12.3141100000          |
|123131414141.3143131100|
+-----------------------+

> select a from high_decimal_precision
+----+
|a   |
+----+
|null|
|null|
+----+

> select a, b from high_decimal_precision
+----+-----------------------+
|a   |b                      |
+----+-----------------------+
|null|12.3141100000          |
|null|123131414141.3143131100|
+----+-----------------------+
```

core/src/main/scala/com/pingcap/tispark/TiUtils.scala
```diff
@@ -120,9 +121,15 @@ object TiUtils {
       case _: IntegerType => sql.types.LongType
       case _: RealType    => sql.types.DoubleType
       // we need to make sure that tp.getLength does not result in negative number when casting.
+      // Decimal precision cannot exceed 38.
       case _: DecimalType =>
+        var len = tp.getLength
+        if (len > MAX_PRECISION) {
+          println("Decimal precision exceeding 38, value will be truncated")
```
ilovesoup: Don't use println. Use logger.

PR #287, opened 2018-03-19
[TISPARK-9]Optimize scan plan cost model

Novemser:
Adjust some CBO parameters
1. Refine double read cost.
2. Table size estimation.


tikv-client/src/main/java/com/pingcap/tikv/predicates/ScanAnalyzer.java
```diff
@@ -153,6 +153,7 @@ public ScanPlan buildScan(List<TiColumnInfo> columnList, List<Expression> condit
       // table name, index and handle column
       int indexSize = index.getIndexColumns().size() + 2;
       if (isDoubleRead) {
+        cost *= 4; // we assume double read's cost is around 4 times of simple table scan.
```
birdstorm: Could you plz explain why the factor is multiplied over the whole cost here?

From my perspective this factor should be EITHER multiplied over tableSize (indicating IO for each table row) OR indexSize (IO for each index row).

birdstorm: If a factor is applied to `indexSize`, the same factor should be applied to the `indexSize` in the else-clause

core/src/main/scala/com/pingcap/tispark/statistics/StatisticsManager.scala
```diff
@@ -240,6 +240,61 @@ class StatisticsManager(tiSession: TiSession) {
...
+      case TypeDouble => 8
+      case TypeDecimal => 10
+      case TypeNewDecimal => 10
+      case TypeNull => 1
+      case TypeTimestamp => 4
+      case TypeDate => 3
+      case TypeYear => 1
+      case TypeDatetime => 8
+      case TypeDuration => 3
+      case TypeString => 255 * goldenSplitFactor
```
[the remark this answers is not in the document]
Novemser: A higher level of abstraction is done in new commits.

Novemser:
PTAL, thx @birdstorm

PR #288, opened 2018-03-20
Fix bucket loading issue

Novemser:
Originally we may load index column buckets to non-index column buckets, fix this by handling each logic separately.

### DIFF
--- a/core/src/main/scala/com/pingcap/tispark/statistics/StatisticsHelper.scala
+++ b/core/src/main/scala/com/pingcap/tispark/statistics/StatisticsHelper.scala
@@ -74,8 +74,13 @@ object StatisticsHelper {
     val histVer = row.getLong(4)
     val nullCount = row.getLong(5)
     val cMSketch = if (checkColExists(histTable, "cm_sketch")) row.getBytes(6) else null
-    val indexInfos = table.getIndices.filter(_.getId == histID)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/pingcap/tispark/statistics/StatisticsHelper.scala
```diff
@@ -121,15 +147,20 @@ object StatisticsHelper {
...
-        var upperBound: Key = null
-        // all bounds are stored as blob in bucketTable currently, decode using blob type
-        lowerBound = TypedKey.toTypedKey(row.getBytes(2), DataTypeFactory.of(MySQLType.TypeBlob))
-        upperBound = TypedKey.toTypedKey(row.getBytes(3), DataTypeFactory.of(MySQLType.TypeBlob))
-        totalCount += count
-        buckets += new Bucket(totalCount, repeats, lowerBound, upperBound)
+        val isRowIndex = if (row.getLong(6) > 0) true else false
+        val isRequestIndex = matched.isIndex > 0
+        // if required DTO type(index/non index) is the same with the row
+        if (isRequestIndex && isRowIndex || !isRequestIndex && !isRowIndex) {
```
birdstorm: may use equality here.

Novemser: Done

PR #296, opened 2018-03-28
Decrease size function call times

Novemser:
Suppose calling`ranges.size()` takes `O(N)` in time complexity, the for loop below will take up to `O(N^2)` time.
Decrease this call to remain `O(N)` time complexity.


ilovesoup:
I don't think it's O(n).
For such case, I'd suggest you confirm
1. What concrete type of list does it received. Usually lists cache size instead of count it each time. You can debug it or simply print class name and read source code.
2. If you care about if compiler call method to load length or cache it automatically (suppose it knows length not changing over iteration it might choose to load just once), you can decompile it and check byte code.

PR #297, opened 2018-03-28
TISPARK-16 fix excessive dag columns

birdstorm:
Heretofore we added both project columns and filter columns into requiredColumns of dagRequest without distinguishing them when aggregations are encountered. When we preform joins with filters, the duplicated columns will affect row decoding and cause offset drift in decoding. 

Fixes by folding those columns beforehand.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
+++ b/core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
@@ -409,11 +409,12 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
       }
     }
 
-    val projectionTiRefs: Seq[TiColumnRef] = projects
-      .map { _.toAttribute.name }
-      .map { ColumnRef.create }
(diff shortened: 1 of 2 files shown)


birdstorm:
PTAL @ilovesoup

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -409,15 +409,12 @@ class TiStrategy(context: SQLContext) extends Strategy with Logging {
...
     }
 
-    val projectionTiRefs: Seq[TiColumnRef] = projects
-      .map { _.toAttribute.name }
-      .map { ColumnRef.create }
-
-    val filterTiRefs = filters
-      .collect { case BasicExpression(tiExpr) => tiExpr }
-      .flatMap { referencedTiColumns }
+    val projectSet = AttributeSet(projects.flatMap(_.references))
```
birdstorm: These lines are also used in filterDagRequest. Maybe better to unify them.

PR #298, opened 2018-03-28
 update functions tidbMapDatabase

buggithubs:
@Novemser PTAL


core/src/main/scala/org/apache/spark/sql/TiContext.scala
```diff
@@ -138,9 +138,13 @@ class TiContext(val session: SparkSession) extends Serializable with Logging {
     sqlContext.baseRelationToDataFrame(tiRelation)
   }
 
+  def tidbMapDatabase(dbName: String, dbNameAsPrefix:Boolean ): Unit = {
```
birdstorm: Could you please specify the reason for this modification and use scalafmt before commiting(you may do this simply by `mvn clean install -Dmaven.test.skip=true` in project's root)?

PR #304, opened 2018-04-08
TISPARK-21 delete redundant dag column that affects index single read

birdstorm:
The additional dag column when field is empty seems unnecessary. It also causes index single read to require a non index nor handle column. 
Delete its usage to fix index single read.

### DIFF
--- a/core/src/test/scala/org/apache/spark/sql/BaseTiSparkSuite.scala
+++ b/core/src/test/scala/org/apache/spark/sql/BaseTiSparkSuite.scala
@@ -123,7 +123,7 @@ class BaseTiSparkSuite extends QueryTest with SharedSQLContext {
 
   def execDBTSAndJudge(str: String, skipped: Boolean = false): Boolean =
     try {
-      compResult(querySpark(str), queryTiDB(str))
+      compResult(querySpark(str), queryTiDB(str), str.contains(" order by "))
     } catch {
       case e: Throwable => fail(e)
     }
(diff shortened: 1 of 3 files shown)


core/src/test/scala/org/apache/spark/sql/IssueTestSuite.scala
```diff
@@ -57,30 +82,6 @@ class IssueTestSuite extends BaseTiSparkSuite {
...
-    spark.sql("select b from prefix where b = \"bbc\"").explain
-    spark.sql("select b from prefix where b >= \"bbc\" and b < \"bbd\"").explain
-    spark.sql("select c, b from prefix where b > \"bb\" and b < \"bbc\"").explain
-    assert(execDBTSAndJudge("select a, b from prefix where b < \"bbc\""))
-    assert(execDBTSAndJudge("select a, b from prefix where a = 1 and b = \"bbb\""))
-    assert(execDBTSAndJudge("select b from prefix where b = \"bbc\""))
-    assert(execDBTSAndJudge("select b from prefix where b >= \"bbc\" and b < \"bbd\""))
-    assert(execDBTSAndJudge("select c, b from prefix where b = \"bb\" and b < \"bbc\""))
-  }
-
```
birdstorm: Prefix index tests have been moved to `prefixIndexTestSuite`.

core/src/test/scala/org/apache/spark/sql/BaseTiSparkSuite.scala
```diff
@@ -123,7 +123,7 @@ class BaseTiSparkSuite extends QueryTest with SharedSQLContext {
 
   def execDBTSAndJudge(str: String, skipped: Boolean = false): Boolean =
     try {
-      compResult(querySpark(str), queryTiDB(str))
+      compResult(querySpark(str), queryTiDB(str), str.contains(" order by "))
```
birdstorm: missed `isOrdered` check here which results in test case failure.

PR #305, opened 2018-04-09
[TISPARK-17][WIP]Partial support for spark 2.2.0

Novemser:
This is a partial upgrade for Spark 2.2.0, `Thriftserver` module is removed due to it's severe in-compatibility with 2.2.0. There's also minor change in `CollectHandles.scala` and `PromotedSum.scala` which removed some fields.


Novemser:
pyspark works normally.

Novemser:
We will move to 2.3
https://github.com/pingcap/tispark/pull/360

PR #306, opened 2018-04-09
make tidbMapDatabase support spark 2.2

birdstorm:


birdstorm:
seems this fix is not supported with spark v2.1

PR #310, opened 2018-04-11
Improve downgrade logic and efficiency

birdstorm:


core/src/main/scala/org/apache/spark/sql/execution/CoprocessorRDD.scala
```diff
@@ -374,13 +358,31 @@ case class RegionTaskExec(child: SparkPlan,
                   s", downgrade to table scan with ${tasks.size()} region tasks, " +
                   s"original index scan task has ${taskRanges.size()} ranges, will try to merge."
               )
+              downgrade = true
```
Novemser: It seems that variable `downgrade` has the same semantics as `satisfyDowngradeThreshold`, why do we need to maintain tow of these variables?

tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
```diff
@@ -797,4 +797,18 @@ public String toString() {
...
+  public TiDAGRequest copy() {
+    try {
+      ByteArrayOutputStream baos = new ByteArrayOutputStream();
+      ObjectOutputStream oos = new ObjectOutputStream(baos);
+      oos.writeObject(this);
+      ByteArrayInputStream bais = new ByteArrayInputStream(baos.toByteArray());
+      ObjectInputStream ois = new ObjectInputStream(bais);
+      return ((TiDAGRequest) ois.readObject());
+    } catch (Exception e) {
+      e.printStackTrace();
```
Novemser: It is not advisable to do `printStackTrace` in production code. Some exception handling is recommended here.

core/src/main/scala/com/pingcap/tispark/TiDBRelation.scala
```diff
@@ -73,10 +72,16 @@ class TiDBRelation(session: TiSession, tableRef: TiTableReference, meta: MetaMan
       )
       .head
 
+    val downgradeDagRequest = dagRequest.copy()
+    // We need to clear index info in order to perform table scan
+    downgradeDagRequest.clearIndexInfo()
+    downgradeDagRequest.resetFilters(downgradeDagRequest.getDowngradeFilters)
```
Novemser: I think maintaining `downgradeDagRequest` out of `RegionTaskExec` increases coupling of downgrade logic and it's caller, it might be more OOP to make downgrade logic fully encapsulated in `RegionTaskExec` and transparent to it's caller.

core/src/main/scala/org/apache/spark/sql/execution/CoprocessorRDD.scala
birdstorm: if checkisunaryrange is true, downgrade will not be proceeded.

tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
birdstorm: i will modify this function later.

core/src/main/scala/com/pingcap/tispark/TiDBRelation.scala
birdstorm: you are right.

birdstorm:
PTAL @Novemser @ilovesoup

Novemser:
Test Query:
`select avg(c_nationkey) from cust_test_select where c_custkey < 11000000` (11000000 rows)
Downgraded table scan CPU usage vs Index scan CPU usage:
The highest point can be reached is caused by heavy index scan requests issued to TiKV, the lowers are downgrade table scans. It seems downgraded plan released TiKV from extremely high pressure state.
image

Novemser:
There's still some optimization space left on the second read phase of downgrade tasks since it has such a low CPU utilization.

birdstorm:
@Novemser Thx for your help! I will dig into the second read phase for more optimization

birdstorm:
I think index double read is slower than we expected. @Novemser 
We might need to reconsider the value of `double_read_factor`.

birdstorm:
After profiling by @Novemser , the reason for slow index scan has been found and fixed.

core/src/main/scala/org/apache/spark/sql/execution/CoprocessorRDD.scala
```diff
@@ -179,29 +185,63 @@ case class RegionTaskExec(child: SparkPlan,
     val numHandles = longMetric("numHandles")
     val numIndexScanTasks = longMetric("numIndexScanTasks")
     val numDowngradedTasks = longMetric("numDowngradedTasks")
+    val numRegions = longMetric("numRegions")
+    val numIndexRangesScanned = longMetric("numIndexRangesScanned")
+    val numDowngradeRangesScanned = longMetric("numDowngradeRangesScanned")
+
+    val downgradeSplitFactor = 4
```
Novemser: Is this split still performance-boosting here?

PR #311, opened 2018-04-12
[DNM]Try to optimize codec and row reading process

ilovesoup:
1. Fix type and length calculation and cache result.
2. Change use byte array directly instead of stream class.
3. Use UTF8String of spark instead of raw string to avoid decoding twice.
4. Use java Unsafe to read directly instead of decode by bit manipulation.
5. Avoid bytes copy multiple times in bytes reading.
6. Remove unused decoder classes.

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/tispark/311)


zhexuany:
closing due too old.

PR #317, opened 2018-04-17
[TISPARK-26] fix binary test and delete redundant issue tests

birdstorm:


birdstorm:
seems that we could add back some tests concerning blob and text

Novemser:
Are blob and text test cases run normally?

birdstorm:
@Novemser They should be fixed by......NOW！

PR #318, opened 2018-04-18
delete unused ExprType

birdstorm:
Since tipb has deleted most of `ExprType`, now master breaks compilation.

Delete unused `ExprType` usage in our project.


birdstorm:
PTAL @ilovesoup @Novemser

tikv-client/src/main/java/com/pingcap/tikv/types/YearType.java
```diff
@@ -83,7 +83,7 @@ protected void encodeProto(CodecDataOutput cdo, Object value) {
 
   @Override
   public ExprType getProtoExprType() {
-    return ExprType.Year;
+    return ExprType.Uint64;
```
Novemser: Why we need to change this to `UInt64`?

birdstorm: year no longer exists in ExprType. Uint64 is its replacement.

birdstorm: Year has changed back to Int64. ICYMI @Novemser

Issue #321, opened 2018-04-19
New Error handling strategy will trigger zero storeid error and fail a lot of tasks


ilovesoup:
On handling KV not leader error, it will trigger PD retrieve using kv suggested store id. But sometimes it might be zero and will trigger infinite retry on PD handler and eventually fail that task.

Novemser:
Fixed.

PR #324, opened 2018-04-23
[TISPARK-27]Fix KV error handler logic when encounter store zero store id problem.

Novemser:
Related to #321


tikv-client/src/main/java/com/pingcap/tikv/operation/KVErrorHandler.java
```diff
@@ -31,11 +31,13 @@
 import io.grpc.StatusRuntimeException;
 import org.apache.log4j.Logger;
 
+import java.nio.ByteOrder;
 import java.util.function.Function;
 
 // TODO: consider refactor to Builder mode
 public class KVErrorHandler<RespT> implements ErrorHandler<RespT> {
   private static final Logger logger = Logger.getLogger(KVErrorHandler.class);
+  private static final int NO_LEADER_STORE_ID = 0;
```
ilovesoup: Please add comment for this: when zero id will be send.

PR #325, opened 2018-04-23
Refix stale epoch handling

ilovesoup:
According to TiDB's logic, we should trigger split and resend request.

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/tispark/325)

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/operation/KVErrorHandler.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/operation/KVErrorHandler.java
@@ -136,7 +136,7 @@ public class KVErrorHandler<RespT> implements ErrorHandler<RespT> {
       } else if (error.hasStaleEpoch()) {
         logger.warn(String.format("Stale Epoch encountered for region [%s]", ctxRegion));
         this.regionManager.onRegionStale(ctxRegion.getId());
-        throw new StatusRuntimeException(Status.fromCode(Status.Code.UNAVAILABLE).withDescription(error.toString()));
+        return false;
       } else if (error.hasServerIsBusy()) {
         logger.warn(String.format("Server is busy for region [%s], reason: %s", ctxRegion, error.getServerIsBusy().getReason()));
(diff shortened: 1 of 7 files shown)


tikv-client/src/main/java/com/pingcap/tikv/operation/KVErrorHandler.java
```diff
@@ -144,7 +144,7 @@ public boolean handleResponseError(BackOffer backOffer, RespT resp) {
       } else if (error.hasStaleEpoch()) {
         logger.warn(String.format("Stale Epoch encountered for region [%s]", ctxRegion));
         this.regionManager.onRegionStale(ctxRegion.getId());
-        throw new StatusRuntimeException(Status.fromCode(Status.Code.UNAVAILABLE).withDescription(error.toString()));
+        return false;
```
Novemser: 1. Since region stale indicates that we need to refresh region cache info, do we need to send this invalidation info back to driver?
2. In TiDB
```go
if staleEpoch := regionErr.GetStaleEpoch(); staleEpoch != nil {
	log.Debugf("tikv reports `StaleEpoch`, ctx: %v, retry later", ctx)
	err = s.regionCache.OnRegionStale(ctx, staleEpoch.NewRegions)
	return false, errors.Trace(err)
}
```
Such `StaleEpoch` error returns `false, errors.Trace(err)`(not `false, nil`), the error here will be captured by
```go
if regionErr != nil {
	retry, err := s.onRegionError(bo, ctx, regionErr)
	if err != nil {   // here
		return nil, errors.Trace(err)
	}
	if retry {
		continue
	}
}
```
and all the way up to
```go
tasks, err := worker.handleTaskOnce(bo, remainTasks[0], respCh)
if err != nil {
	resp := copResponse{err: errors.Trace(err)}
	worker.sendToRespCh(resp, respCh)
	return
}
```
So won't do re-split and retry logic in this layer(may be handled in upper transaction layer). I'm curious about why it does not not trigger re-split logic directly in this layer in TiDB. More investigation is needed.

PR #327, opened 2018-04-24
Add tests for unsigned data types

birdstorm:
Given the fact that TiDB is also incorrect on this issue, This PR also modifies the test framework to compare our result with pre-calculated answers rather than TiDB/Spark-JDBC output.


birdstorm:
PTAL @ilovesoup @Novemser

PR #328, opened 2018-04-25
[TISPARK-28] [TISPARK-29] Fix unsigned index

birdstorm:
* unsigned long will be treated as decimal(20,0)
* add function to read unsigned long from row


birdstorm:
There might be some problem with our `tableScanThreadPool` implementation.

PR #344, opened 2018-05-04
[TISPARK-36] Improve test logic

birdstorm:
TestSuite cannot handle tests with limit but without orderBy, e.g., select * from t limit 20.

If sql contains limit but not orderBy, the test result should be undefined.


birdstorm:
PTAL @Novemser

Novemser:
@birdstorm PTAL to these cases before merge.

PR #346, opened 2018-05-14
[TISPARK-38] Fix count(1) result is always 0 if subquery contains limit

birdstorm:
If subquery contains limit and we select count(1) from outer query, required columns will be empty so we cannot fetch correct answer of count(1). Fix by adding a dummy column to DAGRequest.


birdstorm:
PTAL @Novemser @ilovesoup

PR #347, opened 2018-05-16
[TISPARK-39] fix daylight saving time (DST)

birdstorm:
Fix DateTime parse error when DateTime is in DST gap.

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/codec/Codec.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/codec/Codec.java
@@ -16,16 +16,17 @@
 package com.pingcap.tikv.codec;
 
 
-import static com.google.common.base.Preconditions.checkArgument;
-
 import com.pingcap.tikv.exception.InvalidCodecFormatException;
 import gnu.trove.list.array.TIntArrayList;
-import java.math.BigDecimal;
-import java.util.Arrays;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.joda.time.LocalDate;
 import org.joda.time.LocalDateTime;
+
+import java.math.BigDecimal;
+import java.util.Arrays;
+
+import static com.google.common.base.Preconditions.checkArgument;
 
 public class Codec {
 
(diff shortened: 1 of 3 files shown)


tikv-client/src/main/java/com/pingcap/tikv/codec/Codec.java
```diff
@@ -510,10 +511,13 @@ static DateTime fromPackedLong(long packed, DateTimeZone tz) {
       int minute = (hms >> 6) & ((1 << 6) - 1);
       int hour = hms >> 12;
       int microsec = (int) (packed % (1 << 24));
-      return new DateTime(
+
+      LocalDateTime localDateTime = new LocalDateTime(
           year, month, day, hour,
-          minute, second, microsec / 1000,
-          tz);
+          minute, second, microsec / 1000);
```
ilovesoup: 1. Maybe you can put normal code path into a try-catch and only when exception happens then do the heavy path. I guess that might save you some time.
2. Please test the behavior of TiDB itself. For example, Timestamp type involves a TZ conversion from / to UTC. If time doesn't "jump" like in our solution, please write it down in document.

birdstorm: Timestamp is always UTC (stored in UTC and decoded as UTC). It should not be affected by this issue in any code path. DateTime is the source of bug since it needs to be decoded as different timezones according to our implementation.

tikv-client/src/main/java/com/pingcap/tikv/codec/Codec.java
```diff
@@ -510,10 +508,17 @@ static DateTime fromPackedLong(long packed, DateTimeZone tz) {
...
+      try {
+        return new DateTime(year, month, day, hour, minute, second, microsec / 1000, tz);
+      } catch (IllegalInstantException e) {
+        LocalDateTime localDateTime = new LocalDateTime(
+            year, month, day, hour,
+            minute, second, microsec / 1000);
+        DateTime dt = localDateTime.toLocalDate().toDateTimeAtStartOfDay(tz);
+        long millis = dt.getMillis() + localDateTime.toLocalTime().getMillisOfDay();
+        return new DateTime(millis, tz);
+      }
```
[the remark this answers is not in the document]
birdstorm: ok

PR #349, opened 2018-05-17
Add return DataFrame to mapTable function

ilovesoup:
This allow user to use DF directly instead of use "select *".

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/tispark/349)


birdstorm:
LGTM

PR #354, opened 2018-05-24
[TISPARK-43] Fix request fail with Key not in region after retrying NotLeaderError

birdstorm:
Should not update region context when error is not leader


birdstorm:
/run-all-tests tidb=release-2.0 tikv=release-2.0 pd=release-2.0

PR #355, opened 2018-05-28
[DNM] [TISPARK-44] Sleep a while in Statistics Test to fetch correct analyze result

birdstorm:
Previously, tests might not be able to get analyze result immediately from TiDB.

The root cause is that TiDB cannot guarantee that statistics information has written to TiKV when analyze statement is completed. 

So after we execute analyze in TiDB, we should sleep for a while to ensure CI has correct result.


ilovesoup:
I believe this is not the right thing to do. I will close the branch since stats related tests are shadowed for now.

PR #357, opened 2018-05-31
[TISPARK-45] Fix ScanIterator logic where index may be out of bound

birdstorm:
In previous implementation, if `index` is not initialized, `IndexOutOfBoundsException` would be thrown.


tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ScanIterator.java
```diff
@@ -115,8 +115,7 @@ public boolean hasNext() {
       if (!loadCache()) {
         endOfRegion = true;
```
birdstorm: I wonder if we should return false here. Since endOfRegion is true, hasNext should be false.

PR #360, opened 2018-06-04
[TISPARK-2] Update spark support to Spark2.3.0

### LINKED ISSUE #447: Add Spark 2.3 document
Using Spark 2.3, TiThriftServer is no longer needed, nor do we need `tidbMapDatabase` or define `TiContext` beforehand.
All above is switched into a new Spark Extension named TiExtension that can be configured in spark-default.conf. Thus the usage is totally different, and all documents need to be rewritten.

### LINKED ISSUE #450: TPCH tests will repeat when some query fails
`fail` will only quit current test, and result in reinitializing result set

birdstorm:
we will jump to Spark 2.3 from Spark 2.1

* Supports Spark 2.3.0+
* Eliminates TiContext, use Spark's native catalog as entrance
* Eliminates TiHiveThriftServer, now it is possible to use Spark's native HiveThriftServer
* Supports `use $database` syntax
* Supports `show tables` syntax
* Adds Spark 2.3 Documents, closes #447 
* Fixes TPCH tests, closes #450 
* Refactor StatisticsManager
* Fix memory leak #453


birdstorm:
run-all-tests

birdstorm:
The CI failure concerning decimal multiplication seems to be a bug in Spark2.3.0. I have raised an issue in Spark JIRA [SPARK-24468](https://issues.apache.org/jira/browse/SPARK-24468)

core/src/main/scala/org/apache/spark/sql/catalyst/catalog/TiCompositeSessionCatalog.scala
```diff
@@ -0,0 +1,344 @@
...
+ */
+case class IdentityPolicy(tiContext: TiContext) extends CompositeCatalogPolicy {
+  override val primaryCatalog: SessionCatalog = tiContext.tiConcreteCatalog
+  override val secondaryCatalog: SessionCatalog = tiContext.tiConcreteCatalog
+  override val tiConcreteCatalog: TiSessionCatalog = tiContext.tiConcreteCatalog
+  override val legacyCatalog: SessionCatalog = tiContext.legacyCatalog
+}
+
+/**
+ * Legacy catalog first policy.
```
ilovesoup: Any comment for the behavior?

core/src/main/scala/org/apache/spark/sql/catalyst/catalog/TiCompositeSessionCatalog.scala
```diff
@@ -0,0 +1,344 @@
...
+class TiCompositeSessionCatalog(tiContext: TiContext)
+    extends SessionCatalog(
+      tiContext.tiConcreteCatalog.externalCatalog,
+      EmptyFunctionRegistry,
+      tiContext.sqlContext.conf
+    )
+    with CompositeCatalogPolicy
+    with TiSessionCatalog {
+  // TODO: configuration for policy choosing.
+  val policy: CompositeCatalogPolicy = LegacyFirstPolicy(tiContext)
```
ilovesoup: Add an issue for this otherwise delete the one not used.

core/src/main/scala/org/apache/spark/sql/catalyst/catalog/TiCompositeSessionCatalog.scala
```diff
@@ -0,0 +1,344 @@
...
+  override val primaryCatalog: SessionCatalog = policy.primaryCatalog
+  override val secondaryCatalog: SessionCatalog = policy.secondaryCatalog
+  override val tiConcreteCatalog: TiSessionCatalog = policy.tiConcreteCatalog
+  override val legacyCatalog: SessionCatalog = policy.legacyCatalog
+
+  // Used to manage catalog change by setting current database.
+  private var currentCatalog: SessionCatalog = primaryCatalog
```
ilovesoup: When does this change?

core/src/main/scala/org/apache/spark/sql/catalyst/catalog/TiCompositeSessionCatalog.scala
```diff
@@ -0,0 +1,344 @@
...
+      .getOrElse(throw new NoSuchDatabaseException(dbDefinition.name))
+      .alterDatabase(dbDefinition)
+
+  override def getDatabaseMetadata(db: String): CatalogDatabase =
+    catalogOf(Some(db))
+      .getOrElse(throw new NoSuchDatabaseException(db))
+      .getDatabaseMetadata(db)
+
+  override def databaseExists(db: String): Boolean =
+    primaryCatalog.databaseExists(db) || secondaryCatalog.databaseExists(db)
```
ilovesoup: Why here use directly primary and secondary instead of catalogOf?

core/src/main/scala/org/apache/spark/sql/catalyst/catalog/TiCompositeSessionCatalog.scala
```diff
@@ -0,0 +1,344 @@
...
+  override val legacyCatalog: SessionCatalog = tiContext.legacyCatalog
+}
+
+/**
+ * A composition of two catalogs that behaves as a concrete catalog.
+ * @param tiContext
+ */
+class TiCompositeSessionCatalog(tiContext: TiContext)
+    extends SessionCatalog(
+      tiContext.tiConcreteCatalog.externalCatalog,
```
ilovesoup: How does name conflict resolved?

core/src/main/scala/org/apache/spark/sql/catalyst/catalog/TiCompositeSessionCatalog.scala
```diff
@@ -0,0 +1,344 @@
...
+        legacyCatalog
+          .listTables(legacyCatalog.getCurrentDatabase, pattern)
+          .filter(_.database.isEmpty)
+      case _ => Seq()
+    }
+    tables ++ extraLocalTempViews
+  }
+
+  // Following are all routed to primary catalog.
+  override def createDatabase(dbDefinition: CatalogDatabase, ignoreIfExists: Boolean): Unit =
```
ilovesoup: For TiSpark, it for now only make sense to create table on legacy catalog. Either you hardcode it, or you delete the other policy.

core/src/main/scala/org/apache/spark/sql/catalyst/catalog/TiCompositeSessionCatalog.scala
```diff
@@ -0,0 +1,344 @@
...
+    primaryCatalog.createDatabase(dbDefinition, ignoreIfExists)
+
+  override def getCachedPlan(t: QualifiedTableName, c: Callable[LogicalPlan]): LogicalPlan =
```
ilovesoup: Same below.

core/src/main/scala/org/apache/spark/sql/catalyst/catalog/TiCompositeSessionCatalog.scala
```diff
@@ -0,0 +1,344 @@
...
+  override def getDefaultDBPath(db: String): URI = primaryCatalog.getDefaultDBPath(db)
+
+  override def createTable(tableDefinition: CatalogTable, ignoreIfExists: Boolean): Unit =
+    primaryCatalog.createTable(tableDefinition, ignoreIfExists)
+
+  override def loadTable(name: TableIdentifier,
+                         loadPath: String,
+                         isOverwrite: Boolean,
+                         isSrcLocal: Boolean): Unit =
+    primaryCatalog.loadTable(name, loadPath, isOverwrite, isSrcLocal)
```
ilovesoup: What exactly is loadTable for?

core/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CollectHandles.scala
```diff
@@ -66,7 +62,7 @@ case class CollectHandles(child: Expression,
 
   // `CollectHandles` is non-deterministic since it's result depends on the
   // actual order of input rows.
-  override def deterministic: Boolean = false
+  override lazy val deterministic: Boolean = false
```
ilovesoup: Why is this lazy?

core/src/main/spark-2.1/scala/com/pingcap/tispark/TiConfigConst.scala
```diff
@@ -0,0 +1,42 @@
...
+  val ALLOW_INDEX_READ: String = "spark.tispark.plan.allow_index_read"
+  val COPROCESS_STREAMING: String = "spark.tispark.coprocess.streaming"
+  val UNSUPPORTED_PUSHDOWN_EXPR: String = "spark.tispark.plan.unsupported_pushdown_exprs"
+  val REGION_INDEX_SCAN_DOWNGRADE_THRESHOLD: String = "spark.tispark.plan.downgrade.index_threshold"
+  val KV_TIMEZONE_OFFSET: String = "spark.tispark.request.timezone.offset"
+  val UNSUPPORTED_TYPES: String = "spark.tispark.type.unsupported_mysql_types"
+  val ENABLE_AUTO_LOAD_STATISTICS: String = "spark.tispark.statistics.auto_load"
+  val CACHE_EXPIRE_AFTER_ACCESS: String = "spark.tispark.statistics.expire_after_access"
+  val SHOW_ROWID: String = "spark.tispark.show_rowid"
+  val DB_PREFIX: String = "spark.tispark.db_prefix"
```
ilovesoup: We should remove this right?

core/src/main/spark-2.1/scala/com/pingcap/tispark/TiSparkVersion.scala
```diff
@@ -0,0 +1,6 @@
+package com.pingcap.tispark
+
+object TiSparkVersion {
+  val version: String =
+    "Release Version: 1.1-SNAPSHOT\nGit Commit Hash: 22256357d90ee27fca93a22bcc56cf35943d2e5f\nGit Branch: support_spark_2.3.0\nUTC Build Time: 2018-09-21 06:22:11"
```
ilovesoup: Why is this here? Did you by mistake check it in?

ilovesoup: Why is this here? Did you by mistake check it in?

core/src/main/scala/org/apache/spark/sql/extensions/TiDDLRule.scala
```diff
@@ -0,0 +1,27 @@
...
+import org.apache.spark.sql.catalyst.rules.Rule
+import org.apache.spark.sql.execution.command._
+
+case class TiDDLRule(getOrCreateTiContext: SparkSession => TiContext)(sparkSession: SparkSession)
+    extends Rule[LogicalPlan] {
+  protected lazy val tiContext: TiContext = getOrCreateTiContext(sparkSession)
+
+  override def apply(plan: LogicalPlan): LogicalPlan = plan transformUp {
+    // TODO: support other commands that may concern TiSpark catalog.
+    case ShowDatabasesCommand(databasePattern) =>
```
ilovesoup: Please check if any database searching related behavior consistent.
If we issue a drop database in TiDB, ideally it should print "drop table not supported". But here it looks like it would be "no such database".

ilovesoup:
Please add a section of migration guide. The section should tell user how to modify the code. For example, if user uses map table functions, please tell them to change it. And we must test if the old style still works since user might have a lot of jobs that uses the old way to map table.

ilovesoup:
Totally copy all the files to new version is kind of ugly. What I actually expected is, make individual folders for those files not compatible only. If change is too much, you might consider let user use 1.1 release for 2.1 support and master and future version for 2.3.

README.md
```diff
@@ -4,11 +4,13 @@
...
 [![License](https://img.shields.io/github/license/pingcap/tispark.svg)](https://github.com/pingcap/tispark/blob/master/LICENSE)
 
-TiSpark is a thin layer built for running Apache Spark on top of TiDB/TiKV to answer the complex OLAP queries. It takes advantages of both the Spark platform and the distributed TiKV cluster, at the same time, seamlessly glues to TiDB, the distributed OLTP database, to provide a Hybrid Transactional/Analytical Processing (HTAP) to serve as a one-stop solution for online transactions and analysis.
+TiSpark is a thin layer built for running Apache Spark on top of TiDB/TiKV to answer complex OLAP queries. It takes advantages of both the Spark platform and the distributed TiKV cluster while seamlessly glues to TiDB, the distributed OLTP database, to provide a Hybrid Transactional/Analytical Processing (HTAP), and serves as a one-stop solution for online transactions and analysis.
 
 ## Getting TiSpark
 The current stable version is 1.0.
 
+You may download the latest binary build of TiSpark from [here](http://download.pingcap.org/tispark-latest-linux-amd64.tar.gz).
```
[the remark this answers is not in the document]
birdstorm: It will contain a default 2.3 version, I can also add 2.1 version in this gz file if needed.

core/src/main/scala/org/apache/spark/sql/catalyst/catalog/TiCompositeSessionCatalog.scala
birdstorm: catalogOf determines which catalog to use, it depends on databaseExists to gather the correct result.

core/src/main/scala/org/apache/spark/sql/catalyst/catalog/TiCompositeSessionCatalog.scala
birdstorm: It only implements functions in the abstract class.

core/src/main/scala/org/apache/spark/sql/catalyst/catalog/TiCompositeSessionCatalog.scala
birdstorm: same.

core/src/main/scala/org/apache/spark/sql/catalyst/catalog/TiCompositeSessionCatalog.scala
birdstorm: I think this makes sense, the other policy is not needed for now.

core/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CollectHandles.scala
birdstorm: Here, the lazy val just follows the deterministic value in abstract class Expression.

core/src/main/spark-2.1/scala/com/pingcap/tispark/TiConfigConst.scala
birdstorm: @ilovesoup It should still remain because we still cannot distinguish different databases in Hive and TiDB with the same name. The LegacyFirstPolicy/IdentityPolicy was designed to address their priority.

Consider that we both have database `db` in Hive and TiDB, when user calls `use db`, the result will be problematic if there is not a deterministic priority. Using LegacyFirstPolicy will use Hive databases in prior to TiDB databases, thus the database `db` in TiDB will be invisible.

In order to resolve this problem, a DB_PREFIX configuration is added when user wants to minimize its impact by adding a prefix in TiDB databases, "tidb_", for instance, so that they can call `use tidb_db` explicitly. In this circumstance, all databases in TiDB will have the same prefix.

core/src/main/spark-2.1/scala/com/pingcap/tispark/TiSparkVersion.scala
birdstorm: This SparkVersion is still useful when we call `select ti_version()`, I will modify the release version.

core/src/main/scala/org/apache/spark/sql/extensions/TiDDLRule.scala
birdstorm: I will create an issue for this. For now, DDLs such as `drop database/table` are not supported.

birdstorm:
> Totally copy all the files to new version is kind of ugly. What I actually expected is, make individual folders for those files not compatible only. If change is too much, you might consider let user use 1.1 release for 2.1 support and master and future version for 2.3.

As a matter of fact most files have changed when we switch to Spark 2.3. I think it is appropriate to stop supporting Spark 2.1.

README.md
```diff
@@ -51,47 +52,13 @@ From Spark-shell:
...
 spark.sql("select count(*) from lineitem").show
 ```
 
-## Metadata loading
-If you are using spark-shell, you need to manually load schema information as decribed above.
-
-If you have too many tables, you might choose to disable histogram preparison and loading will be faster.
-
-```
-ti.tidbMapDatabase("tpch", autoLoadStatistics = true)
```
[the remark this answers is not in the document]
birdstorm: I will fix it in another PR

core/src/main/scala/com/pingcap/tispark/TiUtils.scala
```diff
@@ -224,18 +223,31 @@ object TiUtils {
...
     if (req.getEstimatedCount > 0) {
       import java.text.DecimalFormat
       val df = new DecimalFormat("#.#")
       s" EstimatedCount:${df.format(req.getEstimatedCount)}"
     } else ""
+
+  /**
+   * Migrant from Spark 2.1.1 to support non-partial aggregate
+   */
+  def planAggregateWithoutPartial(groupingExpressions: Seq[NamedExpression],
```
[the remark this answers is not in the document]
birdstorm: This is migrated from Spark 2.1.1 because Spark 2.3 does not have `planAggregateWithoutPartial` anymore.

core/src/main/scala/org/apache/spark/sql/catalyst/catalog/TiCompositeSessionCatalog.scala
birdstorm: Commented in documents.

core/src/main/scala/org/apache/spark/sql/catalyst/catalog/TiCompositeSessionCatalog.scala
birdstorm: resolved by corresponding policy that configured in sparkConfig.

core/src/main/scala/org/apache/spark/sql/extensions/TiDDLRule.scala
birdstorm: Currently, all unsupported operations will throw a NotImplementedError

README.md
```diff
@@ -17,14 +21,29 @@ If you are using maven, add the following to your pom.xml:
...
 ```scala
 libraryDependencies += "com.pingcap.tispark" % "tispark-core" % "1.0"
 ```
 
 For other build tools, you can visit search.maven.org and search with GroupId [![Maven Search](https://img.shields.io/badge/com.pingcap-tikv/tispark-green.svg)](http://search.maven.org/#search%7Cga%7C1%7Cpingcap)(This search will also list all available modules of TiSpark including tikv-client).
 
+To build from sources that is compatible with Spark 2.3.0+, please follow the next section. 
+
+## How to build from sources
+TiSpark now supports Spark 2.3.0+. The previous version for Spark 2.1.0+ will only contain bug fixes in future, you may still get Spark2.1 support until release 1.1.
```
ilovesoup: Spark 2.1 (space in btw)

ilovesoup: until release 1.2

README.md
```diff
@@ -118,6 +103,7 @@ Below configurations can be put together with spark-defaults.conf or passed in t
 | spark.tispark.request.timezone.offset |  Local Timezone offset | An integer, represents timezone offset to UTC time(like 28800, GMT+8), this value will be added to requests issued to TiKV |
 | spark.tispark.show_rowid |  true | If to show implicit row Id if exists |
 | spark.tispark.db_prefix |  "" | A string indicating the extra database prefix for all databases in TiDB to distinguish them from Hive databases with the same name |
+| spark.tispark.catalog_policy |  "LegacyFirst" | must be one of "LegacyFirst" and "TiDBFirst", it determines which database to use when database name conflict in TiDB and Hive, the corresponding database not chosen will be hidden |
```
ilovesoup: Remove this and make a default behavior of adding prefix.

ilovesoup: And make a clear document for catalog behavior.

core/pom.xml
```diff
@@ -5,26 +5,27 @@
     <parent>
         <groupId>com.pingcap.tispark</groupId>
```
ilovesoup: bump spark version to latest spark, say, 2.3.3

PR #383, opened 2018-07-05
Fix scan iterator

ilovesoup:
It seems previously scan iterator throw exception if continuously read without end.
```
@Override
  public boolean hasNext() {
    ....
    if (cacheDrain()) {
      if (!loadCache()) {
        endOfRegion = true;
      }
    }
    ....
    return true;
  }
```
If cached drained and cache loading fail, hasNext will return wrongly true.

This PR also slightly refactored the ScanIterator since some parameters never used.

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/tispark/383)

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ScanIterator.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ScanIterator.java
@@ -17,11 +17,10 @@ package com.pingcap.tikv.operation.iterator;
 
 import static java.util.Objects.requireNonNull;
 
-import com.google.common.collect.Range;
 import com.google.protobuf.ByteString;
 import com.pingcap.tikv.TiSession;
 import com.pingcap.tikv.exception.TiClientInternalException;
-import com.pingcap.tikv.kvproto.Coprocessor.KeyRange;
+import com.pingcap.tikv.key.Key;
 import com.pingcap.tikv.kvproto.Kvrpcpb;
 import com.pingcap.tikv.kvproto.Metapb;
 import com.pingcap.tikv.region.RegionManager;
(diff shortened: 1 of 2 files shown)


ilovesoup:
/run-all-tests tikv=release-2.0 tidb=release-2.0 pd=release-2.0

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ScanIterator.java
```diff
@@ -17,59 +17,52 @@
...
-      RegionManager rm,
       long version) {
     this.startKey = requireNonNull(startKey, "start key is null");
-    this.batchSize = batchSize;
-    if (range == null) {
-      this.scanRange = Range.all();
-    } else {
-      this.scanRange = KeyRangeUtils.makeRange(range.getStart(), range.getEnd());
+    if (startKey.isEmpty()) {
+      throw new IllegalArgumentException("start key cannot be empty");
```
birdstorm: is this logic reachable?

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ScanIterator.java
```diff
@@ -83,66 +76,45 @@ private boolean loadCache() {
...
+      endOfScan = true;
       return false;
     }
     return true;
   }
 
   private Kvrpcpb.KvPair getCurrent() {
-    if (cacheDrain() && endOfRegion) {
-      throw new NoSuchElementException();
+    if (isCacheDrained()) {
```
birdstorm: do we need endofScan?

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ScanIterator.java
ilovesoup: If previous region is the right most region, then its EndKey will be next scan's startKey which is empty.

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ScanIterator.java
ilovesoup: Just a guard not to call PD again if user call hasNext multiple times.

ilovesoup:
Tested against current version of TiDB and seems working.

PR #394, opened 2018-07-17
Fix dependency for zeppelin

ilovesoup:
Zeppelin new version added dependencies for GRPC which is conflicted with our current version.
Also added META shade for reflection usage inside GRPC.

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/tispark/394)


birdstorm:
Tested running TPCH on zeppelin 0.8 with Spark interpreter on Spark cluster.
Yarn and Mesos not tested yet.

PR #396, opened 2018-07-18
[TISPARK-47] Fix StoreNotMatch

birdstorm:
Only invalidate store info when StoreNotLeader Error occurred

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/operation/KVErrorHandler.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/operation/KVErrorHandler.java
@@ -125,6 +125,10 @@ public class KVErrorHandler<RespT> implements ErrorHandler<RespT> {
     Errorpb.Error error = getRegionError(resp);
     if (error != null) {
       if (error.hasNotLeader()) {
+        // this error is reported from raftstore:
+        // peer of current request is not leader, the following might be its causes:
+        // 1. cache is outdated, region has changed its leader, can be solved by re-fetching from PD
+        // 2. leader of current region is missing, need to wait and then fetch region info from PD
         long newStoreId = error.getNotLeader().getLeader().getStoreId();
 
         // update Leader here
@@ -155,6 +159,9 @@ public class KVErrorHandler<RespT> implements ErrorHandler<RespT> {
 
         return true;
       } else if (error.hasStoreNotMatch()) {
+        // this error is reported from raftstore:
+        // store_id requested at the moment is inconsistent with that expected
+        // Solution：re-fetch from PD
         long storeId = ctxRegion.getLeader().getStoreId();
         logger.warn(String.format("Store Not Match happened with region id %d, store id %d", ctxRegion.getId(), storeId));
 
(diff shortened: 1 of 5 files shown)


ilovesoup:
I had some tests on TiDB's logic. If a store not match occurs (the tikv instance connected to does not have the correct store id in request):
1. store cache cleaned per error happened.
2. upon retry, RegionCache: GetRPCContext is called and address is generated using the error store id from PD. But this time, a different address is used since the storeID now points to the correct address (newly queried from PD). 
3. Again request sent but might encounter not leader this time. 
4. Follow the correct path for not leader.
Above is why we didn't see any switch leader and still changed address and resolve store not match.

ilovesoup:
We continuously get store not matches means the store address returned by getRegionStorePairByRegionId does not matches the context generated by region using TiRegion.peer.
```
builder.setRegionId(meta.getId()).setPeer(this.peer)
```
Above is the code to set store id (this.peer)
```
Peer leader = region.getLeader();
long storeId = leader.getStoreId();
return Pair.create(region, cache.getStoreById(storeId));
```
And above is getRegionStorePairByRegionId. According to code, region.peer is the same as region.getLeader. So if region is the same, store should match. 
So I guess, there is a racing that changed region causing getRegionById returns a different region than in the context.

ilovesoup:
It looks like, in onStoreNotMatch we might use the region returned by getRegionStorePairByRegionId instead of the one cached in RegionStoreClient.

birdstorm:
PTAL @ilovesoup

tikv-client/src/main/java/com/pingcap/tikv/operation/KVErrorHandler.java
```diff
@@ -145,30 +159,42 @@ public boolean handleResponseError(BackOffer backOffer, RespT resp) {
...
-        recv.onStoreNotMatch();
+        // this error is reported from raftstore:
+        // store_id requested at the moment is inconsistent with that expected
+        // Solution：re-fetch from PD
+        long storeId = ctxRegion.getLeader().getStoreId();
+        logger.warn(String.format("Store Not Match happened with region id %d, store id %d", ctxRegion.getId(), storeId));
+
+        this.regionManager.invalidateStore(storeId);
+        recv.onStoreNotMatch(this.regionManager.getStoreById(storeId));
+        notifyStoreCacheInvalidate(storeId);
```
birdstorm: Notify driver node that store cache is invalidated.

tikv-client/src/main/java/com/pingcap/tikv/operation/KVErrorHandler.java
```diff
@@ -145,30 +159,42 @@ public boolean handleResponseError(BackOffer backOffer, RespT resp) {
...
-
-        invalidateRegionStoreCache(ctxRegion);
-        recv.onStoreNotMatch();
+        // this error is reported from raftstore:
+        // store_id requested at the moment is inconsistent with that expected
+        // Solution：re-fetch from PD
+        long storeId = ctxRegion.getLeader().getStoreId();
+        logger.warn(String.format("Store Not Match happened with region id %d, store id %d", ctxRegion.getId(), storeId));
+
+        this.regionManager.invalidateStore(storeId);
```
birdstorm: Just invalidate store cache is enough for StoreNotMatch error. See [this line in tidb](https://github.com/pingcap/tidb/blob/a8012590dcce643db7a837266f6849b9435ad57e/store/tikv/region_request.go#L207)

PR #401, opened 2018-07-19
Fix decimal decoding

ilovesoup:
Old implementation was wrong dealing with unsigned / signed part of Java and some code path might break in specific values.

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
@@ -15,6 +15,7 @@
 
 package com.pingcap.tikv.codec;
 
+import com.google.common.annotations.VisibleForTesting;
 import java.math.BigDecimal;
 import java.util.Arrays;
 
@@ -92,8 +93,7 @@ public class MyDecimal {
    *
    * @param precision precision specifies total digits that this decimal will be..
    * @param frac frac specifies how many fraction digits
-   * @param bin bin is binary string which represents a decimal value. TODO: (zhexuany) overflow and
-   *     truncated exception need to be done later.
+   * @param bin bin is binary string which represents a decimal value.
    */
   public int fromBin(int precision, int frac, int[] bin) {
     if (bin.length == 0) {
@@ -134,13 +134,13 @@ public class MyDecimal {
         wordsIntTo = wordBufLen;
         wordsFracTo = 0;
         overflow = true;
+      } else {
+        wordsIntTo = wordsInt;
+        wordsFracTo = wordBufLen - wordsInt;
+        truncated = true;
       }
-      wordsIntTo = wordsInt;
-      wordsFracTo = wordBufLen - wordsInt;
-      truncated = true;
     }
-    wordsIntTo = wordsInt;
-    wordsFracTo = wordsFrac;
+
     if (overflow || truncated) {
       if (wordsIntTo < oldWordsIntTo) {
         binIdx += dig2bytes[leadingDigits] + (wordsInt - wordsIntTo) * wordSize;
@@ -151,8 +151,8 @@ public class MyDecimal {
     }
 
     this.negative = mask != 0;
-    this.digitsInt = wordsInt * digitsPerWord + leadingDigits;
-    this.digitsFrac = wordsFrac * digitsPerWord + trailingDigits;
+    this.digitsInt = (byte)(wordsInt * digitsPerWord + leadingDigits);
+    this.digitsFrac = (byte)(wordsFrac * digitsPerWord + trailingDigits);
 
     int wordIdx = 0;
     if (leadingDigits > 0) {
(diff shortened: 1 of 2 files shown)


ilovesoup:
/run-all-tests tikv=release-2.0 tidb=release-2.0 pd=release-2.0

tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
```diff
@@ -151,8 +150,8 @@ public int fromBin(int precision, int frac, int[] bin) {
     }
 
     this.negative = mask != 0;
-    this.digitsInt = wordsInt * digitsPerWord + leadingDigits;
-    this.digitsFrac = wordsFrac * digitsPerWord + trailingDigits;
+    this.digitsInt = (byte)(wordsInt * digitsPerWord + leadingDigits);
+    this.digitsFrac = (byte)(wordsFrac * digitsPerWord + trailingDigits);
```
zhexuany: type conversion seems unnecessary. Digits of int and fraction can never go beyond 127. tidb does this simply because it is static type language. It has to do such conversion otherwise compiler will not be happy.

tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
```diff
@@ -268,21 +267,22 @@ private int digitsToWords(int digits) {
...
    * @param start start indicates the where start to read.
    */
-  private int readWord(int[] b, int size, int start) {
+  @VisibleForTesting
+  public static int readWord(int[] b, int size, int start) {
     int x = 0;
     switch (size) {
       case 1:
-        x = b[start];
+        x = (byte)b[start];
```
zhexuany: same reason.

tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
```diff
@@ -268,21 +267,22 @@ private int digitsToWords(int digits) {
...
         break;
       case 2:
-        x = (b[start] << 8) + b[start + 1];
+        x = (((byte)b[start]) << 8) + (b[start + 1] & 0xFF);
         break;
       case 3:
         int sign = b[start] & 128;
         if (sign > 0) {
-          x = 255 << 24 | (b[start] & 0xFF) << 16 | (b[start + 1] & 0xFF) << 8 | (b[start + 2]);
```
zhexuany: Just a reminder: b's type is byte[] which shares a range from -128 to 127 whereas golang's byte shares range from 0 to 255. We are better take `and` operation with `0xFF`.

zhexuany:
We are better to add some comments for `precision()`  anb `fraction()`. It returns real precision and fraction.

tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
birdstorm: I don't think so, they are still the same in binary forms.

birdstorm: actually, i think we should use <<<

zhexuany: Additionally, `255 << 24` could be overflow. We need convert 255 to a uint and then convert back to int.

zhexuany: my bad, java does not have uint..

zhexuany: 0xFFL << 24 | (b[start] & 0xFF) << 16 | (b[start + 1] & 0xFF) << 8 | (b[start + 2]);

birdstorm: you may use ~(((b[start] & 7FFF) << 16 | (b[start + 1] & 0xFF) << 8 | (b[start + 2] & 0xFF)) + 1);

tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
ilovesoup: In java byte is signed. b[start] is unsigned in fact since we read as unsigned byte.
So actually convert to byte will cause it to be signed. It's not trivial.

tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
ilovesoup: 1. b's type is int not byte[].
2. 255 as const literal is int type.

zhexuany: 255 << 24 has correct binary form. Numeric overflow does not affect binary. ignore my comment.

tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
ilovesoup: To keep the same behavior as in TiDB which does a cap.

ilovesoup:
/run-all-tests tikv=release-2.0 tidb=release-2.0 pd=release-2.0

PR #405, opened 2018-07-23
set IgnoreUnknown=true for JsonCreator

birdstorm:
To prevent we might be incompatible with future tidb version in Json format, we should add IgnoreUnknown=true.


tikv-client/src/main/java/com/pingcap/tikv/meta/TiIndexInfo.java
```diff
@@ -29,6 +30,7 @@
 
 import static java.util.Objects.requireNonNull;
 
+@JsonIgnoreProperties(ignoreUnknown = true)
```
ilovesoup: Do we have other parts that need ignore properties?

PR #410, opened 2018-07-31
Add parentheses when call sides effects method

ariesdevil:
ref: https://github.com/pingcap/tispark/blob/master/core/src/main/scala/org/apache/spark/sql/hive/thriftserver/TiSparkSQLEnv.scala#L50

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/hive/thriftserver/TiSparkSQLEnv.scala
+++ b/core/src/main/scala/org/apache/spark/sql/hive/thriftserver/TiSparkSQLEnv.scala
@@ -47,7 +47,7 @@ private[hive] object TiSparkSQLEnv extends Logging {
         )
 
       // Injection point for TiSparkSession
-      val sparkSession = TiSparkSession.builder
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/sql/TiSparkSession.scala
```diff
@@ -238,7 +238,7 @@ object TiSparkSession {
     }
   }
 
-  def builder(): Builder = new Builder
+  def builder: Builder = new Builder
```
birdstorm: I don't think this is where the modification takes place in this PR. 
If you have checked the `builder` function in `SparkSession` beforehand, you should find that `def builder()` is the correct way to define this Builder.
In contrast, you should modify parentheses in this line https://github.com/pingcap/tispark/blob/9c5063004534c9f76218b4ef7021fc2677eefce4/core/src/main/scala/org/apache/spark/sql/hive/thriftserver/TiSparkSQLEnv.scala#L50

PR #415, opened 2018-08-10
Fix multi region tasks

### LINKED ISSUE #414: ProceedTaskOrThrow might fail the check because of outdated regionManager

birdstorm:
Fix #414 , it is possible that we encounter multiple region tasks after shuffling due to outdated region cache information.


core/src/main/scala/org/apache/spark/sql/execution/CoprocessorRDD.scala
```diff
@@ -313,30 +295,32 @@ case class RegionTaskExec(child: SparkPlan,
               logger.info("Single batch handles size:" + handleList.size())
               // After `splitAndSortHandlesByRegion`, ranges in the task are arranged in order
               // TODO: Maybe we can optimize splitAndSortHandlesByRegion if we are sure the handles are in same region?
-              val task = proceedTasksOrThrow(
+              val indexTask =
```
ilovesoup: if not a single task after split, please name it indexTasks.

core/src/main/scala/org/apache/spark/sql/execution/CoprocessorRDD.scala
```diff
@@ -313,30 +295,32 @@ case class RegionTaskExec(child: SparkPlan,
...
-              numIndexRangesScanned += taskRange.size
+
+              indexTask.foreach { task =>
+                val taskRange = task.getRanges
+                val tasks = splitTasks(task)
+                numIndexScanTasks += tasks.size
+
+                logger.info(s"Single batch RegionTask size:${tasks.size}")
+                tasks.foreach(task => {
+                  logger.info(
```
ilovesoup: I feel potentially this info log is dangerous. If user set batch size to 1, this log might bring down the whole cluster since each handle it will print something.
Do we really need this information? If not, make it trace / debug level.

birdstorm: That truly was a blind spot

core/src/main/scala/org/apache/spark/sql/execution/CoprocessorRDD.scala
birdstorm: done

PR #417, opened 2018-08-16
Feature:  decode json to string

alex-lx:
Convert JSON from binary format to String format. see example below
```
scala>  spark.sql("select * from json_test where ").show
+-----+--------------------+
|   id|         person_desc|
+-----+--------------------+
|    1|{"authors":[{"fir...|
|    2|        {"abc":null}|
|30001|"12345678\n\"世界12...|
|30002|              "1234"|
|30003|{"abc":123,"efg":...|
+-----+--------------------+
```

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -17,7 +17,7 @@
     <properties>
         <scalatest.version>3.0.4</scalatest.version>
         <scalaj.version>2.3.0</scalaj.version>
-        <mysql.connector.version>5.1.18</mysql.connector.version>
+        <mysql.connector.version>5.1.44</mysql.connector.version>
         <play.version>2.6.8</play.version>
     </properties>
     <dependencies>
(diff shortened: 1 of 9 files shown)


birdstorm:
Hi, @alex-lx. Please let me express my gratitude to your brilliant patch. 

I think you could add some tests concerning JSON types, you may refer to some existing tests in IssueTestSuite. I will start the review now.

alex-lx:
@birdstorm, thanks for your review. I will add the test soon.

alex-lx:
Hello, @birdstorm. Test added, Please take a look.

alex-lx:
Sorry, I can‘t reproduce the failure in my environment. I think it may cause by the version of tidb. Would you mind if the test does not use `judge()` to compare with `jdbc`'s result but just compare to the corresponding literal string?

birdstorm:
@alex-lx Of course, I will handle this test failure. Thanks again for your contribution.

birdstorm:
Test fails due to https://github.com/pingcap/tidb/pull/7389

PR #430, opened 2018-08-30
Fix load table might result table not exists

### LINKED ISSUE #393: 通过tispark查询某张表的时候提示table not exist
通过mysql-client查询某张表的时候是正常的
image
通过tispark查询的时候提示Table not exist
image
tispark的版本如下
image

birdstorm:
Fix #393 

And enclose a copy of conclusion:

>Bug in ScanIterator caused seldom table not exists.
>
>When currentCache is null means no keys found, whereas currentCache is empty means no values found. The difference lies in whether to continue scanning, because chances are that the same key is split in another region because of pending entries, region split, e.t.c.
>
>e.g., Assume we fetch table information in DB:123, ScanIterator sends a request containing an encoded key, say [1,2,3,0,0,0,1], and finds the first region that contains this key. If the region's range(e.g., [[1,2,3,0,0,0,0], [1,2,3,0,0,1,1])) was not enough to cover all keys [1,2,3,0,0,0,1], it is possible that it returns a partial result because some other table meta are contained in another region(e.g., [[1,2,3,0,0,1,1], [1,2,3,0,0,2])).
>
>Because ScanIterator was only used in fetching table meta info, tests not covering thousands of tables are unable to reveal this bug beforehand.

### DIFF
--- a/core/src/test/scala/org/apache/spark/sql/IssueTestSuite.scala
+++ b/core/src/test/scala/org/apache/spark/sql/IssueTestSuite.scala
@@ -20,25 +20,6 @@ import org.apache.spark.sql.functions.{col, sum}
 
 class IssueTestSuite extends BaseTiSparkSuite {
 
-  val tableNum = 5000
-
-  test("Test multiple tables across regions") {
-    for (x <- 1 to tableNum) {
-      tidbStmt.execute(s"DROP TABLE IF EXISTS `test$x`")
-      tidbStmt.execute(
-        s"""CREATE TABLE `test$x` (
-          |  `a` int(11) DEFAULT NULL
-          |) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin
-        """.stripMargin
-      )
-    }
-    refreshConnections()
-
-    for (x <- 1 to tableNum) {
-      ti.tidbMapTable("tispark_test", s"test$tableNum")
-    }
-  }
-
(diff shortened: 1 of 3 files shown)


tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ScanIterator.java
```diff
@@ -69,7 +69,11 @@ private boolean loadCache() {
     try (RegionStoreClient client = RegionStoreClient.create(region, store, session)) {
       BackOffer backOffer = ConcreteBackOffer.newScannerNextMaxBackOff();
       currentCache = client.scan(backOffer, startKey, version);
-      if (currentCache == null || currentCache.size() == 0) {
+      // currentCache is null means no keys found, whereas currentCache is empty means no values found
```
ilovesoup: Please  add mock test case that covers empty region in b/w

core/src/test/scala/org/apache/spark/sql/IssueTestSuite.scala
```diff
@@ -20,6 +20,25 @@ import org.apache.spark.sql.functions.{col, sum}
 
 class IssueTestSuite extends BaseTiSparkSuite {
 
+  val tableNum = 5000
```
ilovesoup: OK...Just leave this till we have more advanced mock framework.

PR #438, opened 2018-09-05
[DNM] Test publish 1.1

birdstorm:
Test only


birdstorm:
Closes because we are switching to master

PR #442, opened 2018-09-06
Fix myDecimal bug and add codec test

Handora:
* Add codec unit tests for improving code coverage
* Fix myDecimal bug that causes out of Index error when enlarging the fraction

PR #444, opened 2018-09-06
fix myDecimal bug that can't enlarge the fraction

Handora:
fix myDecimal bug that causes out of Index error when enlarging the fraction.


Handora:
also fix a bug that assign with negative number

birdstorm:
This PR has been merged into #442. Closing.

PR #446, opened 2018-09-09
Fix my decimal bug by adding science notation support

### LINKED ISSUE #445: myDecimal's fromString does not support scientific notation
myDecimal's fromString does not support scientific notation.

affects:
- fromDouble will produce the wrong answer for double inputs whose internal presentation likes scientific notation.

Handora:
1. add scientific notation support to fix fromDouble bugs  
2. add most tests suitable from TIDB and passes all of them
3. fix #445

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
@@ -694,7 +694,7 @@ public class MyDecimal {
     int wordsFracFrom = this.digitsFrac / digitsPerWord;
     int trailingDigitsFrom = this.digitsFrac - wordsFracFrom * digitsPerWord;
     int intSize = wordsInt * wordSize + dig2bytes[leadingDigits];
-    int fracSize = wordsFracFrom * wordSize + dig2bytes[trailingDigits];
+    int fracSize = wordsFrac * wordSize + dig2bytes[trailingDigits];
     int fracSizeFrom = wordsFracFrom * wordSize + dig2bytes[trailingDigitsFrom];
     int originIntSize = intSize;
     int originFracSize = fracSize;
--- a/tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
@@ -728,7 +728,7 @@ public class MyDecimal {
     } else if (intSize > iSizeFrom) {
       for (; intSize > iSizeFrom; ) {
         intSize--;
-        bin[binIdx] = mask;
+        bin[binIdx] = mask & 0xff;
         binIdx++;
       }
     }
@@ -786,7 +786,7 @@ public class MyDecimal {
       int binIdxEnd = originIntSize + originFracSize;
       for (; fracSize > fracSizeFrom && binIdx < binIdxEnd; ) {
         fracSize--;
-        bin[binIdx] = mask;
+        bin[binIdx] = mask & 0xff;
         binIdx++;
       }
     }
--- a/tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
@@ -56,6 +56,9 @@ public class MyDecimal {
   private boolean negative;
   private int[] wordBuf = new int[maxWordBufLen];
 
+  // Integer limit values
+  private static final int maxInt32 = 1 << 31 - 1;
+
   /*
    * Returns total precision of this decimal. Basically, it is sum of digitsInt and digitsFrac. But there
    * are some special cases need to be token care of such as 000.001.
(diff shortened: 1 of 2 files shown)


tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
```diff
@@ -15,7 +15,8 @@
 
 package com.pingcap.tikv.codec;
 
-import com.google.common.annotations.VisibleForTesting;
+import org.spark_project.guava.annotations.VisibleForTesting;
```
birdstorm: Please state the reason of the change in this line.

tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
```diff
@@ -25,7 +26,7 @@
   // how many digits that a word has
   private static final int digitsPerWord = 9;
   // MyDecimal can holds at most 9 words.
-  private static final int wordBufLen = 9;
+  public static int wordBufLen = 9;
```
birdstorm: You should state why you change wordBufLen into public and non-final. You could make a new MyDecimal via copy rather than changing its fields directly.

tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
```diff
@@ -818,6 +913,517 @@ private void writeWord(int[] b, int word, int size, int start) {
...
+
+  // Shift shifts decimal digits in given number (with rounding if it need), shift > 0 means shift to left shift,
+  // shift < 0 means right shift. In fact it is multiplying on 10^shift.
+  public MyDecimalError shiftDecimal(int shift) {
+    MyDecimalError error = MyDecimalError.noError;
+    if (shift == 0) {
+      return MyDecimalError.noError;
+    }
+
+    // digitBegin is index of first non zero digit (all indexes from 0).
```
birdstorm: indexes -> indices

tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
```diff
@@ -440,11 +507,21 @@ private void fromCharArray(char[] str) {
     this.resultFrac = this.digitsFrac;
   }
 
-  // parser a string to a int.
-  private int strToLong(String str) {
+  private class LongWithError {
+    public int result;
```
birdstorm: Since you call it LongWithError, should result be long?

tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
```diff
@@ -87,7 +120,7 @@ public int frac() {
    * @param value an double value
    */
   public void fromDecimal(double value) {
-    String s = Double.toString(value);
+    String s = value + "";
```
birdstorm: It seems that science notation will only occur when you change this line. So is it impossible to meet the assumptions that this science notation bug really exists? Please confirm.

tikv-client/src/test/java/com/pingcap/tikv/codec/MyDecimalTest.java
```diff
@@ -57,14 +135,307 @@ public void fromStringTest() throws Exception {
...
+      try {
+        MyDecimal dec = new MyDecimal();
+        dec.fromString(wt.in);
+        fail();
+      } catch (Exception e) {
+        assertTrue(true);
+      }
+    }
+
+    MyDecimal.wordBufLen = 1;
```
birdstorm: I don't think it makes sense to change this static value of MyDecimal.

tikv-client/src/test/java/com/pingcap/tikv/codec/MyDecimalTest.java
```diff
@@ -57,14 +135,307 @@ public void fromStringTest() throws Exception {
...
+    for (MyDecimalTestStruct t : wrongTests) {
+      MyDecimal dec = new MyDecimal();
+      dec.fromString(t.in);
+      MyDecimal.MyDecimalError err = dec.shiftDecimal(t.scale);
+      assertEquals(t.error, err);
+      String result = dec.toString();
+      assertEquals(t.out, result);
+    }
+
+    MyDecimal.wordBufLen = 2;
```
birdstorm: ditto

tikv-client/src/test/java/com/pingcap/tikv/codec/MyDecimalTest.java
```diff
@@ -57,14 +135,307 @@ public void fromStringTest() throws Exception {
...
+    for (MyDecimalTestStruct t : wrongTests) {
+      MyDecimal dec = new MyDecimal();
+      dec.fromString(t.in);
+      MyDecimal.MyDecimalError err = dec.shiftDecimal(t.scale);
+      assertEquals(t.error, err);
+      String result = dec.toString();
+      assertEquals(t.out, result);
+    }
+
+    MyDecimal.wordBufLen = MyDecimal.maxWordBufLen;
```
birdstorm: ditto

birdstorm:
Close this PR since the case it fixes is only available in tests. Logically, user calls `fromString` only when writing `BigDecimal.toPlainString`, which won't encounter scientific notation.

PR #448, opened 2018-09-14
Exclude conflicting jackson version introduced by typesafe.play

birdstorm:


zanmato1984:
Not enough. You may need to exclude jackson-annotations and jackson-databind as well.

birdstorm:
This is not an alternative solution, closing.

PR #449, opened 2018-09-14
Resolve conflicting jackson version introduced by typesafe.play

birdstorm:
Play API was introduced for debugTools, but it also brings conflicting jackson version consequently.

As a matter of fact, Spark uses json4s for its JSON type support, which we can simply make use of. 

This PR replaces the original play API with json4s provided and does not introduce new dependencies.

@zanmato PTAL


zanmato1984:
LGTM

PR #458, opened 2018-10-11
Use case class for execution commands

birdstorm:
By using case class for execution commands, it will not affect other projects depending on it.


zanmato1984:
LGTM

PR #460, opened 2018-10-12
[TISPARK-51] Fix broken pytispark

birdstorm:
According to [SPARK-25503](https://issues.apache.org/jira/browse/SPARK-25003), pyspark does not correctly read extension information in config.

Fix by adding TiStrategy directly into TiContext.

Fix #460

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ https://github.com/pingcap/tispark/tree/master/tikv-client
 
 
 ## Quick Start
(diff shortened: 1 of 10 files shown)


zanmato1984:
Despite being used in pyspark (where there is no extension entry point), there might exists two `TiExtensions` instances - one in the companion object instantiated during `TiContext` construction and the other instantiated via standard Spark extension loading. Right?

Does it mean that the extended parser/rule/strategy will be added twice?

birdstorm:
@ilovesoup PTAL

PR #475, opened 2018-11-01
Add resolve lock implementations

Handora:
What this PR do
1. finish resolve locks and add tests
2. fix a bug of TiRegion.count disobey its meanings
3. fix a bug of several pd tso request returns the same value
4. Decoupling the client api and sender api for better modularity
5. refresh the region and continue instead of throwing requests for region errors

What the PR lefts
1. batch get multilevel implementations -> like what tidb does
2. streaming coprocessor error handling and resolve locks
3. modular implementation for batchGet and Scan


tikv-client/src/main/java/com/pingcap/tikv/AbstractGRPCClient.java
```diff
@@ -52,7 +52,7 @@ public TiConfiguration getConf() {
   }
 
   // TODO: Seems a little bit messy for lambda part
-  protected <ReqT, RespT> RespT callWithRetry(BackOffer backOffer,
```
Handora: Because I decouple the client and sender, this method should be public for use

tikv-client/src/main/java/com/pingcap/tikv/region/RegionSender.java
```diff
@@ -0,0 +1,97 @@
...
+    return blockingStub.withDeadlineAfter(getConf().getTimeout(), getConf().getTimeoutUnit());
+  }
+
+  @Override
+  protected TikvStub getAsyncStub() {
+    return asyncStub.withDeadlineAfter(getConf().getTimeout(), getConf().getTimeoutUnit());
+  }
+
+  @Override
+  public void close() throws Exception {
```
zhexuany: Do we need to release any resource?

tikv-client/src/main/java/com/pingcap/tikv/region/RegionSender.java
```diff
@@ -0,0 +1,97 @@
...
+
+import com.pingcap.tikv.AbstractGRPCClient;
+import com.pingcap.tikv.TiSession;
+import com.pingcap.tikv.kvproto.Metapb.Store;
+import com.pingcap.tikv.kvproto.TikvGrpc;
+import com.pingcap.tikv.kvproto.TikvGrpc.TikvBlockingStub;
+import com.pingcap.tikv.kvproto.TikvGrpc.TikvStub;
+import io.grpc.ManagedChannel;
+
+public class RegionSender extends AbstractGRPCClient<TikvBlockingStub, TikvStub> implements RegionErrorReceiver  {
```
ilovesoup: What does this class exactly do? From its name, its a sender; but it does not actually have any send method.

tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
```diff
@@ -110,17 +142,17 @@ public void rawPut(BackOffer backOffer, ByteString key, ByteString value, Contex
 
     KVErrorHandler<RawPutResponse> handler =
         new KVErrorHandler<>(
-            regionManager, this, region, resp -> resp.hasRegionError() ? resp.getRegionError() : null);
-    RawPutResponse resp = callWithRetry(backOffer, TikvGrpc.METHOD_RAW_PUT, factory, handler);
+            regionManager, sender, region, resp -> resp.hasRegionError() ? resp.getRegionError() : null);
```
ilovesoup: I don't quite see the benefit of passing another class instead of "this".

tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
```diff
@@ -190,25 +239,51 @@ private ByteString getHelper(GetResponse resp) {
...
-            regionManager, this, region, resp -> resp.hasRegionError() ? resp.getRegionError() : null);
-    ScanResponse resp = callWithRetry(backOffer, TikvGrpc.METHOD_KV_SCAN, request, handler);
-    return scanHelper(resp);
+            regionManager, sender, region, resp -> resp.hasRegionError() ? resp.getRegionError() : null);
+    ScanResponse resp = sender.callWithRetry(backOffer, TikvGrpc.METHOD_KV_SCAN, request, handler);
+    return scanHelper(resp, backOffer);
   }
 
-  private List<KvPair> scanHelper(ScanResponse resp) {
+  // same as BatchGet Helper, can we optimize it?
```
ilovesoup: If logic is largely the same, please do.

tikv-client/src/main/java/com/pingcap/tikv/region/RegionSender.java
```diff
@@ -0,0 +1,97 @@
...
+  @Override
+  public boolean onNotLeader(Store newStore) {
+    if (logger.isDebugEnabled()) {
+      logger.debug(region + ", new leader = " + newStore.getId());
+    }
+    TiRegion cachedRegion = regionManager.getRegionById(region.getId());
+    // When switch leader fails or the region changed its key range,
+    // it would be necessary to re-split task's key range for new region.
+    if (!region.switchPeer(newStore.getId()) ||
+        !region.getStartKey().equals(cachedRegion.getStartKey()) ||
```
zhexuany:
``` !region.getStartKey().equals(cachedRegion.getStartKey()) || 
        !region.getEndKey().equals(cachedRegion.getEndKey())``` 
can wrap into a function. 

isStartEndKeyValid() boolean

File: tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java [hunk 7]
```diff
@@ -262,6 +350,7 @@ private ByteString getHelper(GetResponse resp) {
     return null;
   }
 
+  // TODO: wait for future fix
```
ilovesoup: Please explain what need to be fixed in the future.

tikv-client/src/main/java/com/pingcap/tikv/region/RegionSender.java
```diff
@@ -0,0 +1,97 @@
...
+
+  @Override
+  public void close() throws Exception {
+  }
+
+  /**
+   * onNotLeader deals with NotLeaderError and returns whether re-splitting key range is needed
+   *
+   * @param newStore the new store presented by NotLeader Error
+   * @return false when re-split is needed.
```
zhexuany: what if this function returns true?  Any operation needed for this case?

tikv-client/src/main/java/com/pingcap/tikv/region/TiRegion.java
```diff
@@ -124,8 +141,8 @@ boolean switchPeer(long leaderStoreID) {
   }
 
   public boolean contains(ByteString key) {
-    return meta.getStartKey().equals(key)
-        && (meta.getEndKey().equals(key) || meta.getEndKey().isEmpty());
+    return (meta.getStartKey().toStringUtf8().compareTo(key.toStringUtf8()) <=0 || meta.getStartKey().isEmpty())
```
ilovesoup: This looks wrong to me. Once bytes are decoded as UTF8, essentially the content changed and hence sort order changed as well. Use FastByteComparisons instead.

tikv-client/src/main/java/com/pingcap/tikv/txn/Lock.java
```diff
@@ -0,0 +1,68 @@
...
+    key = l.getKey();
+    primary = l.getPrimaryLock();
+    ttl = l.getLockTtl() == 0 ? defaultLockTTL : l.getLockTtl();
+  }
+
+  public long getTxnID() {
+    return txnID;
+  }
+
+  public void setTxnID(long txnID) {
```
ilovesoup: I searched around, there is almost no need to change the status of Lock once created. If so, make this an immutable object. Usually immutable is preferred since you don't need to bother hashcode change, reasoning field change, multi-threading and etc.

tikv-client/src/main/java/com/pingcap/tikv/txn/LockResolver.java
```diff
@@ -0,0 +1,226 @@
...
+import static com.pingcap.tikv.util.BackOffFunction.BackOffFuncType.BoRegionMiss;
+
+// LockResolver resolves locks and also caches resolved txn status.
+public class LockResolver {
+  // ResolvedCacheSize is max number of cached txn status.
+  private static final long ResolvedCacheSize = 2048;
+  // By default, locks after 3000ms is considered unusual (the client created the
+  // lock might be dead). Other client may cleanup this kind of lock.
+  // For locks created recently, we will do backoff and retry.
+  private static long defaultLockTTL = 3000;
```
ilovesoup: If it's default, why not final. All above fields looks constant for me.

tikv-client/src/main/java/com/pingcap/tikv/util/Oracle.java
```diff
@@ -0,0 +1,15 @@
+package com.pingcap.tikv.util;
+
+final public class Oracle {
+  private static final long physicalShiftBits = 18;
+
+  public static boolean IsExpired(long lockTS, long ttl) {
```
ilovesoup: Why you have method start with uppercase?

tikv-client/src/main/java/com/pingcap/tikv/util/Oracle.java
```diff
@@ -0,0 +1,15 @@
+package com.pingcap.tikv.util;
+
+final public class Oracle {
```
ilovesoup: Usually utility class we named it like an utility. For example, OracleUtil or tsutil.

tikv-client/src/main/java/com/pingcap/tikv/util/Oracle.java
zhexuany: I am guessing `IsExpired` is the name of the method in tidb. uppercase is how golang let outside package access the method.

tikv-client/src/main/java/com/pingcap/tikv/util/Oracle.java
zhexuany: I think tsoutil is better name.

tikv-client/src/main/java/com/pingcap/tikv/txn/LockResolver.java
```diff
@@ -0,0 +1,226 @@
...
+  private final RegionSender sender;
+
+  public LockResolver(RegionSender sender) {
+    resolved = new HashMap<>();
+    recentResolved = new LinkedList<>();
+    readWriteLock = new ReentrantReadWriteLock();
+    this.sender = sender;
+  }
+
+  private void SaveResolved(long txnID, long status) {
```
zhexuany: s/SaveResolved/saveResolved

tikv-client/src/main/java/com/pingcap/tikv/txn/LockResolver.java
```diff
@@ -0,0 +1,226 @@
...
+  private static long maxLockTTL = 120000;
+  // ttl = ttlFactor * sqrt(writeSizeInMiB)
+  private static long ttlFactor = 6000;
+  private static final Logger logger = Logger.getLogger(LockResolver.class);
+
+  private final ReadWriteLock readWriteLock;
+  // Note: Because the internal of long is same as unsigned_long
+  // and Txn id are never changed. Be careful to compare between ts
+
+  // the mapping is from TxnId -> TxnStatus
```
zhexuany: maybe better to say @code resovled's key is txnid and its value is txnStatus.

tikv-client/src/main/java/com/pingcap/tikv/txn/LockResolver.java
```diff
@@ -0,0 +1,226 @@
...
+  // ResolveLocks tries to resolve Locks. The resolving process is in 3 steps:
+  // 1) Use the `lockTTL` to pick up all expired locks. Only locks that are too
+  //    old are considered orphan locks and will be handled later. If all locks
+  //    are expired then all locks will be resolved so the returned `ok` will be
+  //    true, otherwise caller should sleep a while before retry.
+  // 2) For each lock, query the primary key to get txn(which left the lock)'s
+  //    commit status.
+  // 3) Send `ResolveLock` cmd to the lock's region to resolve all locks belong to
+  //    the same transaction.
+  public boolean ResolveLocks(BackOffer bo, List<Lock> locks) {
```
zhexuany: s/ResolveLocks/resolveLocks

tikv-client/src/main/java/com/pingcap/tikv/region/RegionSender.java
Handora: Absolutely no, while we need the sender to inherit the AbstractGrpcClient.
Now, I simply remove the `sender` and add `lockResolverClient` and this code is not necessary

Handora:
./run-all-tests

Handora:
./run-all-tests

Handora:
@zhexuany @birdstorm @ilovesoup 
Hi all, I have pushed the test and change the documentation.

tikv-client/src/main/java/com/pingcap/tikv/txn/LockResolverClient.java
```diff
@@ -0,0 +1,288 @@
...
+
+      saveResolved(txnID, status);
+      return status;
+    }
+  }
+
+  // ResolveLocks tries to resolve Locks. The resolving process is in 3 steps:
+  // 1) Use the `lockTTL` to pick up all expired locks. Only locks that are old
+  //    enough are considered orphan locks and will be handled later. If all locks
+  //    are expired then all locks will be resolved so tru will be returned, otherwise
```
zhexuany: so tru?

tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
```diff
@@ -89,19 +93,49 @@ public int getValue() {
...
+          Lock lock = new Lock(resp.getError().getLocked());
+          boolean ok = lockResolverClient.resolveLocks(backOffer, new ArrayList<>(Arrays.asList(lock)));
+          if (!ok) {
+            // if not resolve all locks, we wait and retry
+            backOffer.doBackOff(BoTxnLockFast, new KeyException((resp.getError().getLocked().toString())));
+          }
+
+          continue;
+        } else {
+          // retryable or abort
```
zhexuany: retriable？

tikv-client/src/main/java/com/pingcap/tikv/region/TiRegion.java
```diff
@@ -124,8 +142,11 @@ boolean switchPeer(long leaderStoreID) {
   }
 
   public boolean contains(ByteString key) {
-    return meta.getStartKey().equals(key)
-        && (meta.getEndKey().equals(key) || meta.getEndKey().isEmpty());
+
+    return (meta.getStartKey().isEmpty() || FastByteComparisons.compareTo(meta.getStartKey().toByteArray(),
+        0, meta.getStartKey().size(), key.toByteArray(), 0, key.size()) < 0)
```
zhexuany: change to 
```
meta.getStartKey().isEmpty() || FastByteComparisons.compareTo(meta.getStartKey().toByteArray(),
        0, meta.getStartKey().size(), key.toByteArray(), 0, key.size()) <=0
```
It shoul be left closed, right open.

tikv-client/src/main/java/com/pingcap/tikv/txn/LockResolverClient.java
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  /**
+   * onNotLeader deals with NotLeaderError and returns whether re-splitting key range is needed
+   *
+   * @param newStore the new store presented by NotLeader Error
+   * @return false when re-split is needed.
+   */
+  @Override
+  public boolean onNotLeader(Store newStore) {
```
zhexuany:
```
asyncStub
blockingStub
```
and etc can be moved into `AbstractGrpcClient`. In this way, we can also move `onNotLeader` to `AbstractGrpcClient` to avoid duplicated code.

tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
Handora: No, this means that the txn is retriable or aborted, we should instead rerunning this txn. I will add more comments.

Handora:
./run-all-tests

core/src/test/scala/org/apache/spark/sql/txn/TxnTestSuite.scala
```diff
@@ -0,0 +1,234 @@
...
+      }
+    }
+
+  /**
+   * get Spark query thread using q2 query, a long running transaction
+   *
+   * @param i the query number
+   * @return thread
+   */
+  protected def secondQueryThread(i: Int): Thread =
```
birdstorm: It is the same as `firstQueryThread` except for its query string, isn't it? You should merge them to avoid duplicate code

core/src/test/scala/org/apache/spark/sql/txn/TxnTestSuite.scala
```diff
@@ -0,0 +1,234 @@
...
+      }
+    }
+
+  /**
+   * get Tidb txn thread using 100 transactions(A => B, A - x and B + x)
+   *
+   * @param i the query number
+   * @return thread
+   */
+  protected def secondTxnThread(i: Int): Thread =
```
birdstorm: same as above

core/src/test/scala/org/apache/spark/sql/txn/TxnTestSuite.scala
```diff
@@ -0,0 +1,234 @@
...
+
+  /**
+   * query to tidb with jdbc and txn style
+   *
+   * @param query all querys run in the txn
+   * @param wait whether wait every 2 seconds
+   * @return Unit
+   */
+  protected def queryTIDBTxn(query: Seq[String], wait: Boolean): Unit = {
+    val conn = DriverManager.getConnection(jdbcUrl, "root", "")
```
birdstorm: you might want to wrap this getConnection inside `SharedSQLContext` because users should be able to configure username and password via `properties`. For instance, calling `createJDBCConnection()` will create a new Connection using `DriverManager.getConnection(jdbcUrl, jdbcUsername, jdbcPassword)` with code from 
https://github.com/pingcap/tispark/blob/de938f0037863bb6e36611d20fa2c3dc8de5b7ab/core/src/test/scala/org/apache/spark/sql/test/SharedSQLContext.scala#L171-L173

core/src/test/scala/org/apache/spark/sql/txn/TxnTestSuite.scala
```diff
@@ -0,0 +1,234 @@
...
+   * get Tidb txn thread using 2 transactions(A => B, A - x and B + x)
+   *
+   * @param i the query number
+   * @return thread
+   */
+  protected def firstTxnThread(i: Int): Thread =
+    new Thread {
+      override def run {
+        var ok = true
+        while (ok) {
```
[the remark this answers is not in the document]
birdstorm:
```suggestion
        while (
          try {
            val num = rnd.nextInt(600).toString
            val id1 = (1 + rnd.nextInt(150)).toString
            val id2 = (1 + rnd.nextInt(150)).toString
            val querys = Seq[String](
              giveString.replace("$1", num).replace("$2", id1),
              getString.replace("$1", num).replace("$2", id2)
            )
            queryTIDBTxn(querys, wait = true)
            logger.info("txn1 " + i.toString + " success!")
            false
          } catch {
            case _: SQLException =>
              Thread.sleep(1000 + rnd.nextInt(3000))
              true
          }) {}
```

tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
```diff
@@ -190,25 +250,54 @@ private ByteString getHelper(GetResponse resp) {
...
-    return scanHelper(resp);
+    return scanHelper(resp, backOffer);
   }
 
-  private List<KvPair> scanHelper(ScanResponse resp) {
+  // TODO: remove helper and change to while style
+  // needs to be fixed as batchGet
+  // which we shoule retry not throw
+  // exception
+  private List<KvPair> scanHelper(ScanResponse resp, BackOffer bo) {
```
birdstorm: You are using the same logic as `batchGetHelper`. Try to merge them to eliminate redundant code.

tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
```diff
@@ -190,25 +250,54 @@ private ByteString getHelper(GetResponse resp) {
             .setStartKey(startKey)
             .setVersion(version)
             .setKeyOnly(keyOnly)
-            .setLimit(getConf().getScanBatchSize())
+            .setLimit(session.getConf().getScanBatchSize())
```
birdstorm: why should we change it to `session.getConf()`?

tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
```diff
@@ -396,7 +508,7 @@ public boolean onNotLeader(Store newStore) {
   @Override
   public void onStoreNotMatch(Store store) {
     String addressStr = store.getAddress();
-    ManagedChannel channel = getSession().getChannel(addressStr);
+    ManagedChannel channel = session.getChannel(addressStr);
```
birdstorm: what is the reason to change this line?

tikv-client/src/main/java/com/pingcap/tikv/region/TiRegion.java
```diff
@@ -124,8 +142,10 @@ boolean switchPeer(long leaderStoreID) {
   }
 
   public boolean contains(ByteString key) {
-    return meta.getStartKey().equals(key)
-        && (meta.getEndKey().equals(key) || meta.getEndKey().isEmpty());
+    return (meta.getStartKey().isEmpty() || FastByteComparisons.compareTo(meta.getStartKey().toByteArray(),
```
birdstorm: omg this line is not what it should be.
If you are not using this line, you may simply delete it.
A correct version should be: using `getRegionByKey` to identify if it is the same region with what you are asking for.

tikv-client/src/main/java/com/pingcap/tikv/txn/LockResolverClient.java
```diff
@@ -0,0 +1,288 @@
...
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Supplier;
+
+import static com.pingcap.tikv.util.BackOffFunction.BackOffFuncType.BoRegionMiss;
+
+// LockResolver resolves locks and also caches resolved txn status.
+public class LockResolverClient extends AbstractGRPCClient<TikvBlockingStub, TikvStub> implements RegionErrorReceiver {
+  // ResolvedCacheSize is max number of cached txn status.
+  private static final long resolvedTxnCacheSize = 2048;
```
birdstorm: For those variable constants, I recommend you to use snake case rather than camel case.
```suggestion
  private static final long RESOLVED_TXN_CACHE_SIZE = 2048;
```

tikv-client/src/main/java/com/pingcap/tikv/txn/LockResolverClient.java
```diff
@@ -0,0 +1,288 @@
...
+    try {
+      readWriteLock.writeLock().lock();
+      if (resolved.containsKey(txnID)) {
+        return ;
+      }
+
+      resolved.put(txnID, status);
+      recentResolved.addLast(txnID);
+      if (recentResolved.size() > resolvedTxnCacheSize) {
+        Long front = recentResolved.removeLast();
```
birdstorm: Your recentResolved seems to be actually a stack. Why do you need a LinkedList instead?
Also, I think it should be addLast and removeFront according to TiDB's implementation.

tikv-client/src/main/java/com/pingcap/tikv/txn/LockResolverClient.java
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    List<Lock> expiredLocks = new ArrayList<>();
+    for (Lock lock: locks) {
+      if (TsoUtils.isExpired(lock.getTxnID(), lock.getTtl())) {
+        expiredLocks.add(lock);
+      }
+    }
+
+    if (expiredLocks.size() == 0) {
```
birdstorm: use `empty()` instead.

tikv-client/src/main/java/com/pingcap/tikv/txn/LockResolverClient.java
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    // TxnID -> []Region, record resolved Regions.
+    // TODO: Maybe put it in all LockResolverClient and share by all txns.
+    Map<Long, Set<RegionVerID>> cleanTxns = new HashMap<>();
+    for (Lock l: expiredLocks) {
+      Long status = getTxnStatus(bo, l.getTxnID(), l.getPrimary());
+
+      cleanTxns.putIfAbsent(l.getTxnID(), new HashSet<>());
+      Set<RegionVerID> cleanRegion = cleanTxns.get(l.getTxnID());
```
birdstorm:
```suggestion
      Set<RegionVerID> cleanRegion = cleanTxns.computeIfAbsent(l.getTxnID(), k -> new HashSet<>());
```
and delete line 183.

core/src/test/scala/org/apache/spark/sql/txn/TxnTestSuite.scala
Handora: no, secondTxnThread have 100 operations each txn for long running write operation

tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
Handora: I think this may be resolved after you have finished your logic because the Helper needs to be deprecated. And merge them and replace with while loop style.
We can present an issue for it

tikv-client/src/main/java/com/pingcap/tikv/region/TiRegion.java
Handora: No, ZheXuan says it should be what I write

tikv-client/src/main/java/com/pingcap/tikv/txn/LockResolverClient.java
Handora: I think this can be merged in another PR, because this change will invoke a large change of AbstractGrpcClient

Handora:
./run-all-tests

Handora:
./run-all-tests

Handora:
./run-all-tests

Handora:
@ilovesoup @zhexuany @birdstorm

zhexuany:
@ilovesoup PTAL

Handora:
PTAL @ilovesoup

tikv-client/src/main/java/com/pingcap/tikv/txn/Lock.java
```diff
@@ -0,0 +1,52 @@
+/*
+ *
+ * Copyright 2017 PingCAP, Inc.
```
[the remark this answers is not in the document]
birdstorm: I will modify them in another pull request.

PR #480, opened 2018-11-07
Fix PDClient generates incorrect tso request

birdstorm:
Previously we didn't correctly used setCount to specify logical offset, thus it was default to 0 and we might retrieve same timestamp in a short period of time.


disksing:
LGTM

PR #486, opened 2018-11-12
adding partition info in titableinfo

zhexuany:
This is the very first PR to make tispark being able to read partition table. 

What this PR does is let titableinfo being aware of partition. 

Reference:
https://github.com/pingcap/parser/blob/master/model/model.go#L169
https://github.com/pingcap/parser/blob/master/model/model.go#L315
https://github.com/pingcap/parser/blob/master/model/model.go#L339

### DIFF
--- a/tikv-client/src/test/java/com/pingcap/tikv/meta/MetaUtils.java
+++ b/tikv-client/src/test/java/com/pingcap/tikv/meta/MetaUtils.java
@@ -119,8 +119,9 @@ public class MetaUtils {
         name = "Table" + tid;
       }
       return new TiTableInfo(
-          tid, CIStr.newCIStr(name), "", "", pkHandle, columns, indices, "", 0, 0, 0, 0,
-          partitionInfo);
+          tid, CIStr.newCIStr(name), "", "", pkHandle, columns,
+          indices, "", 0, 0, 0
+          , 0, null);
     }
   }
 


zhexuany:
@birdstorm @Handora PTAL

tikv-client/src/test/java/com/pingcap/tikv/meta/MetaUtils.java
```diff
@@ -119,7 +119,9 @@ public TiTableInfo build() {
         name = "Table" + tid;
       }
       return new TiTableInfo(
-          tid, CIStr.newCIStr(name), "", "", pkHandle, columns, indices, "", 0, 0, 0, 0);
+          tid, CIStr.newCIStr(name), "", "", pkHandle, columns,
+          indices, "", 0, 0, 0
+          , 0, null);
```
birdstorm: What does setting partitionInfo to null mean exactly?

tikv-client/src/main/java/com/pingcap/tikv/meta/TiTableInfo.java
```diff
@@ -175,7 +178,7 @@ public TiTableInfo copyTableWithRowId() {
           getId(), CIStr.newCIStr(getName()), getCharset(), getCollate(),
           true, newColumns.build(), getIndices(),
           getComment(), getAutoIncId(), getMaxColumnId(),
-          getMaxIndexId(), getOldSchemaId());
+          getMaxIndexId(), getOldSchemaId(), partitionInfo);
```
birdstorm: Should we add a getter method for partitionInfo?

tikv-client/src/test/java/com/pingcap/tikv/meta/MetaUtils.java
zhexuany: It means the table is a normal table rather than a partition table.

tikv-client/src/main/java/com/pingcap/tikv/meta/TiTableInfo.java
zhexuany: It will be added later when I really need to use it.

PR #490, opened 2018-11-18
Finish generate partition expr

zhexuany:
This is the second PR to support partition read within tispark.


ilovesoup:
I don't quite understand the logic. Would you please take some time explain to me?

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -229,10 +255,60 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
...
 
+  // If we have a partition info such as
+  // partition by year(`purchased`)
+  // p0 less than 1990
+  // p1 less than 2000,
+  // we will get
+  // p0 year(`purchased`) < 1990
+  // p1 year(`purchased`) >= 1990 and year(`purchased`) < 2000
+  private def generatePartitionExpr(pInfo: TiPartitionInfo): TiPartitionExpr = {
+    val parser: SparkSqlParser = new SparkSqlParser(sparkSession.sqlContext.conf)
```
ilovesoup: Why does Spark Parser works here? Don't you need your own tiny parser?

tikv-client/src/main/java/com/pingcap/tikv/meta/TiPartitionDef.java
```diff
@@ -1,28 +1,42 @@
 package com.pingcap.tikv.meta;
```
ilovesoup: Please remember to add copyright to all files.

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
zhexuany: SparkSqlParser is enough here. 
column used in partition expression will never be an alias so we can parse it first and then resolve it later by taking advantage of MetaResolver.resolve signature.

ilovesoup:
Please clarify design in doc and then we go back and review this pr.

PR #491, opened 2018-11-20
fix json parsing

zhexuany:


core/src/test/scala/org/apache/spark/sql/IssueTestSuite.scala
```diff
@@ -19,6 +19,25 @@ import com.pingcap.tispark.TiConfigConst
 import org.apache.spark.sql.functions.{col, sum}
 
 class IssueTestSuite extends BaseTiSparkSuite {
+  test("partition read") {
+    tidbStmt.execute("DROP TABLE IF EXISTS `partition_t`")
```
birdstorm: remember to drop this table in `afterAll()`

birdstorm:
You might need to skip some failed tests because TiDB has not fixed them yet. 

ref: https://github.com/pingcap/tidb/issues/8258

PR #492, opened 2018-11-20
throw exception when meet partition table

zhexuany:
Reading data from the partition table is not supported.


zhexuany:
partition reading without partition pruning is supported. This PR is not necessary.

PR #495, opened 2018-11-22
Make raw partition read work

zhexuany:
This is the third PR to support partition read.

Changes in this PR:
* creating ranges by partition id inserted of table id when partition table is enabled. 
 ** for index part, we only have the local index. The global index does not exist since the only pid was encoded in key. 

This PR only supports full table scan, partition pruning will be fixed in the following PRs.


core/src/test/scala/org/apache/spark/sql/IssueTestSuite.scala
```diff
@@ -34,8 +34,14 @@ class IssueTestSuite extends BaseTiSparkSuite {
...
       """.stripMargin)
+    tidbStmt.execute(("insert into partition_t values (1, `dede1`, 1989-01-01)"))
+    tidbStmt.execute(("insert into partition_t values (2, `dede2`, 1994-01-01)"))
+    tidbStmt.execute(("insert into partition_t values (3, `dede3`, 1996-01-01)"))
+    tidbStmt.execute(("insert into partition_t values (4, `dede4`, 1997-01-01)"))
+    tidbStmt.execute(("insert into partition_t values (5, `dede5`, 2001-01-01)"))
+    tidbStmt.execute(("insert into partition_t values (6, `dede6`, 2006-01-01)"))
     refreshConnections()
-    assert(spark.sql("select * from partition_t").count() == 0)
+    assert(spark.sql("select * from partition_t").count() == 6)
```
birdstorm: should add more test cases concerning filters, groupBy, and joins.

birdstorm:
Would you please add description stating that after this PR, requests concerning partition tables will be equivalent to full table scan plus filters.

tikv-client/src/main/java/com/pingcap/tikv/predicates/ScanAnalyzer.java
```diff
@@ -193,60 +195,122 @@ public ScanPlan buildScan(
         keyRanges, result.getResidualPredicates(), index, cost, isDoubleRead, estimatedRowCount);
   }
 
+  Pair<Key, Key> buildTableScanKeyRangeHelper(long id, IndexRange ir) {
```
ilovesoup: Why package level accessible? Give it a better name also.

PR #498, opened 2018-11-24
Add special sum without null initial value

ilovesoup:
Previously count on empty table break in Spark 2.3.
This is because:
1. Count is rewrite into Sum of count value from coproc.
2. Sum on empty records in SQL is initialized into null.
3. Count is not nullable in SQL and thus when rewritten into Sum nullable flag is carried.
4. Spark's codegen will not guard on null value and only when forced to String (when using show instead of collect) will throw. If return schema is int it will return 0 normally. 
5. Spark 2.3 changed plan of Dataset.show into cast(sum as string) and then NPE encountered.

Now we change Sum into 0 initialized if it is to replace count in TiSpark.

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/tispark/498)


core/src/test/scala/org/apache/spark/sql/IssueTestSuite.scala
```diff
@@ -249,6 +249,17 @@ class IssueTestSuite extends BaseTiSparkSuite {
...
   }
 
+  // https://github.com/pingcap/tispark/issues/496
+  test("NPE when count(1) on empty table") {
+    tidbStmt.execute("DROP TABLE IF EXISTS `tmp_empty_tbl`")
+    tidbStmt.execute(
+      "CREATE TABLE `tmp_empty_tbl` (`c1` varchar(20))"
+    )
+    refreshConnections()
+    judge("select count(1) from tmp_debug")
```
zhexuany: table name is wrong.

birdstorm:
LGTM if CI tests are passed

PR #512, opened 2018-12-03
Fix unresolved column name when using table.column

### LINKED ISSUE #511: Cannot resolve column name when specifying table.column
For example: 
`select t.a from t`

birdstorm:
Fix #511, the root cause is missing subquery alias when resolving `UnresolvedRelations`

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
@@ -105,8 +105,7 @@ public class RegionStoreClient extends AbstractGRPCClient<TikvBlockingStub, Tikv
   // lockResolverClient, after implements the
   // write implementation of tispark, we can change
   // it to private
-  @VisibleForTesting
-  public final LockResolverClient lockResolverClient;
+  @VisibleForTesting public final LockResolverClient lockResolverClient;
   private TikvBlockingStub blockingStub;
   private TikvStub asyncStub;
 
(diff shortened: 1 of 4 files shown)


tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
```diff
@@ -105,8 +105,7 @@ public int getValue() {
   // lockResolverClient, after implements the
   // write implementation of tispark, we can change
   // it to private
-  @VisibleForTesting
-  public final LockResolverClient lockResolverClient;
+  @VisibleForTesting public final LockResolverClient lockResolverClient;
```
zhexuany: please undo this change

ilovesoup:
Please give more description of your change to ease your reviewer :) @birdstorm

tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
birdstorm: Why，it is formatted this way.

birdstorm: When you add an annotation before a variable/constant, google format state that they should be in the same line.

birdstorm:
@ilovesoup done

tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
zhexuany: Interesting - -

PR #517, opened 2018-12-06
Support describe table command without primary key information

birdstorm:
usage: `spark.sql("desc t").show()`

Support for primary key will be done in another PR.


birdstorm:
PTAL @ilovesoup @zhexuany

core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
```diff
@@ -0,0 +1,137 @@
...
+      } else {
+        Row(database, tableName, isTemp)
+      }
+    }
+  }
+}
+
+case class TiDescribeTablesCommand(tiContext: TiContext, delegate: DescribeTableCommand)
+    extends TiCommand(delegate) {
+  override val output: Seq[Attribute] = Seq(
```
zanmato1984: Why not simply use `delegate`'s output?

core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
```diff
@@ -0,0 +1,137 @@
...
+      new MetadataBuilder().putString("comment", "whether the column is nullable").build()
+    )(),
+    AttributeReference(
+      "comment",
+      StringType,
+      nullable = true,
+      new MetadataBuilder().putString("comment", "comment of the column").build()
+    )()
+  )
+  override def run(sparkSession: SparkSession): Seq[Row] =
```
zanmato1984: Style: missing separating blank line.

core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
birdstorm: I added nullable column in describe table command :) I would like to add more attributes that is not allowed in Hive but promising in TiDB~

birdstorm: I added nullable column in describe table command :) I would like to add more attributes that is not allowed in Hive but promising in TiDB~

core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
birdstorm: I will fix it after tests are complete.

zanmato1984:
LGTM

core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
```diff
@@ -0,0 +1,138 @@
...
+        val information = tiCatalog.getTempViewOrPermanentTableMetadata(tableIdent).simpleString
+        Row(database, tableName, isTemp, s"$information\n")
+      } else {
+        Row(database, tableName, isTemp)
+      }
+    }
+  }
+}
+
+case class TiDescribeTablesCommand(tiContext: TiContext, delegate: DescribeTableCommand)
```
ilovesoup: Please add link to original spark logic in case we need to check change btw versions. Please also from now on track most of the change like this. We can keep a independent docs inside tispark project.

core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
```diff
@@ -0,0 +1,138 @@
...
+
+  private def append(buffer: ArrayBuffer[Row],
+                     column: String,
+                     dataType: String,
+                     comment: String,
+                     nullable: String = ""): Unit =
+    buffer += Row(column, dataType, nullable, comment)
+
+  private def describeFormattedTableInfo(table: CatalogTable, buffer: ArrayBuffer[Row]): Unit = {
+    // So far we only have engine name and primary key for extended information.
```
ilovesoup: What exactly is engine name here? Like BTree?

birdstorm: Oh, that was not yet done in this PR. I will delete them.

core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
birdstorm: sure

Issue #522, opened 2018-12-11
Depend on TiKV/Client-Java


birdstorm:
Contains the following issues:

- [ ] Move `GRPC-ProtoBuf` out to resolve `Guava` conflict with `Spark`
- [ ] Depend on `TiKV/Client-Java` in order to use shaded version in `test` phase
- [ ] Rename group-id to `org.tikv`

birdstorm:
postpone after we merged batch-write.

PR #523, opened 2018-12-11
Separate raw client and txn client

birdstorm:
For #522 

separates raw client and txn client to https://github.com/tikv/client-java


ilovesoup:
I will close it for now since we already had a separation.

PR #525, opened 2018-12-12
Resolve multi session problem

### LINKED ISSUE #521: Support multi-session Spark Jobs which may cause tempViews not found among sessions
Relates to #493 

The issue relates to the singleton usage of TiContext when injecting Extensions into session. If the session was not shared in a job, the TiContext maintains the first SparkSession the job creates, and results in catalog incorrect for all other sessions except the first one.

birdstorm:
Fix #521


birdstorm:
We should redesign TiExtensions in order to make it stateless so that it can cooperate with newSession().

PR #526, opened 2018-12-14
support partition read with partition pruning 

zhexuany:
To be continued.


zhexuany:
It is not finished yet but can read date from tikv with partition table enabled. Also, parition pruning is working but need more attention to some corner case.

core/src/test/scala/org/apache/spark/sql/IssueTestSuite.scala
```diff
@@ -33,6 +33,73 @@ class IssueTestSuite extends BaseTiSparkSuite {
...
+    // extract is not supported
+    // https://dev.mysql.com/doc/refman/5.7/en/date-and-time-functions.html#function_extract
+    assert(spark.sql("SELECT floor(-0.1)").count() == 1)
+    assert(spark.sql("SELECT hour('2009-07-30 12:58:59')").count() == 1)
+    // microsecond is not supported
+    assert(spark.sql("SELECT minute('2009-07-30 12:58:59')").count() == 1)
+    assert(spark.sql("SELECT 2 mod 1.8").count() == 1)
+    assert(spark.sql("SELECT month('2016-07-30')").count() == 1)
+    assert(spark.sql("SELECT quarter('2016-08-31')").count() == 1)
+    assert(spark.sql("SELECT second('2009-07-30 12:58:59')").count() == 1)
```
birdstorm: do we need to check if answer is correct?

core/src/main/scala/org/apache/spark/sql/catalyst/expressions/PartExprFunction.scala
```diff
@@ -0,0 +1,338 @@
...
+
+  @transient private lazy val c = {
+    val c = Calendar.getInstance(DateTimeUtils.getTimeZone("UTC"))
+    c.setFirstDayOfWeek(Calendar.MONDAY)
+    c.setMinimalDaysInFirstWeek(4)
+    c
+  }
+
+  private def adjustCalendarByMode(mode: Int): Unit =
+    mode match {
```
birdstorm: I think we can add a link to mysql doc for these modes.

pom.xml
```diff
@@ -172,6 +172,11 @@
             </plugins>
         </pluginManagement>
         <plugins>
+          <plugin>
+            <groupId>org.ensime.maven.plugins</groupId>
+            <artifactId>ensime-maven</artifactId>
```
birdstorm: are they necessary?

tikv-client/src/main/java/com/pingcap/tikv/meta/TiPartitionInfo.java
```diff
@@ -45,13 +46,41 @@ public TiPartitionInfo(
...
 
+  private PartitionType toPartTp(long type) {
+    if (type == 0L) {
+      return PartitionType.RangePartition;
+    } else if (type == 1L) {
+      return PartitionType.HashPartition;
+    } else if (type == 2L) {
+      return PartitionType.ListPartition;
+    }
+    // only allow range, hash, range partition.
```
birdstorm: list?

tikv-client/src/main/java/com/pingcap/tikv/types/SetType.java
```diff
@@ -49,7 +49,7 @@ protected Object decodeNotNull(int flag, CodecDataInput cdi) {
   /** {@inheritDoc} */
   @Override
   protected void encodeKey(CodecDataOutput cdo, Object value) {
-    throw new UnsupportedTypeException("Set type not supported");
+    throw new UnsupportedTypeException("Set type notTiPartitionInfo.java supported");
```
birdstorm: is this a typo?

zhexuany: Yes. It is a typo.

.gitignore
```diff
@@ -71,3 +71,6 @@ python/pytispark.egg-info/
 
 # ignore spark warehouse
 spark-warehouse
+
+.ensime
+.ensime_cache
```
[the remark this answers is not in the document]
zhexuany: It is for emacs scala config.

ilovesoup:
> To be continued.

What else you need to do? Would you specify it instead of leaving TBC.

.gitignore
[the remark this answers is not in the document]
ilovesoup: I think it would be fine to leave those here.

core/src/main/scala/com/pingcap/tispark/PartitionExpr.scala
```diff
@@ -0,0 +1,205 @@
...
+case class PartitionExpr(partitionExprs: List[Expression],
+                         locateExprs: List[Expression],
+                         partitionExpr: Expression,
+                         col: String) {
+  val ranges = partitionExprs
+  val partExpr = partitionExpr
+  val upperBound = locateExprs
+  val columnName = col
+
+  private def buildExprRange(e: Expression): Range[AsOrdered[Long]] = {
```
ilovesoup: Why this cannot be in with ScanAnalyzer class? This should be with similar logic.

pom.xml
ilovesoup: Maybe this should move to maven profile.

core/src/main/scala/com/pingcap/tispark/PartitionExpr.scala
```diff
@@ -0,0 +1,205 @@
...
+        if (!left.isInstanceOf[Literal] && right.isInstanceOf[Literal]) {
+          range = Range.open[AsOrdered[Long]](
+            AsOrdered(right.eval().asInstanceOf[Number].longValue()),
+            AsOrdered(Long.MaxValue)
+          )
+        }
+        if (!right.isInstanceOf[Literal] && left.isInstanceOf[Literal]) {
+          range = Range.open[AsOrdered[Long]](
+            AsOrdered(Long.MinValue),
+            AsOrdered(left.eval().asInstanceOf[Number].longValue())
```
ilovesoup: Why cast to integer here?

core/src/main/scala/com/pingcap/tispark/PartitionExpr.scala
```diff
@@ -0,0 +1,205 @@
...
+          )
+        }
+        range
+      }
+      case LessThan(left, right) => {
```
ilovesoup: Do you have thorough test on all those cases?

core/src/main/scala/org/apache/spark/sql/TiSparkFunctionRegistry.scala
```diff
@@ -0,0 +1,114 @@
+package org.apache.spark.sql
```
ilovesoup: Copyright.

core/src/main/scala/org/apache/spark/sql/TiSparkFunctionRegistry.scala
```diff
@@ -0,0 +1,114 @@
...
+    // misc non-aggregate functions
+    expression[Microsecond]("microsecond"),
+    expression[WeekDay]("weekday"),
+    expression[ToSeconds]("to_seconds"),
+    expression[ToDays]("to_days"),
+    expression[TimeToSec]("time_to_sec"),
+    expression[YearWeek]("yearweek")
+    // TODO: this need import antler 4 into our code base.
+    // the behavior is like the following:
+    // select extract year from '2019-01-10' which is just
```
ilovesoup: Is this syntax supported in create table? If so, you do need our own antlr parser. If it only appears in select, why do we need this?

core/src/main/scala/org/apache/spark/sql/TiSparkFunctionRegistry.scala
```diff
@@ -0,0 +1,114 @@
...
+  )
+
+  /** See usage above. */
+  private def expression[T <: Expression](
+    name: String
+  )(implicit tag: ClassTag[T]): (String, (ExpressionInfo, FunctionBuilder)) = {
+
+    // For `RuntimeReplaceable`, skip the constructor with most arguments, which is the main
+    // constructor and contains non-parameter `child` and should not be used as function builder.
+    val constructors = if (classOf[RuntimeReplaceable].isAssignableFrom(tag.runtimeClass)) {
```
ilovesoup: If this logic comes from Spark, please add comments refers to spark. When we upgrade, we need to check those logic in case bug fix ignored.

ilovesoup:
You need more test before merge. Add test for different types of partition columns, filter ranges and etc.

core/src/main/scala/org/apache/spark/sql/catalyst/expressions/PartExprFunction.scala
```diff
@@ -0,0 +1,340 @@
+package org.apache.spark.sql.catalyst.expressions
```
ilovesoup: copyright

PR #528, opened 2018-12-20
Adding set and enum support

zhexuany:
This PR let spark can read table with set or enum type. 
```
CREATE TABLE `set` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `title` varchar(255) NOT NULL,
  `priority` set('Low','Medium','High') NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin AUTO_INCREMENT=30001
```

For above table, before this PR, spark cannot read. But it can after this PR.

### DIFF
--- a/core/src/test/scala/org/apache/spark/sql/IssueTestSuite.scala
+++ b/core/src/test/scala/org/apache/spark/sql/IssueTestSuite.scala
@@ -19,17 +19,21 @@ import com.pingcap.tispark.TiConfigConst
 import org.apache.spark.sql.functions.{col, sum}
 
 class IssueTestSuite extends BaseTiSparkSuite {
-  test("adding set and enum"){
-    tidbStmt.execute("drop table set_t if exists ")
-    tidbStmt.execute("drop table enum_t if exists")
-    tidbStmt.execute("CREATE TABLE `set_t` (" +
-      "`priority` set('Low','Medium','High') NOT NULL," +
-      " PRIMARY KEY (`id`));")
-    judge("select * from set_t")
-    tidbStmt.execute("CREATE TABLE `enum_t` (" +
-      "`priority` set('Low','Medium','High') NOT NULL," +
-      " PRIMARY KEY (`id`));")
-    judge("select * from enum_t")
+  test("adding set and enum") {
+    tidbStmt.execute("drop table if exists set_t")
+    tidbStmt.execute("drop table if exists enum_t")
+    tidbStmt.execute(
+      "CREATE TABLE `set_t` (" +
+        "`priority` set('Low','Medium','High') NOT NULL);"
+    )
+    tidbStmt.execute("INSERT INTO set_t(priority) VALUES('High');")
+    judge("select * from set_t;")
+    tidbStmt.execute(
+      "CREATE TABLE `enum_t` (" +
+        "`priority` set('Low','Medium','High') NOT NULL);"
+    )
+    tidbStmt.execute("INSERT INTO enum_t(priority) VALUES('High');")
+    judge("select * from enum_t;")
   }
   test("cannot resolve column name when specifying table.column") {
     spark.sql("select full_data_type_table.id_dt from full_data_type_table").explain(true)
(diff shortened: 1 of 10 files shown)


tikv-client/src/main/java/com/pingcap/tikv/types/SetType.java
```diff
@@ -43,25 +39,26 @@ protected SetType(TiColumnInfo.InternalTypeHolder holder) {
...
-    throw new UnsupportedTypeException("Set type not supported");
+    if (flag != Codec.UVARINT_FLAG) throw new TypeException("Invalid IntegerType flag: " + flag);
+    return this.getElems().get((int) IntegerCodec.readUVarLong(cdi) - 2);
   }
 
   /** {@inheritDoc} */
   @Override
   protected void encodeKey(CodecDataOutput cdo, Object value) {
-    throw new UnsupportedTypeException("Set type not supported");
+    IntegerCodec.writeULongFully(cdo, Converter.convertToLong(value), true);
```
ilovesoup: How would you test write encode process? I suppose here you might need to add index and filter pushdown tests for these two types.

tikv-client/src/main/java/com/pingcap/tikv/types/SetType.java
```diff
@@ -43,25 +39,26 @@ protected SetType(TiColumnInfo.InternalTypeHolder holder) {
   /** {@inheritDoc} */
   @Override
   protected Object decodeNotNull(int flag, CodecDataInput cdi) {
-    throw new UnsupportedTypeException("Set type not supported");
+    if (flag != Codec.UVARINT_FLAG) throw new TypeException("Invalid IntegerType flag: " + flag);
+    return this.getElems().get((int) IntegerCodec.readUVarLong(cdi) - 2);
```
ilovesoup: Why minus 2 here? Would you please add comments here?

birdstorm: same question.

zhexuany: Parsing Set logic is different with paring enum logic. I just updated my implementation.

tikv-client/src/main/java/com/pingcap/tikv/types/SetType.java
zhexuany: @ilovesoup set or enum cannot be pushed down to tikv. So encoding logic is not necessary. We can do it later once tikv allows set or enum to be pushed down to tikv.

core/src/test/scala/org/apache/spark/sql/IssueTestSuite.scala
```diff
@@ -19,6 +19,29 @@ import com.pingcap.tispark.TiConfigConst
 import org.apache.spark.sql.functions.{col, sum}
 
 class IssueTestSuite extends BaseTiSparkSuite {
+  test("adding set and enum") {
+    tidbStmt.execute("drop table if exists set_t")
+    tidbStmt.execute("drop table if exists enum_t")
+    tidbStmt.execute(
+      "CREATE TABLE `set_t` (" +
+        "`priority` set('Low','Medium','High') NOT NULL)"
```
birdstorm: I think it would be better if you drop those tables explicitly in afterAll() method.

tikv-client/src/main/java/com/pingcap/tikv/types/EnumType.java
```diff
@@ -41,25 +40,26 @@ protected EnumType(TiColumnInfo.InternalTypeHolder holder) {
   /** {@inheritDoc} */
   @Override
   protected Object decodeNotNull(int flag, CodecDataInput cdi) {
-    throw new UnsupportedTypeException("Enum type not supported");
+    if (flag != Codec.UVARINT_FLAG) throw new TypeException("Invalid IntegerType flag: " + flag);
+    return this.getElems().get((int) IntegerCodec.readUVarLong(cdi) - 1);
```
birdstorm: How about a double check here to prevent NullPointerException

PR #536, opened 2018-12-27
Add spark reference pattern

birdstorm:
Use `CHECK Spark` pattern when we reference Spark classes. So that we ca find the references when we update Spark version quickly.


ilovesoup:
We might need to add a guideline docs for such things.

PR #540, opened 2018-12-29
Add DateCodec tests to achieve full coverage in codec code path

birdstorm:
image
CodeDataInput and CodeDataOutput are implemented by DataInput and DataOutput. So they will not be counted in coverage.


tikv-client/src/main/java/com/pingcap/tikv/codec/CodecDataInput.java
```diff
@@ -36,13 +36,6 @@
       this.count = buf.length;
     }
 
-    public UnSyncByteArrayInputStream(byte buf[], int offset, int length) {
```
birdstorm: unused.

PR #544, opened 2019-01-04
Use tidb as metastore

zhexuany:
In this PR， we add a hive-site.xml template which can be used as an example if user wants to store their metastore in tidb. In addition, to make user happy, a doc is also created to teach user how to set tidb up.


zhexuany:
Why tidb after 2.1 release canot be used please refer to https://github.com/pingcap/tidb/issues/8793

PR #545, opened 2019-01-04
Make ti version usable without selecting database

### LINKED ISSUE #543: cannot use `ti_version` w/o use a database
What did you do? 
executes `spark.sql("select ti_version()").show(false)` in a fresh started spark-shell.

What do you expect to see?
print version info. 

What do you see instead? 
```
scala> spark.sql("select ti_version()").show(false)
19/01/03 17:37:35 WARN ObjectStore: Failed to get database global_temp, returning NoSuchObjectException
org.apache.spark.sql.AnalysisException: Undefined function: 'ti_version'. This function is neither a registered temporary function nor a permanent function registered in the database 'default'.; line 1 pos 7
```

birdstorm:
Fix #543


birdstorm:
@ilovesoup  We do, because `TiResolutionRule` is implied when extension is set. changing it from lazy will make TiContext be initialized in injection state, before we use `use $database` statement.

PR #566, opened 2019-02-13
fix a bug when perform index scan on partition table. 

zhexuany:
Say, we have a schema:
```
CREATE TABLE `pt` (
  `id` int(11) DEFAULT NULL,
  `y` date DEFAULT NULL,
  index `idx_y`(`y`)
PARTITION BY RANGE ( id ) (
  PARTITION p0 VALUES LESS THAN (2),
  PARTITION p1 VALUES LESS THAN (4),
  PARTITION p2 VALUES LESS THAN (6)
);
```
and insert some values:
```
insert into `pt` values(1, '1995-10-10')
insert into `pt` values(2, '1996-10-10')
insert into `pt` values(3, '1997-10-10')
insert into `pt` values(4, '1998-10-10')
```
Before this PR, `select * from pt where y = date'1998-10-10'`yields empty result. After this PR, such query return correct result.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/CoprocessorRDD.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/CoprocessorRDD.scala
@@ -220,30 +220,37 @@ case class RegionTaskExec(child: SparkPlan,
             new ExecutorCompletionService[util.Iterator[TiRow]](session.getThreadPoolForIndexScan)
           var rowIterator: util.Iterator[TiRow] = null
 
-          // After `splitAndSortHandlesByRegion`, ranges in the task are arranged in order
+
+           // After `splitAndSortHandlesByRegion`, ranges in the task are arranged in order
           // TODO: Maybe we can optimize splitAndSortHandlesByRegion if we are sure the handles are in same region?
-          var indexTasks: util.List[RegionTask] = new util.ArrayList[RegionTask]()
-          if (!dagRequest.getTableInfo.isPartitionEnabled) {
-            indexTasks.addAll(
-              RangeSplitter
-                .newSplitter(session.getRegionManager)
-                .splitAndSortHandlesByRegion(
-                  dagRequest.getTableInfo.getId,
-                  new TLongArrayList(handles)
-                )
-            )
-          } else {
-            // when partition table is present, partition id is table id.
-            val partInfo = dagRequest.getPrunedPartInfo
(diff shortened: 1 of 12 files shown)


core/src/main/scala/org/apache/spark/sql/execution/CoprocessorRDD.scala
```diff
@@ -298,16 +301,6 @@ case class RegionTaskExec(child: SparkPlan,
               val handleList = feedBatch()
               numHandles += handleList.size()
               logger.info("Single batch handles size:" + handleList.size())
-              // After `splitAndSortHandlesByRegion`, ranges in the task are arranged in order
```
zhexuany: This part was deleted because it was already initialized before at line 227.

zhexuany:
```
scala> spark.sql("select * from pt where y = date'1995-10-10' and id = 4").explain
== Physical Plan ==
TiSpark RegionTaskExec{downgradeThreshold=10000,downgradeFilter=[Not(IsNull([id])), [[id] EQUAL 4], Not(IsNull([y]))]
+- TiDB HandleRDD{[table: pt] [Index: idx_y] , Columns: [id], [y], Residual Filter: [[y] EQUAL 1995-10-10T00:00:00.000+08:00], PushDown Filter: Not(IsNull([id])), [[id] EQUAL 4], Not(IsNull([y])), KeyRange: [[116,128,0,0,0,0,0,0,82,95,105,128,0,0,0,0,0,0,1,4,25,86,84,0,0,0,0,0], [116,128,0,0,0,0,0,0,82,95,105,128,0,0,0,0,0,0,1,4,25,86,84,0,0,0,0,1])[[116,128,0,0,0,0,0,0,83,95,105,128,0,0,0,0,0,0,1,4,25,86,84,0,0,0,0,0], [116,128,0,0,0,0,0,0,83,95,105,128,0,0,0,0,0,0,1,4,25,86,84,0,0,0,0,1])[[116,128,0,0,0,0,0,0,84,95,105,128,0,0,0,0,0,0,1,4,25,86,84,0,0,0,0,0], [116,128,0,0,0,0,0,0,84,95,105,128,0,0,0,0,0,0,1,4,25,86,84,0,0,0,0,1])}
```

Explain's result is as above.

zhexuany:
/run-all-tests tidb-test=release-2.1 tikv=release-2.1 pd=release-2.1

PR #568, opened 2019-02-14
log load data is enabled or not

zhexuany:
This can help us to quickly address the real problem when ci has a failure.


zhexuany:
Closing due to duplicated.

PR #570, opened 2019-02-18
Simple partition pruning

zhexuany:
Simple means partition expression pruning only works on the column, not a function. 

If we have the following table schema:
```
CREATE TABLE `pt` (
  `id` int(11) DEFAULT NULL,
  `y` date DEFAULT NULL,
  index `idx_y`(`y`)
PARTITION BY RANGE (id ) (
  PARTITION p0 VALUES LESS THAN (2),
  PARTITION p1 VALUES LESS THAN (6),
  PARTITION p2 VALUES LESS THAN (10)
)
```

`select * from pt where id > 3 and id < 6` will only access p1. If partition expression is `year(y)` not `id`, then partition pruning will not work.

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PrunedPartitionBuilder.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PrunedPartitionBuilder.java
@@ -1,0 +1,207 @@
+/*
+ * Copyright 2019 PingCAP, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.pingcap.tikv.expression.visitor;
+
+import com.google.common.collect.Range;
+import com.google.common.collect.RangeSet;
+import com.google.common.collect.TreeRangeSet;
+import com.google.common.primitives.Longs;
+import com.pingcap.tikv.exception.TiExpressionException;
+import com.pingcap.tikv.expression.ColumnRef;
+import com.pingcap.tikv.expression.ComparisonBinaryExpression;
+import com.pingcap.tikv.expression.ComparisonBinaryExpression.NormalizedPredicate;
+import com.pingcap.tikv.expression.Constant;
+import com.pingcap.tikv.expression.Expression;
+import com.pingcap.tikv.expression.IsNull;
+import com.pingcap.tikv.expression.LogicalBinaryExpression;
+import com.pingcap.tikv.expression.Not;
+import com.pingcap.tikv.expression.visitor.CanBePrunedValidator.ContextPartExpr;
+import com.pingcap.tikv.meta.TiPartitionDef;
+import com.pingcap.tikv.meta.TiPartitionInfo;
+import com.pingcap.tikv.meta.TiPartitionInfo.PartitionType;
+import com.pingcap.tikv.meta.TiTableInfo;
+import com.pingcap.tikv.predicates.PredicateUtils;
+import java.util.ArrayList;
+import java.util.List;
+
+public class PrunedPartitionBuilder extends DefaultVisitor<RangeSet<Long>, Void> {
+
+  private static Expression partExpr;
+
+  private static void throwOnError(Expression node) {
+    final String errorFormat = "Unsupported conversion to Range: %s";
+    throw new TiExpressionException(String.format(errorFormat, node));
+  }
+
+  protected RangeSet<Long> process(Expression node, Void context) {
+    throwOnError(node);
+    return null;
+  }
+
+  /**
+   * return false if table cannot be pruning or partition table is not enabled. return true if
+   * partition expression is simple.
+   *
+   * @param tblInfo is table info which is used for creaeting columnRef.
+   * @param filter is a where condition. It must be a cnf and does not contain not or isnull.
+   * @return true if partition pruning can apply on filter.
+   */
+  private static boolean canBePruned(TiTableInfo tblInfo, Expression filter) {
+    if (!tblInfo.isPartitionEnabled()) {
+      return false;
+    }
+
+    if (TiPartitionInfo.toPartType((int) tblInfo.getPartitionInfo().getType())
+        != PartitionType.RangePartition) {
+      return false;
+    }
+
+    // if query is select * from t, then filter will be null.
+    if (filter == null) return false;
+
+    // Ideally, part expr should be parsed by a parser
+    // but it requires a lot work, we chose a brute force way
+    // to workaround.
+    boolean canBePruned = true;
+    partExpr = ColumnRef.create(tblInfo.getPartitionInfo().getExpr(), tblInfo);
+    if (!tblInfo.getPartitionInfo().getExpr().contains("(")) {
+      // skip not(isnull(col)) case
+      canBePruned = CanBePrunedValidator.canBePruned(filter, new ContextPartExpr(partExpr));
+    }
+
+    return canBePruned;
+  }
+
+  private static List<Expression> filterFilters(List<Expression> filters) {
+    List<Expression> filteredFilters = new ArrayList<>();
+    for (Expression expr : filters) {
+      if (expr instanceof Not || expr instanceof IsNull) {
+        continue;
+      }
+      filteredFilters.add(expr);
+    }
+    return filteredFilters;
+  }
+
+  @Override
+  @SuppressWarnings("Duplicates")
+  protected RangeSet<Long> visit(ComparisonBinaryExpression node, Void context) {
+    NormalizedPredicate predicate = node.normalize();
+    RangeSet<Long> ranges = TreeRangeSet.create();
+    if (!predicate.getColumnRef().equals(partExpr)) return TreeRangeSet.<Long>create().complement();
+    Long literal = (Long) predicate.getValue().getValue();
+    switch (predicate.getType()) {
+      case GREATER_THAN:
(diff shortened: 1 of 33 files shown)


tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PrunedPartitionBuilder.java
```diff
@@ -0,0 +1,208 @@
...
+    return filter.accept(this, null);
+  }
+
+  private static List<Expression> generatePartExprs(TiTableInfo tableInfo) {
+    // only support column ref for now. Fn Expr will be supported later.
+    TiPartitionInfo partInfo = tableInfo.getPartitionInfo();
+    List<Expression> partExprs = new ArrayList<>();
+    for (int i = 0; i < partInfo.getDefs().size(); i++) {
+      TiPartitionDef pDef = partInfo.getDefs().get(i);
+      Constant cst = Constant.create(Longs.tryParse(pDef.getLessThan().get(0)));
```
ilovesoup: Again, why long? Even if current stage of TiDB implements long only, your implementation should not hard-code it like this.

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PrunedPartitionBuilder.java
```diff
@@ -0,0 +1,208 @@
...
+      if (i == 0) {
+        partExprs.add(ComparisonBinaryExpression.lessThan(partExpr, cst));
+      } else {
+        Constant previous =
+            Constant.create(Longs.tryParse(partInfo.getDefs().get(i - 1).getLessThan().get(0)));
+        Constant current =
+            Constant.create(Longs.tryParse(partInfo.getDefs().get(i).getLessThan().get(0)));
```
ilovesoup: Would you please explain why get(0)? If you are only to deal with range type partition only, please at least leave room for other type in your code. If this method is just to generate expr for range type, name it like generateRangePartExpr. Also guard on partition type at beginning.

tikv-client/src/main/java/com/pingcap/tikv/meta/TiColumnInfo.java
```diff
@@ -130,7 +130,8 @@ public String getName() {
   }
 
   public boolean matchName(String name) {
-    return this.name.equalsIgnoreCase(name);
+    return this.name.equalsIgnoreCase(name)
+        || String.format("`%s`", this.name).equalsIgnoreCase(name);
```
ilovesoup: When will a column be quoted?

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PrunedPartitionBuilder.java
```diff
@@ -0,0 +1,208 @@
...
+      if (expr instanceof Not || expr instanceof IsNull) {
+        continue;
+      }
+      filteredFilters.add(expr);
+    }
+    return filteredFilters;
+  }
+
+  @Override
+  @SuppressWarnings("Duplicates")
```
ilovesoup: Is it possible to reuse index range builder code?

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PrunedPartitionBuilder.java
```diff
@@ -0,0 +1,208 @@
...
+    for (int i = 0; i < partExprs.size(); i++) {
+      Expression partExpr = partExprs.get(i);
+      RangeSet<Long> partRange = buildPartRange(partExpr);
+      partRange.removeAll(filterRange.complement());
+      if (!partRange.isEmpty()) {
+        // part range is empty indicates this partition can be pruned.
+        pDefs.add(partInfo.getDefs().get(i));
+      }
+    }
+    return new TiPartitionInfo(
```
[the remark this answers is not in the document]
zhexuany: Partition type info is also needed.

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PrunedPartitionBuilder.java
zhexuany: For range partition, only int or function return int can be accepted as partition expression.

tikv-client/src/main/java/com/pingcap/tikv/meta/TiColumnInfo.java
zhexuany:
```
CREATE TABLE `pt` (
  `id` int(11) DEFAULT NULL,
  `y` date DEFAULT NULL,
  KEY `idx_y` (`y`)
PARTITION BY RANGE ( `id` + 1 ) (
  PARTITION p0 VALUES LESS THAN (2),
  PARTITION p1 VALUES LESS THAN (4),
  PARTITION p2 VALUES LESS THAN (6)
)
```

If we have above table schema, `partExpr` in `PartitionInfo` is "`id`". This is just a hard code version. When parser gets merged, I will rewrite this part.

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PrunedPartitionBuilder.java
zhexuany: It is possible. I will work on that.

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PrunedPartitionBuilder.java
zhexuany: It is hardcode version and will be replaced by a parser. The parser can handle this well. I should comment on this. Maybe we should first review parser code?

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PrunedPartitionBuilder.java
```diff
@@ -0,0 +1,208 @@
...
+
+    // if query is select * from t, then filter will be null.
+    if (filter == null) return false;
+
+    // Ideally, part expr should be parsed by a parser
+    // but it requires a lot work, we chose a brute force way
+    // to workaround.
+    boolean canBePruned = true;
+    partExpr = ColumnRef.create(tblInfo.getPartitionInfo().getExpr(), tblInfo);
+    if (!tblInfo.getPartitionInfo().getExpr().contains("(")) {
```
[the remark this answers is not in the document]
zhexuany: It should rebase master once parser code gets merged into master. - -

tikv-client/src/test/java/com/pingcap/tikv/expression/visitor/CanBePrunedValidatorTest.java
```diff
@@ -0,0 +1,74 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+public class CanBePrunedValidatorTest {
+
+  @Test
+  public void canBePrunedOrWithUnrelatedColCase() {
+    TiTableInfo tableInfo = createTable();
+    Expression partExpr = ColumnRef.create("c1", tableInfo);
+    Expression ts = ColumnRef.create("c3", tableInfo);
```
birdstorm: I think they are ColumnRef

tikv-client/src/test/java/com/pingcap/tikv/expression/visitor/CanBePrunedValidatorTest.java
```diff
@@ -0,0 +1,74 @@
+package com.pingcap.tikv.expression.visitor;
+
+import static com.pingcap.tikv.expression.visitor.ExpressionTypeCoercerTest.createTable;
```
birdstorm: Seems a bit strange to use createTable in another test... We are only using c1 and c3 here.

tikv-client/src/test/java/com/pingcap/tikv/expression/visitor/CanBePrunedValidatorTest.java
```diff
@@ -0,0 +1,74 @@
...
+import com.pingcap.tikv.expression.visitor.CanBePrunedValidator.ContextPartExpr;
+import com.pingcap.tikv.meta.TiTableInfo;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class CanBePrunedValidatorTest {
+
+  @Test
+  public void canBePrunedOrWithUnrelatedColCase() {
+    TiTableInfo tableInfo = createTable();
```
birdstorm: do we need to add partition info in createTable?

tikv-client/src/test/java/com/pingcap/tikv/predicates/ScanAnalyzerTest.java
```diff
@@ -93,7 +93,7 @@ public void buildTableScanKeyRangeTest() throws Exception {
     ScanAnalyzer scanAnalyzer = new ScanAnalyzer();
 
     List<Coprocessor.KeyRange> keyRanges =
-        scanAnalyzer.buildTableScanKeyRange(table, irs, table.getPartitionInfo());
+        scanAnalyzer.buildTableScanKeyRange(table, irs, table.getPartitionInfo().getDefs());
```
birdstorm: any test for buildTableScanKeyRange when the table contains PartitionInfo?

tikv-client/src/test/java/com/pingcap/tikv/expression/visitor/CanBePrunedValidatorTest.java
zhexuany: No. This is just to verify the correctness of canBePrunedValidator .

tikv-client/src/test/java/com/pingcap/tikv/predicates/ScanAnalyzerTest.java
zhexuany: We do not need to. We just need to verify the correctness of pruning logic. 

There is no significant difference between the partitioned table and the normal table. If we meet a partitioned table, we only need to replace the table id with partition id.

birdstorm: Then you may comment this :)

PR #574, opened 2019-02-21
remove duplicate function in scala side

zhexuany:


core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -200,43 +200,29 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
...
-    val tiColumnSet: Seq[TiExpression] = attributeSet.toSeq.collect {
+    val tiColumnSeq: Seq[TiExpression] = attributeSet.toSeq.collect {
       case BasicExpression(expr) => expr
     }
     val resolver = new MetaResolver(source.table)
-    val tiColumns: Seq[TiColumnRef] = extractTiColumnRefFromExpressions(tiColumnSet)
+    var tiColumns: mutable.HashSet[TiColumnRef] = mutable.HashSet.empty[TiColumnRef]
+    for (expression <- tiColumnSeq) {
+      val colSetPerExpr = PredicateUtils
+        .extractColumnRefFromExpression(expression.asInstanceOf[com.pingcap.tikv.expression.Expression])
```
birdstorm: `.asInstanceOf[com.pingcap.tikv.expression.Expression]` is useless

zhexuany: DONE.

PR #581, opened 2019-03-01
remove unused class

zhexuany:
This class is not being used by any other class. So I just delete it from our codebase.


zhexuany:
@birdstorm PTAL

PR #587, opened 2019-03-05
Fix cannot initialize pd if using ipv6 address

birdstorm:
Previously: if the pd is using an ipv6 address, HostAndPort cannot resolve address correctly.

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
@@ -15,7 +15,6 @@
 
 package com.pingcap.tikv;
 
-import com.google.common.net.HostAndPort;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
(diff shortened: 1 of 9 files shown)


tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
```diff
@@ -110,17 +110,17 @@ public synchronized RegionManager getRegionManager() {
   public synchronized ManagedChannel getChannel(String addressStr) {
     ManagedChannel channel = connPool.get(addressStr);
     if (channel == null) {
-      HostAndPort address;
+      URI address;
       try {
-        address = HostAndPort.fromString(addressStr);
+        address = URI.create("http://" + addressStr);
```
zhexuany: Maybe you can move it to the field variable?

birdstorm: fixed in #589

PR #591, opened 2019-03-11
Fix incorrect pom in deploy phase

birdstorm:
Fix an incorrect phase setting in pom.


zhexuany:
LGTM if tests passed.

PR #598, opened 2019-03-18
Support for flash-read

hanfei1991:


tikv-client/src/main/java/com/pingcap/tikv/region/TiRegion.java
```diff
@@ -86,6 +87,18 @@ public Peer getLeader() {
     return peer;
   }
 
+  public List<Peer> getLearnerList() {
+    int peerCnt = getMeta().getPeersCount();
+    List<Peer> peers = new ArrayList<Peer>();
+    for (int i = 0; i < peerCnt; i++) {
+      Peer peer = getMeta().getPeers(i);
```
birdstorm: you can use `for (Peer peer: getMeta.getPeersList)`

hanfei1991:
@birdstorm PTAL

PR #599, opened 2019-03-18
TISPARK-3: support range partition pruning 

zhexuany:
Say we have a table schema `t2` as follows

```
create table t2(c1 int, y date) partition by range(c1)(partition p0 values less than 5, partition p1 values less than maxvalue). 
```

```
select * from t2 where c1 > 5
```

will only access partition p1. 

Additionally, when `year` function being part of partition expression is also supported. 

Other expressions rather than column or year will be downgraded to full table scan.


core/src/test/scala/org/apache/spark/sql/PartitionTableSuite.scala
```diff
@@ -15,22 +15,435 @@
...
+            .sql(
+              "select * from pt3 where year(purchased) < 1995"
+            )
+        ).getPrunedParts
+        pDef.size() == 4
+      }
+    )
+  }
+
+  test("adding part pruning test when index is on partitioned column") {
```
birdstorm: Do we need to add a test which partition expression is not a column? e.g., (`a` + 1)

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PartAndFilterExprRewriter.java
```diff
@@ -0,0 +1,104 @@
...
+import com.pingcap.tikv.predicates.PredicateUtils;
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * PartAndFilterExprRewriter takes partition expression as an input. Rewriting rule is based on the
+ * type of partition expression. 1. If partition expression is a columnRef, no rewriting will be
+ * performed. 2. If partition expression is year and the expression to be rewrote in the form of y <
+ * '1995-10-10' then its right hand child will be replaced with "1995". 3. If partition expression
+ * is year and the expression to be rewrote in the form of year(y) < '1995' then its left hand child
```
ilovesoup: year is not added to supported expression list and TiStrategy will not allow it to be pushed down. How did you even tested this case?

core/src/test/scala/org/apache/spark/sql/PartitionTableSuite.scala
```diff
@@ -15,22 +15,435 @@
...
+    )
+
+    assert(
+      {
+        val pDef = extractDAGReq(
+          // for complicated expression, we do not support for now.
+          // this will be improved later.
+          spark
+            .sql(
+              "select * from pt3 where year(purchased) < 1995"
```
ilovesoup: I tried this case and it seems filter cannot pass pushdown test and hence not even reached PartitionPruning code path.

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PartAndFilterExprRewriter.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  @Override
+  public Expression visit(IsNull node, Void context) {
+    return node;
+  }
+
+  @Override
+  public Expression visit(FuncCallExpr node, Void context) {
+    return node.getExpression();
```
ilovesoup: I don't think this is right. Consider it a generic func call method, how can you simply extract argument and return?

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PartAndFilterExprRewriter.java
```diff
@@ -0,0 +1,104 @@
...
+    return new LogicalBinaryExpression(node.getCompType(), left, right);
+  }
+
+  @Override
+  public Expression visit(Not node, Void context) {
+    return node;
+  }
+
+  @Override
+  public Expression visit(IsNull node, Void context) {
```
ilovesoup: As last time review I commented, this class simply inherits from DefaultVisitor and its default behavior for all non-defined node is return null. Consider you have a expression like where col1 > 2000+col2 which will not be const folded by spark optimizer and contains an arithmetic expr, your code will break like this:
```
19/03/21 02:13:53 ERROR SparkSQLDriver: Failed in [select * from ptms where id > 2000+id]
java.lang.NullPointerException: left expression is null
	at java.util.Objects.requireNonNull(Objects.java:228)
	at com.pingcap.tikv.expression.ComparisonBinaryExpression.<init>(ComparisonBinaryExpression.java:103)
```
Would you please kindly tell me why you didn't address this comment last time?

ilovesoup:
I also tested a simple case like this:
```
create table ptms (id int, purchased date DEFAULT NULL) PARTITION BY RANGE (id) (PARTITION p0 VALUES LESS THAN (1995),PARTITION p1 VALUES LESS THAN (2000),PARTITION p2 VALUES LESS THAN (2005),PARTITION p3 VALUES LESS THAN (MAXVALUE));

insert into ptms values(1, '1990-01-01');

insert into ptms values(2, '1991-01-01');

insert into ptms values(3, '1999-01-01');

insert into ptms values(4, '2009-01-01');
```
and a simple 
```
SELECT * FROM ptms;
```
In TiDB client it works but in TiSpark it returns nothing. Neither select all / count(*) works. Neither this branch and your previous partition table read on master works. It seems not related to partition table itself because some table works some not. It might related to your encoding / pushdown logic.

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PartAndFilterExprRewriter.java
zhexuany: It is partition expression, not filter condition. My bad, I should state it more clearly.

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PartAndFilterExprRewriter.java
```diff
@@ -0,0 +1,104 @@
...
+import com.pingcap.tikv.expression.LogicalBinaryExpression;
+import com.pingcap.tikv.expression.Not;
+import com.pingcap.tikv.predicates.PredicateUtils;
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * PartAndFilterExprRewriter takes partition expression as an input. Rewriting rule is based on the
+ * type of partition expression. 1. If partition expression is a columnRef, no rewriting will be
+ * performed. 2. If partition expression is year and the expression to be rewrote in the form of y <
```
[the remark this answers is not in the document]
zhexuany: Yes.

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PartAndFilterExprRewriter.java
zhexuany: You are right. Originally, this code is for Year. After discussion, I change year to funcCallExpr and forget this part.

zhexuany:
I ran the sql you provided and it works fine. We need to discuss them tmw.

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PartAndFilterExprRewriter.java
zhexuany: It caused by the behavior of DefaultVistor and it can be overridden via `process` method.

birdstorm:
> I also tested a simple case like this:
> 
> ```
> create table ptms (id int, purchased date DEFAULT NULL) PARTITION BY RANGE (id) (PARTITION p0 VALUES LESS THAN (1995),PARTITION p1 VALUES LESS THAN (2000),PARTITION p2 VALUES LESS THAN (2005),PARTITION p3 VALUES LESS THAN (MAXVALUE));
> 
> insert into ptms values(1, '1990-01-01');
> 
> insert into ptms values(2, '1991-01-01');
> 
> insert into ptms values(3, '1999-01-01');
> 
> insert into ptms values(4, '2009-01-01');
> ```
> and a simple
> 
> ```
> SELECT * FROM ptms;
> ```
> In TiDB client it works but in TiSpark it returns nothing. Neither select all / count(*) works. Neither this branch and your previous partition table read on master works. It seems not related to partition table itself because some table works some not. It might related to your encoding / pushdown logic.

Cannot reproduce.

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/PartAndFilterExprRewriter.java
```diff
@@ -0,0 +1,106 @@
...
+  private boolean unsupportedPartFnFound;
+
+  PartAndFilterExprRewriter(Expression partExpr) {
+    Objects.requireNonNull(partExpr, "partition expression cannot be null");
+    this.partExpr = partExpr;
+    this.columnRefs = PredicateUtils.extractColumnRefFromExpression(partExpr);
+  }
+
+  private boolean isYear() {
+    return partExpr instanceof FuncCallExpr;
```
ilovesoup: Is this right? At least you should check enum value is year.

zhexuany: Yes. I will add this check.

PR #605, opened 2019-03-25
TISPARK-78: Adjust type mapping relationship

zhexuany:
This PR will maps IntergerType.INT to IntegerType rather than LongType. 

In order to avoid compability issue, we introduce a configuration item. Once you set the value to 1, new code will work.


birdstorm:
@zhexuany How about other types like short and byte? Also, we should update the document as well.

PR #610, opened 2019-03-26
Fix PDClient tests not closing ScheduledExecutionService

birdstorm:
which results in dirty error logs in console.


tikv-client/src/main/java/com/pingcap/tikv/PDClient.java
```diff
@@ -183,12 +182,10 @@ public Store getStore(BackOffer backOffer, long storeId) {
 
   @Override
   public void close() throws InterruptedException {
+    isTerminated = true;
```
zhexuany: `isTerminated ` seems redundant.

birdstorm: done.

PR #617, opened 2019-04-01
improve backoff's code readability

zhexuany:


tikv-client/src/main/java/com/pingcap/tikv/util/BackOffer.java
```diff
@@ -31,26 +31,22 @@
...
-  int splitRegionBackoff = 20000;
-  int commitMaxBackoff = 3000;
-
-  /** 20 seconds */
-  int batchPrewriteBackoff = 20000;
-
-  /** 3 seconds */
-  int batchCommitBackoff = 3000;
+  int seconds = 1000;
+  int copBuildTaskMaxBackoff = 5 * seconds;
```
marsishandsome: it's better to use `final static int COP_BUILD_TASK_MAX_BACK_OFF = ...`

zhexuany: `final static` is useless in an interface type.

marsishandsome: u r right.

marsishandsome:
LGTM

PR #618, opened 2019-04-01
fix pyspark's doc

zhexuany:
Original code snippet does not work.


marsishandsome:
spark's old version also works?

zhexuany:
> spark's old version also works?

It depends on tispark's version.

Issue #621, opened 2019-04-02
ManagedChannel is not closed after we close a PDClient. 


zhexuany:
This was found by runing tikv-client's unit test at my own machine. 

The detailed log is attached as follows:

```
Apr 02, 2019 2:12:35 PM io.grpc.internal.ManagedChannelImpl$ManagedChannelReference cleanQueue
SEVERE: *~*~*~ Channel io.grpc.internal.ManagedChannelImpl-121 for target 127.0.0.1:52769 was not shutdown properly!!! ~*~*~*
    Make sure to call shutdown()/shutdownNow() and awaitTermination().
java.lang.RuntimeException: ManagedChannel allocation site
	at io.grpc.internal.ManagedChannelImpl$ManagedChannelReference.<init>(ManagedChannelImpl.java:991)
	at io.grpc.internal.ManagedChannelImpl.<init>(ManagedChannelImpl.java:421)
	at io.grpc.internal.AbstractManagedChannelImplBuilder.build(AbstractManagedChannelImplBuilder.java:329)
	at com.pingcap.tikv.util.ChannelFactory.lambda$getChannel$0(ChannelFactory.java:49)
	at java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1660)
	at com.pingcap.tikv.util.ChannelFactory.getChannel(ChannelFactory.java:34)
	at com.pingcap.tikv.PDClient.createLeaderWrapper(PDClient.java:278)
	at com.pingcap.tikv.PDClient.switchLeader(PDClient.java:266)
	at com.pingcap.tikv.PDClientTest.testSwitchLeader(PDClientTest.java:78)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
[...]

PR #622, opened 2019-04-02
Fix managed channel is not closed

zhexuany:
channelFactory is not closed after tisession being closed. 

This fixes https://github.com/pingcap/tispark/issues/621

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
@@ -15,29 +15,20 @@
 
 package com.pingcap.tikv;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.pingcap.tikv.catalog.Catalog;
 import com.pingcap.tikv.event.CacheInvalidateEvent;
 import com.pingcap.tikv.meta.TiTimestamp;
-import com.pingcap.tikv.pd.PDUtils;
 import com.pingcap.tikv.region.RegionManager;
 import com.pingcap.tikv.region.RegionStoreClient;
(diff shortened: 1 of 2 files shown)


tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
```diff
@@ -205,10 +173,10 @@ public void close() throws Exception {
     getThreadPoolForTableScan().shutdownNow();
     getThreadPoolForIndexScan().shutdownNow();
     getPDClient().close();
+    channelFactory.close();
```
birdstorm: I think ChannelFactory should be closed before PDClient.

tikv-client/src/main/java/com/pingcap/tikv/util/ChannelFactory.java
```diff
@@ -52,9 +52,14 @@ public ManagedChannel getChannel(String addressStr) {
...
   @Override
   public void close() {
-    for (ManagedChannel ch : connPool.values()) {
-      ch.shutdown();
-    }
+    connPool.forEach(
+        (k, v) -> {
+          v.shutdownNow();
+          try {
+            v.awaitTermination(1, TimeUnit.SECONDS);
```
marsishandsome: why wait for 1 second?

tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
zhexuany: agreed.

tikv-client/src/main/java/com/pingcap/tikv/util/ChannelFactory.java
zhexuany: No particular reason. This is added to make a graceful shutdown.

PR #625, opened 2019-04-02
change ecodeRow's signature

zhexuany:


core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -248,7 +248,7 @@ object TiBatchWrite {
       values.update(i, tiRow.get(i, colDataTypes(i)))
     }
 
-    TableCodec.encodeRow(colDataTypes, colIDs, values, new CodecDataOutput())
+    TableCodec.encodeRow(colDataTypes, colIDs, values)
```
ilovesoup: If you are not sure about this interface you can have a simplified version calling the one with CodecDataOutput.

PR #626, opened 2019-04-03
cleanup json type

zhexuany:


tikv-client/src/main/java/com/pingcap/tikv/types/JsonType.java
```diff
@@ -137,37 +138,26 @@ private JsonElement parseValue(byte type, DataInput di) {
   // * notice use this as a unsigned long
   private long parseUint64(DataInput cdi) {
     byte[] readBuffer = new byte[8];
-    readFully(cdi, readBuffer, 0, 8);
+    readFully(cdi, readBuffer);
+
+    return getaLong(readBuffer);
+  }
 
+  private long getaLong(byte[] readBuffer) {
```
ilovesoup: geta?

tikv-client/src/main/java/com/pingcap/tikv/types/JsonType.java
```diff
@@ -137,37 +138,26 @@ private JsonElement parseValue(byte type, DataInput di) {
   // * notice use this as a unsigned long
   private long parseUint64(DataInput cdi) {
```
ilovesoup: What is the difference btw this one and the one inside CDI?

ilovesoup:
Also this file has no copyright. Please port this change into client project or log an issue.

tikv-client/src/main/java/com/pingcap/tikv/types/JsonType.java
zhexuany: I am not sure. I prefer not to change the logic if we only do a cleanup.

zhexuany:
closed since @birdstorm takes over.

PR #634, opened 2019-04-08
don't import spark's dep

zhexuany:


tikv-client/src/main/java/com/pingcap/tikv/types/Converter.java
```diff
@@ -229,7 +229,7 @@ public static long convertStrToDuration(String value) {
       }
       minute = Integer.parseInt(splitBySemiColon[1]);
       if (splitBySemiColon[2].contains(".")) {
-        String[] splitByDot = splitBySemiColon[2].split(".");
+        String[] splitByDot = splitBySemiColon[2].split("\\.");
```
birdstorm: I think our logic is incorrect here. We may need to fix this.

birdstorm: Also we need to add some tests.

zhexuany: I created an issue https://github.com/pingcap/tispark/issues/635 to trace this. It will be fixed later.

zhexuany:
local ci passed. idc's ci has some problems.

PR #638, opened 2019-04-09
Fix bazel build

zhexuany:
Currently, thee bazel build is broken. This PR only fixes tikv-client's build.  Later we will support bazel build for core.


birdstorm:
add some document for bazel build?

ilovesoup:
I'd rather have this removed if not used wildly and we need constantly maintain this. Give me a reason if you think this should stay. I'm open for any reasons. @zhexuany

birdstorm:
Closed according to discussion.

PR #643, opened 2019-04-10
Fix desc table command fails on hive tables

### LINKED ISSUE #642: Describe Hive table breaks
Due to the schema change of DESC command result.

birdstorm:
Closes #642 
desc table for hive table fails because the schema changes in new desc command


marsishandsome:
what do you mean by `new desc command`?

PR #651, opened 2019-04-15
Fix partition parser bug

zhexuany:
Zero_Decimal is not considered. When it tries to parse "id < 0", it throws an exception.


birdstorm:
lgtm

PR #652, opened 2019-04-15
Adding tidb adapter

zhexuany:
Adding a plugin for datanucelus.


pom.xml
```diff
@@ -120,6 +120,74 @@
...
+            <version>2.7.2</version>
+            <scope>${spark.scope}</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.pingcap.tidb</groupId>
+            <artifactId>tidb-datanucleus-adapter</artifactId>
+            <version>1.0</version>
+            <scope>compile</scope>
+        </dependency>
+    </dependencies>
```
marsishandsome: why add these dependencies?

zhexuany: In this way, we can use the plugin at runtime.

zhexuany:
build

PR #660, opened 2019-04-16
[TISPARK-102] Use reflect to support multi-spark-version rather than profile

birdstorm:


birdstorm:
we may first check if it works locally.

core/src/main/scala/org/apache/spark/sql/execution/CoprocessorRDD.scala
```diff
@@ -57,53 +57,57 @@ case class CoprocessorRDD(output: Seq[Attribute], tiRdd: TiRDD) extends LeafExec
...
-    //      isOrderSensitive: Boolean = false): RDD[U]
-    //
-    // Hereby we use reflection to support different Spark versions.
-    val method = try {
+  // In Spark 2.3.0 the method declaration is:
+  // private[spark] def mapPartitionsWithIndexInternal[U: ClassTag](
+  //      f: (Int, Iterator[T]) => Iterator[U],
+  //      preservesPartitioning: Boolean = false): RDD[U]
+  //
+  // In other Spark 2.3.x versions, the method declaration is:
```
marsishandsome: x > 0

core/src/main/scala/org/apache/spark/sql/execution/CoprocessorRDD.scala
```diff
@@ -57,53 +57,57 @@ case class CoprocessorRDD(output: Seq[Attribute], tiRdd: TiRDD) extends LeafExec
...
+    classOf[RDD[InternalRow]]
+      .getDeclaredMethod(
+        "mapPartitionsWithIndexInternal",
+        classOf[(Int, Iterator[InternalRow]) => Iterator[UnsafeRow]],
+        classOf[Boolean],
+        classOf[Boolean],
+        classOf[ClassTag[UnsafeRow]]
+      )
+  } catch {
+    case _: Throwable =>
```
marsishandsome: 1. it's better to provide a general function to do spark version related reflection
2. you can read spark version from `org.apache.spark.SPARK_VERSION`, and then call the correct reflected function. Do not need to use try catch.

birdstorm: agree.

core/src/main/scala/org/apache/spark/sql/execution/CoprocessorRDD.scala
birdstorm: word `other` excluded 2.3.0

PR #661, opened 2019-04-17
Refractor reflection

birdstorm:


core/src/main/scala/com/pingcap/tispark/utils/ReflectionUtil.scala
```diff
@@ -0,0 +1,99 @@
...
+  // Hereby we use reflection to support different Spark versions.
+  private val mapPartitionsWithIndexInternal: Method = spark_version match {
+    case "2.3.0" | "2.3.1" =>
+      classOf[RDD[InternalRow]].getDeclaredMethod(
+        "mapPartitionsWithIndexInternal",
+        classOf[(Int, Iterator[InternalRow]) => Iterator[UnsafeRow]],
+        classOf[Boolean],
+        classOf[ClassTag[UnsafeRow]]
+      )
+    case _ =>
```
marsishandsome: add some comment like
// >= 2.3.2

birdstorm: ok

PR #663, opened 2019-04-17
Fix missing reflection

birdstorm:
Some reflection missed in CoprocessorRDD.


marsishandsome:
@zhexuany can we add a command `/run-all-regression-tests`,  which will run `mvn test -Pspark=${all supported spark version}`.

currently all supported spark version include:
1. 2.3.0
2. 2.3.1
3. 2.3.2
3. 2.3.3

The command is off default and can be triggered on github PR page, and will also triggered automatically every night.

PR #664, opened 2019-04-17
Modify version info

birdstorm:
Improve version info because we are not using minor versions of Spark.


birdstorm:
@zhexuany i think we can skip integration test for this pr.

PR #665, opened 2019-04-18
Add profile to support spark 2.4.0

birdstorm:
Support Spark 2.4.0 by profile.


birdstorm:
> please run format

done.

core/pom.xml
```diff
@@ -305,7 +309,8 @@
                 <version>0.10_1.5.1</version>
                 <configuration>
                     <sourceDirectories> <!-- (Optional) Paths to source-directories. Overrides ${project.build.sourceDirectory} -->
-                        <param>${project.basedir}/src/main/scala</param>
+                        <sourceDirectory>${project.basedir}/src/main/scala</sourceDirectory>
+                        <sourceDirectory>${project.basedir}/src/main/${profile.source.directory}</sourceDirectory>
```
marsishandsome: shall we add both folder `spark-2.3` and `spark-2.4`?

birdstorm: I'm not sure of that. Both implementation is reasonable.

marsishandsome:
in order to keep compatibility, there are three ways:
1. use different source code in compiling stage (what birdstorm did)
2. compile both version's target classes, but not generate in classpath (still in jar), according to spark's version at runtime, dynamically load the correct classes
3. according to spark's version at runtime, dynamically generate classes and load classes

core/pom.xml
marsishandsome: we should format all the source code

birdstorm:
> * use different source code in compiling stage (what birdstorm did)
> * compile both version's target classes, but not generate in classpath (still in jar), according to spark's version at runtime, dynamically load the correct classes
> * according to spark's version at runtime, dynamically generate classes and load classes

After discussion with @zanmato1984 , I believe by using scheme with profile is a better way to deal with multi-version dependency. We can use reflection that generates classes at runtime for minor versions -- Spark 2.3.x -- and combine with using different source roots for major versions -- Spark 2.x -- so that alternately we will introduce two releases for TiSpark, i.e., tispark_spark2.3 and tispark_spark2.4.

core/pom.xml
birdstorm: so that we shall hardcode all available profiles, and remember to change them when new profiles are provided/modified.

marsishandsome:
> After discussion with @zanmato1984 , I believe using scheme with profile is a better way to deal with multi-version dependency. We can use reflection that generates classes at runtime for minor versions -- Spark 2.3.x -- and combine with using different source roots for major versions -- Spark 2.x -- so that alternately we will introduce two releases for TiSpark, i.e., tispark_spark2.3 and tispark_spark2.4.

sounds good

core/pom.xml
marsishandsome: or ci `/build` command will trigger both `-Pspark2.3` and `-Pspark2.4` profile

birdstorm: It is!

Issue #647, opened 2019-04-12
LockerResolverClient need retry after resolved lock


zhexuany:
According to current LockerResolverClient, it did not retry after resolved lock. It may lead to task failure. We need to address this.

PR #666, opened 2019-04-18
Fix retry when scan meet lock

zhexuany:
This PR closes https://github.com/pingcap/tispark/issues/647

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ScanIterator.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ScanIterator.java
@@ -60,10 +60,10 @@ public class ScanIterator implements Iterator<Kvrpcpb.KvPair> {
   // return false if current cache is loaded or not empty.
   private boolean loadCache() {
     if (endOfScan) {
-      return true;
+      return false;
     }
     if (startKey.isEmpty()) {
-      return true;
+      return false;
     }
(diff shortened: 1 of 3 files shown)


tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ScanIterator.java
```diff
@@ -52,13 +57,13 @@ public ScanIterator(ByteString startKey, TiSession session, long version) {
     this.version = version;
   }
 
-  // return false if current cache is not loaded or empty
+  // return false if current cache is loaded or not empty.
```
birdstorm: I think this method is incorrect in definition.

tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
```diff
@@ -293,43 +296,50 @@ public void rawDelete(BackOffer backOffer, ByteString key, Context context) {
...
+      throw new KeyException(
+          String.format(
+              "scan meet key conflict on primary key %s at commit ts %s",
+              conflict.getPrimary(), conflict.getConflictTs()));
     }
 
-    if (resp.hasRegionError()) {
-      // TODO, we should redo the split and redo the batchGet
-      throw new RegionException(resp.getRegionError());
+    if (!keyError.getRetryable().equals("")) {
```
birdstorm: pretty strange here... perhaps it should be `!keyError.isRetryable`

PR #668, opened 2019-04-19
Fix prefix length is larger than the value used. 

zhexuany:
With the following schema:

```
 CREATE TABLE `idx` (
  `c_id` text DEFAULT NULL,
  KEY `c_id` (`c_id`(10))
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
```
the query `select * from idx where c_id == 'dairy'` will fail. This pr fixes this.

### DIFF
--- a/core/src/test/scala/org/apache/spark/sql/expression/index/PrefixIndexTestSuite.scala
+++ b/core/src/test/scala/org/apache/spark/sql/expression/index/PrefixIndexTestSuite.scala
@@ -57,6 +57,7 @@ class PrefixIndexTestSuite extends BaseTiSparkSuite {
     explainAndRunTest("select a, b from prefix where b LIKE 'ÿ%'", skipJDBC = true)
     explainAndRunTest("select a, b from prefix where b LIKE '%b'")
     explainAndRunTest("select a, b from prefix where b LIKE '%'")
+    explainAndRunTest("select * from prefix where b = 'b'")
   }
 
   // https://github.com/pingcap/tispark/issues/397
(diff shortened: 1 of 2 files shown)


tikv-client/src/main/java/com/pingcap/tikv/types/Converter.java
```diff
@@ -88,9 +88,11 @@ public static String convertToString(Object val) {
   static byte[] convertUtf8ToBytes(Object val, int prefixLength) {
     requireNonNull(val, "val is null");
     if (val instanceof byte[]) {
-      return new String((byte[]) val).substring(0, prefixLength).getBytes(StandardCharsets.UTF_8);
+      return new String((byte[]) val).substring(0, Math.min(((byte[]) val).length, prefixLength))
+          .getBytes(StandardCharsets.UTF_8);
     } else if (val instanceof String) {
-      return ((String) val).substring(0, prefixLength).getBytes(StandardCharsets.UTF_8);
+      return ((String) val).substring(0, Math.min(((String) val).length(), prefixLength))
```
ilovesoup: String valStr = (String)val;
valStr.substring(0, Math.min(valStr.length(), prefixLength))
Make it clean.

core/src/test/scala/org/apache/spark/sql/expression/index/PrefixIndexTestSuite.scala
```diff
@@ -57,6 +57,7 @@ class PrefixIndexTestSuite extends BaseTiSparkSuite {
     explainAndRunTest("select a, b from prefix where b LIKE 'ÿ%'", skipJDBC = true)
     explainAndRunTest("select a, b from prefix where b LIKE '%b'")
     explainAndRunTest("select a, b from prefix where b LIKE '%'")
+    explainAndRunTest("select * from prefix where b = 'b'")
```
birdstorm: I purpose to move this line to Line 38 and add issue link.

tikv-client/src/main/java/com/pingcap/tikv/types/Converter.java
zhexuany: done

PR #669, opened 2019-04-19
fix index double read bug

marsishandsome:


birdstorm:
Close because the issue has been fixed by #670 .

PR #675, opened 2019-04-24
fix some issues about data source api

marsishandsome:
1. mysql connector should packaged in target jar, since tispark will use jdbc to connect to tidb
2. 127.0.0.1 -> pd0 & tidb, user can run example without change a single line (edit /etc/hosts)
3. add parameter prefix `spark.tispark.`, since all parameters in spark-defaults.conf should start with `spark.tispark.`

### DIFF
--- a/core/src/main/scala/com/pingcap/tispark/examples/TiDataSourceExample.scala
+++ b/core/src/main/scala/com/pingcap/tispark/examples/TiDataSourceExample.scala
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-package com.pingcap.tispark
+package com.pingcap.tispark.examples
 
 import org.apache.spark.SparkConf
 import org.apache.spark.sql.{DataFrame, SQLContext, SaveMode, SparkSession}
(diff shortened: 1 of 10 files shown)


marsishandsome:
@birdstorm @zhexuany PTAL

core/src/main/scala/com/pingcap/tispark/examples/TiDataSourceExample.scala
```diff
@@ -202,12 +198,11 @@ object TiDataSourceExample {
                       |OPTIONS (
                       |  database 'tpch_test',
                       |  table 'target_table',
-                      |  tidb.addr '127.0.0.1',
+                      |  tidb.addr 'tidb',
```
zhexuany: this look wired. Ang reason do this

zhexuany:
LGTM

core/src/main/scala/com/pingcap/tispark/examples/TiDataSourceExample.scala
marsishandsome: https://github.com/pingcap/tispark/pull/681

PR #676, opened 2019-04-24
Fix timestamp inconsistent

### LINKED ISSUE #682: Timestamp may be inconsistent in different tables in same query.

birdstorm:
Fix #682 #683 
Timestamp may be inconsistent because ResolutionRule may be applied multiple times.


core/src/main/scala/org/apache/spark/sql/extensions/rules.scala
```diff
@@ -55,22 +56,45 @@ case class TiResolutionRule(getOrCreateTiContext: SparkSession => TiContext)(
         tiSession,
         TiTableReference(dbName, tableName, sizeInBytes),
         meta,
-        Some(ts)
+        Option.apply(ts)
```
zhexuany: Why change to `Option.apply`?

birdstorm: unintended, i will revert it

tikv-client/src/main/java/com/pingcap/tikv/Snapshot.java
```diff
@@ -38,18 +38,17 @@
...
 import org.tikv.kvproto.Kvrpcpb.KvPair;
 import org.tikv.kvproto.Metapb.Store;
 
 public class Snapshot {
   private final TiTimestamp timestamp;
   private final TiSession session;
-  private final TiConfiguration conf;
 
-  public Snapshot(TiTimestamp timestamp, TiSession session) {
+  public Snapshot(@Nonnull TiTimestamp timestamp, TiSession session) {
```
marsishandsome: what if `timestamp` is null?
how can you ensure `timestamp` is not null?

core/src/main/scala/org/apache/spark/sql/extensions/rules.scala
```diff
@@ -55,22 +56,45 @@ case class TiResolutionRule(getOrCreateTiContext: SparkSession => TiContext)(
...
+          condition = condition.transform {
+            case e @ Exists(p, _, _) =>
+              e.copy(plan = p.transform(applyTimestamp))
+            case ls @ ListQuery(p, _, _, _) =>
+              ls.copy(plan = p.transform(applyTimestamp))
+            case s @ ScalarSubquery(p, _, _) =>
+              s.copy(plan = p.transform(applyTimestamp))
+          }
+        )
+      case p @ Project(projectList, _) =>
```
marsishandsome: is there any other cases (except `UnresolvedRelation`, `Filter` and `Project`)?

tikv-client/src/main/java/com/pingcap/tikv/Snapshot.java
birdstorm: yes.

core/src/main/scala/org/apache/spark/sql/extensions/rules.scala
birdstorm: It uses same logic as TiResolutionRule. If it breaks the resolution rule will break first.

birdstorm: It uses same logic as TiResolutionRule. If it breaks the resolution rule will break first.

core/src/main/spark-2.3/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -253,6 +250,19 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
       tableScanPlan
     }
 
+    if (tiColumns.isEmpty) {
+      // we cannot send a request with empty columns
+      if (!scanPlan.isIndexScan) {
```
zhexuany: invert the logic here for better readability?

core/src/main/spark-2.4/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -253,6 +250,19 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
       tableScanPlan
     }
 
+    if (tiColumns.isEmpty) {
```
zhexuany: Since it is used in both spark 2.3 and spark 2.4, maybe we can wrap this logic into a function?

tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
```diff
@@ -369,7 +370,9 @@ else if (col.getColumnInfo().isPrimaryKey() && tableInfo.isPkHandle()) {
       }
 
       dagRequestBuilder.addExecutors(executorBuilder.setTblScan(tblScanBuilder));
+    }
 
+    if (!buildIndexScan || !isDoubleRead()) {
```
zhexuany: It should have some comments.

birdstorm: sure.

PR #677, opened 2019-04-25
change info to debug

zhexuany:


core/src/main/scala/org/apache/spark/sql/execution/CoprocessorRDD.scala
```diff
@@ -371,7 +371,7 @@ case class RegionTaskExec(child: SparkPlan,
 
       val schemaInferer: SchemaInfer = if (satisfyDowngradeThreshold) {
         // Should downgrade to full table scan for one region
-        logger.info(
+        logger.debug(
```
ilovesoup: Such short info is ok and it's useful.

zhexuany: how about setting its log level to warn?

zhexuany:
No need test since its just log level change.

PR #679, opened 2019-04-25
Fix supported spark version not updated

birdstorm:
read from config property file rather than maven system property


birdstorm:
No tests involved. I will merge it without integration tests.

PR #680, opened 2019-04-25
Optimize version control by minimizing duplicate code

### LINKED ISSUE #678: Optimize Version Dependency in TiStrategy
Seems that we can extract TiAggregation from TiStrategy and thus optimize multi-version code base.

birdstorm:
Close #678 
delete some useless code


zhexuany:
> delete some useless code

LGTM

PR #687, opened 2019-04-28
Merge master to batch-write

zhexuany:


birdstorm:
how about #676 and #686

zhexuany:
@birdstorm PTAL

PR #690, opened 2019-04-29
Fix tempView may be unresolved when applying timestamp to plan

birdstorm:


core/src/main/scala/org/apache/spark/sql/extensions/rules.scala
```diff
@@ -55,31 +53,28 @@ case class TiResolutionRule(getOrCreateTiContext: SparkSession => TiContext)(
...
+        meta
       )(sqlContext)
       // Use SubqueryAlias so that projects and joins can correctly resolve
       // UnresolvedAttributes in JoinConditions, Projects, Filters, etc.
       SubqueryAlias(tableName, LogicalRelation(tiDBRelation))
     }
 
-  protected def resolveRelations(ts: TiTimestamp): PartialFunction[LogicalPlan, LogicalPlan] = {
+  protected def resolveTiDBRelations: PartialFunction[LogicalPlan, LogicalPlan] = {
+    case i @ InsertIntoTable(UnresolvedRelation(tableIdentifier), _, _, _, _)
```
birdstorm: move InsertIntoTable to keep same style as ResolveRelations.apply

core/src/test/scala/org/apache/spark/sql/catalyst/plans/logical/LogicalPlanTestSuite.scala
```diff
@@ -62,31 +60,31 @@ class LogicalPlanTestSuite extends BaseTiSparkSuite {
...
+      case plan: HandleRDDExec =>
+        check(plan.tiHandleRDD.dagRequest.getStartTs)
+      case plan: RegionTaskExec =>
+        check(plan.dagRequest.getStartTs)
+      case _ =>
     }
-    println(df.queryExecution.analyzed)
-    df.queryExecution.analyzed.foreach { checkTimestamp }
+    df.explain
+    println(df.queryExecution.executedPlan)
```
birdstorm: StartTs will be added after Strategies are applied.

tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
```diff
@@ -823,6 +823,7 @@ public String toString() {
       sb.append(", Limit: ");
       sb.append("[").append(limit).append("]");
     }
+    sb.append(", startTs: ").append(startTs.getVersion());
```
birdstorm: I added a startTs information in `TiDagRequest` output. Not really sure if it is useful? @zhexuany

zhexuany: It is better to print it out.

PR #695, opened 2019-05-05
Save Mode

marsishandsome:
TiSpark only support `Append` SaveMode.

`Append` in TiSpark means `upsert`,
1. if primary key exists in db, data will be updated
2. if no same primary key exists, data will be inserted.

| SaveMode | Support | Semantics |
| -------- | ------- | --------- |
| Append | true | TiSpark's `Append` means upsert. If primary key is same, data will be updated; if no same primary key exists, data will be inserted.  |
| Overwrite | false |  - |
| ErrorIfExists | false | - |
| Ignore | false | - |

Currently TiSpark only support writing data to such tables:
1. the table does not contain a primary key
2. the table's primary key is `TINYINT`、`SMALLINT`、`MEDIUMINT` or `INTEGER`
3. the table's primary key is `auto increment`

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/codec/TableCodec.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/codec/TableCodec.java
@@ -12,6 +12,15 @@ import java.util.ArrayList;
 import java.util.List;
 
 public class TableCodec {
+
+  /**
+   * Row layout: colID1, value1, colID2, value2, .....
+   * @param colTypes
+   * @param colIDs
+   * @param values
+   * @return
+   * @throws IllegalAccessException
+   */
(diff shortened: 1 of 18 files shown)


marsishandsome:
@birdstorm @zhexuany PTAL

tikv-client/src/main/java/com/pingcap/tikv/codec/TableCodec.java
```diff
@@ -24,6 +21,9 @@
               colTypes.length, colIDs.size()));
     }
 
+    // Row layout: colID1, value1, colID2, value2, .....
```
zhexuany: this comment belongs to `encodeRow`

marsishandsome: fix

PR #705, opened 2019-05-09
Fix copyFromUtf8 does not decode UTF8 String correctly

birdstorm:


birdstorm:
Close due to not important.

PR #708, opened 2019-05-10
add enable config for batch write

marsishandsome:
default disable batch write


marsishandsome:
@zhexuany @birdstorm PTAL

core/src/main/scala/com/pingcap/tispark/TiConfigConst.scala
```diff
@@ -39,6 +39,8 @@ object TiConfigConst {
   val CACHE_EXPIRE_AFTER_ACCESS: String = "spark.tispark.statistics.expire_after_access"
   val SHOW_ROWID: String = "spark.tispark.show_rowid"
   val DB_PREFIX: String = "spark.tispark.db_prefix"
+  val WRITE_ALLOW_SPARK_SQL: String = "spark.tispark.write.allow_spark_sql"
```
zhexuany: add a todo? We need to remove this flag once batch write is stable.

Issue #709, opened 2019-05-10
FilterPushdownSuite is not stable. 


zhexuany:
FilterPushdownSuite works as expected if i run it in IDE, but fails when I just run `mvn test`
```
FilterPushdownSuite:
- Test Simple Comparisons *** FAILED ***
  com.pingcap.tikv.exception.TiClientInternalException: Table not exist
  at com.pingcap.tispark.TiDBRelation$$anonfun$1.apply(TiDBRelation.scala:41)
  at com.pingcap.tispark.TiDBRelation$$anonfun$1.apply(TiDBRelation.scala:41)
  at scala.Option.getOrElse(Option.scala:121)
  at com.pingcap.tispark.TiDBRelation.<init>(TiDBRelation.scala:41)
  at com.pingcap.tispark.TiDBDataSource.createRelation(TiDBDataSource.scala:47)
  at org.apache.spark.sql.execution.datasources.DataSource.resolveRelation(DataSource.scala:341)
  at org.apache.spark.sql.DataFrameReader.loadV1Source(DataFrameReader.scala:239)
  at org.apache.spark.sql.DataFrameReader.load(DataFrameReader.scala:227)
  at org.apache.spark.sql.DataFrameReader.load(DataFrameReader.scala:164)
  at com.pingcap.tispark.datasource.FilterPushdownSuite.com$pingcap$tispark$datasource$FilterPushdownSuite$$testFilter(FilterPushdownSuite.scala:52)
  ...
- Test >= and <= *** FAILED ***
  com.pingcap.tikv.exception.TiClientInternalException: Table not exist
  at com.pingcap.tispark.TiDBRelation$$anonfun$1.apply(TiDBRelation.scala:41)
  at com.pingcap.tispark.TiDBRelation$$anonfun$1.apply(TiDBRelation.scala:41)
  at scala.Option.getOrElse(Option.scala:121)
  at com.pingcap.tispark.TiDBRelation.<init>(TiDBRelation.scala:41)
  at com.pingcap.tispark.TiDBDataSource.createRelation(TiDBDataSource.scala:47)
  at org.apache.spark.sql.execution.datasources.DataSource.resolveRelation(DataSource.scala:341)
  at org.apache.spark.sql.DataFrameReader.loadV1Source(DataFrameReader.scala:239)
  at org.apache.spark.sql.DataFrameReader.load(DataFrameReader.scala:227)
  at org.apache.spark.sql.DataFrameReader.load(DataFrameReader.scala:164)
[...]

PR #710, opened 2019-05-10
fix Table not exist Exception in Test

marsishandsome:
fix issue https://github.com/pingcap/tispark/issues/709


marsishandsome:
@zhexuany @birdstorm PTAL

PR #711, opened 2019-05-10
set ut parameter -DskipAfterFailureCount=1

marsishandsome:
skip remaining tests after first failure or error has happened


marsishandsome:
@zhexuany @birdstorm PTAL

PR #715, opened 2019-05-13
batch write test: add unsupport check test

marsishandsome:


marsishandsome:
@zhexuany PTAL

tikv-client/src/main/java/com/pingcap/tikv/meta/TiTableInfo.java
```diff
@@ -155,6 +155,15 @@ TableInfo toProto() {
         .build();
   }
 
+  public boolean hasPrimaryKey() {
+    for (TiColumnInfo col : getColumns()) {
+      if (col.isPrimaryKey()) {
+        return true;
```
zhexuany: maybe we should add a field value?  Once we find a primary key column, we assign the value to the field value.

marsishandsome: done

marsishandsome:
@zhexuany PTAL

PR #716, opened 2019-05-14
add edge condition unit test for batch write

marsishandsome:


marsishandsome:
@zhexuany PTAL

marsishandsome:
@zhexuany PTAL

tikv-client/src/main/java/com/pingcap/tikv/TwoPhaseCommitter.java
```diff
@@ -288,6 +288,10 @@ private void doPrewriteSecondaryKeysInBatches(
         Kvrpcpb.Mutation mutation =
             Kvrpcpb.Mutation.newBuilder().setKey(key).setValue(value).setOp(Kvrpcpb.Op.Put).build();
         mutations.put(key, mutation);
+      } else {
```
zhexuany: Should we explain when value can be nil?

marsishandsome: done

PR #722, opened 2019-05-15
add datasource api ut for batch write

marsishandsome:
1. add datasource api ut
2. visit 2 different tidb (fix TiSession cache issue)
3. clean up code
4. make most of the datasource ut for `with TiExtension`


marsishandsome:
@zhexuany PTAL

core/src/main/scala/com/pingcap/tispark/TiSessionCache.scala
```diff
@@ -17,17 +17,20 @@ package com.pingcap.tispark
...
+import scala.collection.mutable
+
 object TiSessionCache {
-  private var sessionCached: TiSession = null
+  private val sessionCachedMap = mutable.Map[String, TiSession]()
 
   // Since we create session as singleton now, configuration change will not
   // reflect change
   def getSession(conf: TiConfiguration): TiSession = this.synchronized {
+    val pdfAddrs = conf.getPdAddrsString
```
zhexuany: @birdstorm PTAL

core/src/main/scala/com/pingcap/tispark/TiSessionCache.scala
```diff
@@ -17,17 +17,20 @@ package com.pingcap.tispark
...
+    var sessionCached = sessionCachedMap.getOrElse(pdfAddrs, null)
```
birdstorm: how about 
```suggestion
    sessionCachedMap.getOrElseUpdate(pdfAddrs, TiSession.create(conf))
```

birdstorm: we do not need sessionCached anymore.

core/src/main/scala/com/pingcap/tispark/examples/TiDataSourceExample.scala
```diff
@@ -108,7 +110,7 @@ object TiDataSourceExample {
     val df = readUsingDataSourceAPI(sqlContext)
 
     // Append
-    // if target_table_append does not exist, it will be created automatically
+    // target_table_append should exists in tidb
```
birdstorm: exist

PR #724, opened 2019-05-15
add error test for batch write

marsishandsome:


core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -127,6 +127,15 @@ object TiBatchWrite {
     val kvClient = tiSession.createTxnClient()
     val tableRef = TiTableReference(options.database, options.table)
     val (tiDBInfo, tiTableInfo) = getDBAndTableInfo(tableRef, tiContext)
+
+    // check check check
```
birdstorm: LOL

core/src/test/scala/com/pingcap/tispark/datasource/ErrorSuite.scala
```diff
@@ -0,0 +1,126 @@
+package com.pingcap.tispark.datasource
+
+import com.pingcap.tikv.exception.TiBatchWriteException
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.types.{IntegerType, LongType, StringType, StructField, StructType}
+
+// without TiExtensions
+// will not load tidb_config.properties to SparkConf
+class ErrorSuite extends BaseDataSourceSuite("test_datasource_error") {
```
birdstorm: should we rename it as `ExceptionTestSuite`?

marsishandsome: nice, will rename in next patch

PR #730, opened 2019-05-16
Disable downgrade scan temporarily

birdstorm:


core/src/main/scala/org/apache/spark/sql/execution/CoprocessorRDD.scala
```diff
@@ -176,9 +176,8 @@ case class RegionTaskExec(child: SparkPlan,
       .createMetric(sparkContext, "number of downgrade ranges scanned")
   )
 
-  private val sqlConf = sqlContext.conf
-  private val downgradeThreshold =
-    sqlConf.getConfString(TiConfigConst.REGION_INDEX_SCAN_DOWNGRADE_THRESHOLD, "10000").toInt
+  private val downgradeThreshold = 1000000000
```
marsishandsome: pls add a TODO and an issue ID

birdstorm: sure

PR #735, opened 2019-05-16
fix partition table index scan

zhexuany:
This is only temp fix, I am investigating is there any better fix. 

In addition to the real fix, I also improve the explain info and refactor the way to construct ScanPlan.

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/region/TiRegion.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/region/TiRegion.java
@@ -195,12 +195,7 @@ public class TiRegion implements Serializable {
 
   @Override
   public int hashCode() {
-    int hash = 7;
-    hash = 31 * hash * meta.hashCode();
-    hash = 31 * hash * peer.hashCode();
-    hash = 31 * hash * commandPri.hashCode();
-    hash = 31 * hash * isolationLevel.hashCode();
-    return hash;
+    return Objects.hash(meta, peer, isolationLevel, commandPri);
   }
 
   @Override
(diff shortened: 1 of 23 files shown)


tikv-client/src/main/java/com/pingcap/tikv/util/RangeSplitter.java
```diff
@@ -79,6 +83,25 @@ public String getHost() {
...
+      return regionTask.host.equals(this.host)
+          && regionTask.ranges.equals(this.ranges)
+          && regionTask.region.equals(this.region)
+          && regionTask.store.equals(this.store);
+    }
+
+    @Override
+    public int hashCode() {
+      int hash = 7;
+      hash = 31 * hash * this.region.hashCode();
```
birdstorm: I think you should use the following instead.
```suggestion
      hash = 31 * hash + this.region.hashCode();
```

zhexuany: I already remove this. It seems we don't need this.

tikv-client/src/main/java/com/pingcap/tikv/region/TiRegion.java
```diff
@@ -183,6 +183,26 @@ public Region getMeta() {
...
+  }
+
+  @Override
+  public int hashCode() {
+    int hash = 7;
+    hash = 31 * hash * meta.hashCode();
+    hash = 31 * hash * peer.hashCode();
+    hash = 31 * hash * commandPri.hashCode();
+    hash = 31 * hash * isolationLevel.hashCode();
+    return hash;
```
marsishandsome: `return Objects.hash(meta, peer, isolationLevel, commandPri);`

zhexuany:
Maybe file this "suggest to add test: read from partition table with 1000 partitions" in another PR.

PR #742, opened 2019-05-21
remove index read conf

zhexuany:
We will enable index scan by default.


zhexuany:
We need to enable it by default. We need to make the index scan stable. The only use of this parameter is to let the user use table scan in case there is some bug in the index scan.

PR #749, opened 2019-05-21
Fix lazy val not working due to incorrect implementation

birdstorm:
Some lazy val were unable to work due to incorrect order of initialization.
It is also related to extensions settings in test config.

### DIFF
--- a/core/src/main/spark-2.4/org/apache/spark/sql/catalyst/catalog/TiCompositeSessionCatalog.scala
+++ b/core/src/main/spark-2.4/org/apache/spark/sql/catalyst/catalog/TiCompositeSessionCatalog.scala
@@ -60,15 +60,29 @@ class TiCompositeSessionCatalog(val tiContext: TiContext)
     override val sessionCatalog: SessionCatalog = tiContext.sessionCatalog
   }
 
-  val policy: CompositeCatalogPolicy = TiLegacyPolicy(tiContext)
+  lazy val policy: CompositeCatalogPolicy = TiLegacyPolicy(tiContext)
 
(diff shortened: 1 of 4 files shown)


core/src/test/scala/org/apache/spark/sql/test/SharedSQLContext.scala
```diff
@@ -193,18 +210,6 @@ object SharedSQLContext extends Logging {
       _spark = _sparkSession
     }
 
-  private def initializeJDBC(): Unit =
-    if (_sparkJDBC == null) {
-      _sparkJDBC = _sparkSession
-    }
-
-  protected def initializeTiContext(): Unit =
```
birdstorm: We do not need to initialize TiContext for tests anymore because it is already present in SparkExtensions

core/src/main/spark-2.4/org/apache/spark/sql/catalyst/catalog/TiCompositeSessionCatalog.scala
```diff
@@ -60,15 +60,29 @@ class TiCompositeSessionCatalog(val tiContext: TiContext)
     override val sessionCatalog: SessionCatalog = tiContext.sessionCatalog
   }
 
-  val policy: CompositeCatalogPolicy = TiLegacyPolicy(tiContext)
+  lazy val policy: CompositeCatalogPolicy = TiLegacyPolicy(tiContext)
```
birdstorm: Previous implementation causes tiContext to be initialized during construction despite it is marked as `lazy`

PR #752, opened 2019-05-22
fix from spark type

marsishandsome:


marsishandsome:
@birdstorm @zhexuany PTAL

PR #753, opened 2019-05-22
add column mapping for batch write

marsishandsome:


marsishandsome:
@zhexuany PTAL

marsishandsome:
@zhexuany PTAL

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -98,10 +102,30 @@ class TiBatchWrite(@transient val df: DataFrame,
...
     }
+    dfColSize = df.columns.length
+    tableColSize = tiTableInfo.getColumns.size()
 
     // check check check
     checkUnsupported()
     checkColumnNumbers()
+
+    // initialize dfTiColumnInfo
+    dfTiColumnInfo = new Array[TiColumnInfo](dfColSize)
```
zhexuany: this should move to sparkRow2TiRow. Otherwise, we do not have a way to allocate id.

PR #757, opened 2019-05-23
Fix type long bug

marsishandsome:


marsishandsome:
@zhexuany PTAL

core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
```diff
@@ -114,28 +114,50 @@ object TiUtil {
...
       case _: IntegerType =>
         if (tp.asInstanceOf[IntegerType].isUnsignedLong) {
           DataTypes.createDecimalType(20, 0)
         } else {
-          sql.types.LongType
+          if (version == 0) {
+            sql.types.LongType
+          } else {
+            tp.getType match {
+              case MySQLType.TypeLong     => sql.types.IntegerType
```
birdstorm: we might need to consider Unsigned Integer here.

ilovesoup:
Please provide corresponding issue / Jira number.

ilovesoup:
And please don't put user name on any description.

PR #760, opened 2019-05-24
add type converter for batch write

marsishandsome:
currently only support convert to:
1. TINYINT 
2. SMALLINT 
3. MEDIUMINT
4.  INT 
5. BIGINT
6. Double 
7. Float

and only support convert from:
1. BooleanType
2. ByteType
3. ShortType
4. IntegerType
5. LongType
6. FloatType
7. DoubleType
8. StringType

other convert type will provide in following PRs

### DIFF
--- a/core/src/main/scala/com/pingcap/tispark/utils/TiConverter.scala
+++ b/core/src/main/scala/com/pingcap/tispark/utils/TiConverter.scala
@@ -133,12 +133,14 @@ object TiConverter {
     }
 
     targetColumnInfo.getType match {
-      //case BitType       =>
-      //case BytesType     =>
-      //case DateTimeType  =>
-      //case DateType      =>
-      //case DecimalType   =>
-      //case EnumType      =>
+      //case _: BitType       =>
+      //case _: BytesType     =>
+      //case _: DateTimeType  =>
+      // TODO: add test for DateType
+      case _: DateType => value
+      // TODO: add test for DecimalType
+      case _: DecimalType => value
+      //case _: EnumType      =>
       case _: IntegerType =>
         if (targetColumnInfo.getType.isUnsigned) {
           convertToUnsigned(targetColumnInfo, value)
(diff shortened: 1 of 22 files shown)


marsishandsome:
@birdstorm @zhexuany PTAL

marsishandsome:
@birdstorm @zhexuany PTAL

core/src/main/scala/com/pingcap/tispark/utils/TiConverter.scala
```diff
@@ -0,0 +1,347 @@
...
+          s"do not support converting from ${value.getClass} to column type: ${targetColumnInfo.getType}"
+        )
+    }
+
+    targetColumnInfo.getType.getType match {
+      case MySQLType.TypeFloat =>
+        val lowerBound = -java.lang.Float.MAX_VALUE
+        val upperBound = java.lang.Float.MAX_VALUE
+
+        if (java.lang.Double.compare(result, upperBound) > 0) {
```
birdstorm: I doubt whether we should use Double.compare to compare double values. It should be better to use EPSILON in order to reduce the impact of precision loss.

marsishandsome: fix

core/src/main/scala/com/pingcap/tispark/utils/TiConverter.scala
```diff
@@ -0,0 +1,333 @@
...
+    Math.round(v).longValue()
+
+  private def doubleToLong(v: java.lang.Double): java.lang.Long =
+    Math.round(v)
+
+  private def stringToDouble(v: String): java.lang.Double =
+    java.lang.Double.parseDouble(v)
+
+  private def stringToLong(v: String): java.lang.Long =
+    doubleToLong(stringToDouble(v))
```
zhexuany: why not Longs.tryParse?

zhexuany:
except for one comment, other LGTM

core/src/main/scala/com/pingcap/tispark/utils/TiConverter.scala
marsishandsome: fix

PR #762, opened 2019-05-24
add parallel number for ci test

marsishandsome:


marsishandsome:
@birdstorm @zhexuany PTAL

PR #764, opened 2019-05-24
random shuffle ci tests

marsishandsome:


marsishandsome:
@birdstorm @zhexuany PTAL

PR #766, opened 2019-05-24
speed ci tests

marsishandsome:
1. random shuffle ci tests (test load balance)
2. clone proto file at prepare stage and stage it
3. set MAVEN_OPTS="-Xmx2G -XX:MaxPermSize=512M -XX:ReservedCodeCacheSize=512M"
4. change Suite to Test if no tests exists (test load balance)


marsishandsome:
@birdstorm @zhexuany PTAL

PR #777, opened 2019-05-30
mock data

marsishandsome:


core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -162,28 +188,46 @@ class TiBatchWrite(@transient val df: DataFrame,
     val encodedTiRowRDD = deduplicateRDD.map {
       case (key, tiRow) => (key, tiRow, encodeTiRow(tiRow))
     }
+    encodedTiRowRDD.persist(storageLevel)
 
     // region pre-split
     if (enableRegionPreSplit) {
       logger.info("region pre split is enabled.")
+      val preSplitStartTime = System.currentTimeMillis()
```
zhexuany: previous region pre-splits' assumption is no longer valid if we support index insertion. We can temporarily remove region pre-split logic from the codebase and add it back once we have a better implementation.

PR #790, opened 2019-06-04
add overflow check for batch write

marsishandsome:


marsishandsome:
@birdstorm @zhexuany PTAL

tikv-client/src/test/java/com/pingcap/tikv/codec/TableCodecTest.java
```diff
@@ -24,8 +24,8 @@ private static TiTableInfo createTable() {
         // TODO: enable when support Timestamp
         // .addColumn("c3", DateTimeType.DATETIME)
         // .addColumn("c4", TimestampType.TIMESTAMP)
-        .addColumn("c5", StringType.VARCHAR)
-        .addColumn("c6", StringType.VARCHAR)
+        .addColumn("c5", StringType.VARCHAR255)
```
zhexuany: We may change the signature of `addColumn` as the following, 

 .addColumn("c5", StringType.VARCHAR, 235)

where 235 specifying the length.

marsishandsome: agree

Issue #778, opened 2019-05-30
KeyNotInRegion error cause TiKV panic


marsishandsome:
TikV panic when using Batch Write

code: https://github.com/pingcap/tispark/pull/777

tikv log
```
2019/05/30 21:32:50.238 ERRO mod.rs:497: thread 'sched-worker-pool' panicked 'key exceed bound: KeyNotInRegion([116, 128, 0, 0, 0, 0, 0, 0, 255, 35, 95, 114, 128, 0, 0, 0, 0, 255, 43, 242, 88, 0, 0, 0, 0, 0, 250, 0, 0, 0, 0, 0, 0, 0, 0], id: 3592 start_key: "t\200\000\000\000\000\000\000\377#_r\200\000\000\000\000\3772`\347\000\000\000\000\000\372" end_key: "t\200\000\000\000\000\000\000\377#_r\200\000\000\000\000\377<\373|\000\000\000\000\000\372" region_epoch {conf_ver: 5 version: 19} peers {id: 3593 store_id: 1} peers {id: 3594 store_id: 5} peers {id: 3595 store_id: 4})' at "src/raftstore/store/region_snapshot.rs:321"
```

spark log
```
com.pingcap.tikv.exception.TiBatchWriteException: prewrite secondary key error
 at com.pingcap.tikv.TwoPhaseCommitter.doPrewriteSecondaryKeySingleBatch(TwoPhaseCommitter.java:347)
 at com.pingcap.tikv.TwoPhaseCommitter.doPrewriteSecondaryKeysInBatches(TwoPhaseCommitter.java:315)
 at com.pingcap.tikv.TwoPhaseCommitter.doPrewriteSecondaryKeys(TwoPhaseCommitter.java:261)
 at com.pingcap.tikv.TwoPhaseCommitter.prewriteSecondaryKeys(TwoPhaseCommitter.java:242)
 at com.pingcap.tispark.TiBatchWrite$$anonfun$com$pingcap$tispark$TiBatchWrite$$write$3.apply(TiBatchWrite.scala:281)
 at com.pingcap.tispark.TiBatchWrite$$anonfun$com$pingcap$tispark$TiBatchWrite$$write$3.apply(TiBatchWrite.scala:268)
 at org.apache.spark.rdd.RDD$$anonfun$foreachPartition$1$$anonfun$apply$29.apply(RDD.scala:935)
[...]

PR #795, opened 2019-06-05
fix key not in range bug

marsishandsome:
https://github.com/pingcap/tispark/issues/778

### DIFF
--- a/core/src/main/scala/com/pingcap/tispark/examples/TiBatchWriteRandomDataTest.scala
+++ b/core/src/main/scala/com/pingcap/tispark/examples/TiBatchWriteRandomDataTest.scala
@@ -1,0 +1,283 @@
+/*
+ * Copyright 2019 PingCAP, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.pingcap.tispark.examples
+
+import com.pingcap.tispark.{TiBatchWrite, TiDBOptions}
+import org.apache.spark.SparkConf
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.types.{IntegerType, StructField, StructType}
+import org.apache.spark.sql.{Row, SparkSession, TiContext}
+
+object TiBatchWriteRandomDataTest {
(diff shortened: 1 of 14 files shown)


tikv-client/src/main/java/com/pingcap/tikv/TwoPhaseCommitter.java
```diff
@@ -298,23 +308,25 @@ private void doPrewriteSecondaryKeysInBatches(
...
     }
 
     // For prewrite, stop sending other requests after receiving first error.
     for (BatchKeys batchKeys : batchKeyList) {
-      doPrewriteSecondaryKeySingleBatch(backOffer, primaryKey, batchKeys, mutations);
+      doPrewriteSecondaryKeySingleBatchWithRetry(backOffer, primaryKey, batchKeys, mutations);
     }
 
-    LOG.debug(
+    LOG.info(
```
zhexuany: debug level seems ok to me.

marsishandsome: reset

core/src/main/scala/com/pingcap/tispark/examples/TiBatchWriteRandomDataTest.scala
```diff
@@ -0,0 +1,283 @@
...
+
+    // time
+    val end = System.currentTimeMillis()
+    val seconds = (end - start) / 1000
+    println(s"total time: $seconds seconds")
+  }
+}
+
+/*
+CREATE TABLE RANDOM_DATA (
```
zhexuany: what is this used for?

tikv-client/src/main/java/com/pingcap/tikv/policy/RetryPolicy.java
```diff
@@ -62,11 +64,9 @@ public RespT callWithRetry(Callable<RespT> proc, String methodName) throws GrpcE
       }
 
       // Handle response error
-      if (handler != null) {
-        boolean retry = handler.handleResponseError(backOffer, result);
-        if (retry) {
-          continue;
-        }
+      boolean retry = handler.handleResponseError(backOffer, result);
```
zhexuany:
```
if(handler.handleResponseError(backOffer, result)) {
    continue; 
}
```  
?

core/src/main/scala/com/pingcap/tispark/examples/TiBatchWriteRandomDataTest.scala
marsishandsome: use this script to create the table before run `TiBatchWriteRandomDataTest`

tikv-client/src/main/java/com/pingcap/tikv/policy/RetryPolicy.java
marsishandsome: code is more understandable with `boolean retry`

Issue #799, opened 2019-06-06
improve prewrite process if region split


marsishandsome:
if region splits during prewrite, pending batches should refetch region info

PR #800, opened 2019-06-06
improve prewrite process if region split

marsishandsome:
close https://github.com/pingcap/tispark/issues/799


tikv-client/src/main/java/com/pingcap/tikv/TwoPhaseCommitter.java
```diff
@@ -106,6 +106,8 @@ public BytePairWrapper(byte[] key, byte[] value) {
   /** unit is second */
   private static final long DEFAULT_BATCH_WRITE_LOCK_TTL = 3000;
 
+  private static final long MAX_RETRY_LEVEL = 3;
```
zhexuany: MAX_RETRY_TIMES makes much more sense.

tikv-client/src/main/java/com/pingcap/tikv/TwoPhaseCommitter.java
```diff
@@ -310,13 +313,45 @@ private void doPrewriteSecondaryKeysInBatchesWithRetry(
...
+              String.format(
+                  "> max retry number %s, oldRegion=%s, currentRegion=%s",
+                  MAX_RETRY_LEVEL, oldRegion, currentRegion));
+        }
+        LOG.debug(
+            String.format(
+                "oldRegion=%s != currentRegion=%s, will refetch region info and retry",
+                oldRegion, currentRegion));
+        retryPrewriteBatch(
+            backOffer, primaryKey, batchKeys, mutations, level == -1 ? 1 : level + 1);
```
zhexuany: why do we need to check -1 here?  What if we pass 0 rather than -1 into this function?

birdstorm: The logic was correct, but we may prettify the code into:
```suggestion
            backOffer, primaryKey, batchKeys, mutations, level + 1);
```
and call `retryPrewriteBatch(.., 0)`

zhexuany: Yes.

tikv-client/src/main/java/com/pingcap/tikv/TwoPhaseCommitter.java
marsishandsome: agree

tikv-client/src/main/java/com/pingcap/tikv/TwoPhaseCommitter.java
marsishandsome: fixed

marsishandsome:
@zhexuany PTAL

PR #805, opened 2019-06-09
Move set out snapshot

### LINKED ISSUE #806: In `Snapshot`, we have a set method which is not correct from the semantic of `Snapshot`
We need to move this method in another place.

zhexuany:
In this PR, we did several things. 
1. change how to create a TxnKvClient 
2. move set out from `Snapshot` 

This PR closes #806


zhexuany:
This PR closes #806

tikv-client/src/main/java/com/pingcap/tikv/TiSessionCache.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class TiSessionCache {
+  private static Map<String, TiSession> sessionCachedMap = new ConcurrentHashMap<>();
+
+  // Since we create session as singleton now, configuration change will not
+  // reflect change
+  public static TiSession getSession(TiConfiguration conf) {
```
birdstorm: do we need to synchronize?

zhexuany: Yes. Just added and change `ConcurentHashMap` to `HashMap`

PR #819, opened 2019-06-13
Support index upsert

zhexuany:
In this PR, we support pk and index upsert.


tikv-client/src/main/java/com/pingcap/tikv/TwoPhaseCommitter.java
```diff
@@ -296,7 +296,7 @@ private void doPrewriteSecondaryKeysInBatchesWithRetry(
             Kvrpcpb.Mutation.newBuilder().setKey(key).setValue(value).setOp(Kvrpcpb.Op.Put).build();
       } else {
         // value can be null (table with one primary key integer column, data is encoded in key)
-        mutation = Kvrpcpb.Mutation.newBuilder().setKey(key).setOp(Kvrpcpb.Op.Put).build();
+        mutation = Kvrpcpb.Mutation.newBuilder().setKey(key).setOp(Kvrpcpb.Op.Del).build();
```
marsishandsome: should line 152 also change?

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -760,16 +899,27 @@ class ToBeCheckedRow(val row: TiRow,
...
     extends Serializable {
 
-  //  val idxId: Long
-  // if handle is conflict or any key in indices key conflict with each other
-  // then two row equals
+  def isDuplicated: Boolean = {
+    handleKey.oldHandle != -1 || indexKeys.exists(key => key.oldHandle != -1)
+  }
+
+  // When the following case happens, two ToBeCheckedRows are considered as same:
```
marsishandsome: `all of the following case happens` or `one of the following case happens`?

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -760,16 +899,27 @@ class ToBeCheckedRow(val row: TiRow,
...
+  // 1. handle key is same if not null
+  // 2. at least one of index keys is same and index keys is not empty
+  // 3. handle key is null and index keys are empty, but row is same.
   override def equals(that: Any): Boolean =
```
marsishandsome: consider the following case:

```
ToBeCheckRow1:
  handleKey = 1
  indexKeys = [11, 111]

ToBecheckRow2:
  handleKey = 1
  indexKeys = [22, 222]
```

`ToBeCheckRow1 == ToBecheckRow2`
but `ToBeCheckRow1.hashcode != ToBecheckRow2.hashcode`

marsishandsome:
the following test fails

```
class TestSuite extends BaseDataSourceTest("test_datasource_basic2") {
  private val row3 = Row(3)
  private val row4 = Row(4)

  private val schema = StructType(
    List(
      StructField("i", IntegerType)
    )
  )

    super.beforeAll()

    dropTable()
    jdbcUpdate(s"create table $dbtable(i int primary key)")
  }

  test("Test Write Append") {
    val data: RDD[Row] = sc.makeRDD(List(row3, row4))
    val df = sqlContext.createDataFrame(data, schema)

    df.write
      .format("tidb")
      .options(tidbOptions)
      .option("database", database)
      .option("table", table)
      .mode("append")
      .save()

    testTiDBSelect(Seq(row3, row4))
  }

  override def afterAll(): Unit =
    try {
      dropTable()
    } finally {
      super.afterAll()
    }
}
```

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
zhexuany: one of.

tikv-client/src/main/java/com/pingcap/tikv/TwoPhaseCommitter.java
zhexuany: Yes.

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
zhexuany: Yes. More refine is on the way

zhexuany:
@marsishandsome The test case is due to a bug in encodeRow. 
Before 
```
  // We could not set nil value into kv.
    if (values.length == 0) {
      return new byte[] {Codec.NULL_FLAG};
    }
```
After 

```
  // We could not set nil value into kv.
    if (cdo.toBytes().length == 0) {
      return new byte[] {Codec.NULL_FLAG};
    }
```

core/src/test/scala/com/pingcap/tispark/datasource/AddingIndexUpsertSuite.scala
```diff
@@ -0,0 +1,100 @@
...
+
+    val options = Some(Map("upsert" -> "true", "deduplicate" -> "true"))
+    tidbWrite(
+      List(row3, row3, conflcitWithOneIndex, conflcitWithTwoIndices),
+      schema,
+      options
+    )
+    testTiDBSelect(Seq(row2, row3, conflcitWithOneIndex, conflcitWithTwoIndices), "c1")
+  }
+}
```
marsishandsome: better to test `deduplicate = false`

core/src/test/scala/com/pingcap/tispark/datasource/OnlyOnePkSuite.scala
```diff
@@ -0,0 +1,45 @@
+package com.pingcap.tispark.datasource
+
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.types.{IntegerType, StructField, StructType}
+
+class OnlyOnePkSuite extends BaseDataSourceTest("test_datasource_basic2") {
```
marsishandsome: `test_datasource_basic2` may be use another name

core/src/test/scala/com/pingcap/tispark/datasource/OnlyOnePkSuite.scala
```diff
@@ -0,0 +1,45 @@
...
+      .option("table", table)
+      .mode("append")
+      .save()
+
+    testTiDBSelect(Seq(row3, row4))
+  }
+
+  override def afterAll(): Unit =
+    try {
+//      dropTable()
```
marsishandsome: delete `//`

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -755,31 +891,37 @@ class KeyWithDupInfo(val key: SerializableKey, val oldRow: TiRow, var oldHandle:
...
-        val mayConflict = row.indexKeys.toSet
-        this.indexKeys.exists(mayConflict)
+        // 1. handle key is same
+        var handleConflict = false
+        if (handleKey != null && row.handleKey != null) {
+          handleConflict = handleKey.equals(row.handleKey)
+        }
+        // 2. index keys are empty, just return false
+        if (indexKeys.isEmpty && row.indexKeys.isEmpty) {
+          return handleConflict
```
marsishandsome: comments and code is different.
comments: just return false
code: return handleConflict

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -755,31 +891,37 @@ class KeyWithDupInfo(val key: SerializableKey, val oldRow: TiRow, var oldHandle:
...
+        }
+        // 3. check is there any index key same
+        this.indexKeys.exists(row.indexKeys.toSet) || handleConflict
```
marsishandsome: maybe change to `handleConflict || this.indexKeys.exists(row.indexKeys.toSet)` for better performance

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -403,33 +414,31 @@ class TiBatchWrite(@transient val df: DataFrame,
...
-              Some(iterable.size)
-            } else {
-              None
-            }
-        }
-
-        if (!duplicateCountRDD.isEmpty()) {
-          throw new TiBatchWriteException("data conflicts! set the parameter deduplicate.")
+  private def deduplicateIfNecessary(rdd: RDD[ToBeCheckedRow]) = {
+    val shuffledRDD = rdd.map(row => (row, 0))
```
marsishandsome: maybe `shuffledRDD` -> `mappedRDD`

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -403,33 +414,31 @@ class TiBatchWrite(@transient val df: DataFrame,
...
+    val rddGroupByKey = shuffledRDD.groupByKey
+    val duplicateCountRDD = rddGroupByKey.flatMap {
```
marsishandsome: move line 420-427 after line 435

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -266,15 +267,25 @@ class TiBatchWrite(@transient val df: DataFrame,
...
+
+    val deduplicatedTiRowRdd = deduplicateIfNecessary(toBeCheckedRdd)
+    val rddToBeInserted = if (options.upsert) {
+      // step 1: remove duplicate item from data to be inserted. The duplicate basically means
+      // two row items violate primary key or unique index constraint.
+      deduplicatedTiRowRdd
+    } else {
+      checkConflictWithInsertKeys(toBeCheckedRdd)
+      deduplicatedTiRowRdd
+    }
```
marsishandsome: better change to 
```
if (!options.upsert) {
  checkConflictWithInsertKeys(toBeCheckedRdd)
}
val rddToBeInserted = deduplicatedTiRowRdd
```

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -755,31 +891,37 @@ class KeyWithDupInfo(val key: SerializableKey, val oldRow: TiRow, var oldHandle:
...
     }
 
+  // there is no way to implement hashCode correctly without breaking the contract of hashCode.
   override def hashCode(): Int = {
-    var hash = 7
-    indexKeys.foreach { kv =>
-      hash += util.Arrays.hashCode(kv.key.bytes)
-    }
-    hash
+    7
```
marsishandsome: @birdstorm @windtalker PTAL
if hashcode always return 7, spark will use `ToBeCheckedRow` to shuffle data, which means
the result rdd will has only one partition. if data is very large, the computing time will be very long.

birdstorm: I hate hash code 😭

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -481,35 +546,41 @@ class TiBatchWrite(@transient val df: DataFrame,
...
+          .map(_._2.head)
       }
-
-      rddGroupByKey.map {
-        case (_, iterable) =>
-          // remove duplicate rows if key equals
-          iterable.head
-      }
+    }
+    if (options.deduplicate) {
```
windtalker: suggest to use `options.deduplicate || options.replace` since when in replace mode, user usually do not expect data conflicts error

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -304,15 +299,41 @@ class TiBatchWrite(@transient val df: DataFrame,
...
+        }
+      }
+
+      val distinctWrappedRowRdd = deduplicate(wrappedRowRdd)
+
+      val rddDeletion = generateDataToBeRemovedRdd(distinctWrappedRowRdd, startTimeStamp)
+      if (!options.replace && !rddDeletion.isEmpty()) {
+        throw new TiBatchWriteException("data to be inserted has conflicts with TiKV data")
+      }
+      generateKV(distinctWrappedRowRdd, remove = false) ++ generateKV(rddDeletion, remove = true)
```
windtalker: For duplicated data, do we need to delete the old version first, then insert the new version? If yes, then how can we ensure that at runtime?

windtalker: Assume a key need to be written 2 times in this batchwrite(delete and insert), during 2pc, what happens if one of the write is chosen as the primary key(or more inopportune, what happens if the insert action of the key is chosen as the primary key during pre-write)?

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -428,6 +460,39 @@ class TiBatchWrite(@transient val df: DataFrame,
     }
   }
 
+  private def generateDataToBeRemovedRdd(rdd: RDD[WrappedRow], startTs: TiTimestamp) = {
+    rdd
```
windtalker: Do we need the consider the duplicated remove case?(a row is removed more than one times as it conflicts with more than one input rows)

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
zhexuany: In that case, we can actually remove `deduplicate` option. @marsishandsome What do you think? As far as I can recall, you think we need `deduplicate` option and let users choose to set deduplicate or not.

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
zhexuany: The same key will not be written twice in one batch.

zhexuany: A new test case is added to cover this. It seems ok to merge two rdd.

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
zhexuany: Data to be inserted will be deduplicated first if the table has pk or unique index.

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -304,15 +299,52 @@ class TiBatchWrite(@transient val df: DataFrame,
...
+      } else {
+        val start = getAutoTableIdStart(tiRowRdd.count)
+        tiRowRdd.zipWithIndex.map { data =>
+          WrappedRow(data._1, data._2 + start)
+        }
+      }
+
+      val distinctWrappedRowRdd = deduplicate(wrappedRowRdd)
+
+      val rddDeletion = generateDataToBeRemovedRdd(distinctWrappedRowRdd, startTimeStamp)
```
marsishandsome: better to use one word, choose one between `Deletion` and `Remove`

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
marsishandsome: is there any test cases?

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -428,6 +471,39 @@ class TiBatchWrite(@transient val df: DataFrame,
     }
   }
 
+  private def generateDataToBeRemovedRdd(rdd: RDD[WrappedRow], startTs: TiTimestamp) = {
+    rdd
+      .map { wrappedRow =>
```
marsishandsome: change `rdd.map` to `rdd.mapPartitions`, cause if use `rdd.map` the following code will be called `rdd.size` times

`val snapshot = TiSessionCache.getSession(tiConf).createSnapshot(startTs)`

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
marsishandsome: what if row duplicates in `rddDeletion`? should we deduplicate? pls add tests to ensure correctness.

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -304,15 +299,52 @@ class TiBatchWrite(@transient val df: DataFrame,
...
+
+      val rddDeletion = generateDataToBeRemovedRdd(distinctWrappedRowRdd, startTimeStamp)
+      if (!options.replace && !rddDeletion.isEmpty()) {
+        throw new TiBatchWriteException("data to be inserted has conflicts with TiKV data")
+      }
+      val mergedRDD = generateKV(distinctWrappedRowRdd, remove = false) ++ generateKV(rddDeletion, remove = true)
+
+      mergedRDD.groupByKey().map {
+        case(key, iterable) =>
+          // if rdd contains same key, it means we need delete and update the value associated the
```
marsishandsome: would u pls give an example which will cause `iterable.size >= 2` ?

core/src/test/scala/com/pingcap/tispark/datasource/AddingIndexUpsertSuite.scala
zhexuany: Did.

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
zhexuany: Resolved.

tikv-client/src/main/java/com/pingcap/tikv/codec/TableCodec.java
```diff
@@ -47,26 +49,42 @@
...
-    if (values.length == 0) {
+    if (cdo.toBytes().length == 0) {
       return new byte[] {Codec.NULL_FLAG};
     }
 
     return cdo.toBytes();
   }
 
-  public static Row decodeRow(byte[] value, List<TiColumnInfo> cols) {
+  public static Row decodeRow(byte[] value, long handle, TiTableInfo tableInfo) {
```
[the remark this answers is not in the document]
zhexuany: Not need to since we pass titableinfo into this function.

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -122,12 +124,13 @@ class TiBatchWrite(@transient val df: DataFrame,
     splitKeys
   }
 
-  private def calcRDDSize(rdd: RDD[Row]): Long =
+  private def calcRDDSize(rdd: RDD[TiRow]): Long =
     rdd
-      .map(_.mkString(",").getBytes("UTF-8").length.toLong)
+    // TODO make estimation more correctly
+      .map(row => row.fieldCount() * 64)
```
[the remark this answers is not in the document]
zhexuany: https://github.com/pingcap/tispark/issues/870

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
zhexuany: key in rdd deletion is unique.

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
zhexuany: We discussed f2f.

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
zhexuany: It is planned to be done when we support batch get.

marsishandsome: it can be done without batch get.

tikv-client/src/main/java/com/pingcap/tikv/codec/TableCodec.java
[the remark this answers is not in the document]
marsishandsome: what if the function user input the following arguments:
1. handle = -1
2. tabelInfo (pk is not handle)

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
zhexuany: Yes. tests defined in AddingIndexReplaceSuite.

docs/datasource_api_userguide.md
```diff
@@ -23,12 +23,18 @@ Since TiDB is a database that supports transaction, TiDB Spark Connector also su
...
+If `replace` is true, then 
+* if primary key or unique index exists in db, data will be updated
+* if no same primary key or unique index exists, data will be inserted.
+
+If `replace` is false, then
+* if primary key or unique index exists in db, data having conflicts expects an expection.
+* if no same primary key or unique index exists, data will be inserted.
 
 | SaveMode | Support | Semantics |
 | -------- | ------- | --------- |
```
marsishandsome: pls update this line
```
TiSpark's `Append` means upsert. If primary key is same, data will be updated; if no same primary key exists, data will be inserted.
```

windtalker:
LGTM

PR #825, opened 2019-06-14
[TISPARK-105] Support refresh lock request

birdstorm:
Support refresh lock request so that we can use a small TTL for long-lasting transaction write

Related PRs:
https://github.com/pingcap/kvproto/pull/414
https://github.com/tikv/tikv/pull/4831


birdstorm:
Close due to unsupported TiDB behavior of resolve lock.

PR #831, opened 2019-06-18
add column map test for batch write

windtalker:
fix #753 , 
1. add batch write column map test
2. fix batch write bug when column order is not the same as original table schema(on behalf of zhexuany)


marsishandsome:
@zhexuany PTAL

PR #832, opened 2019-06-18
check schema change before commit

marsishandsome:
if(schema not change) {
commit
} else {
rollback
}


marsishandsome:
only support writing to such tidb that support lock table

PR #836, opened 2019-06-19
batch write support write to Timestamp Type

marsishandsome:
support write to `Timestamp` Type, solve timezone problem


marsishandsome:
@zhexuany @birdstorm PTAL

marsishandsome:
see https://github.com/pingcap/tispark/pull/842

Issue #840, opened 2019-06-19
add time zone document


marsishandsome:
1. add time zone document
2. remove config spark.tispark.request.timezone.offset, use JVM Default Timezone

PR #842, opened 2019-06-20
add time zone document and support write to Timestamp Type

marsishandsome:
1. batch write support write to Timestamp Type
2. add time zone document
3. remove config `spark.tispark.request.timezone.offset`, use JVM Default Timezone

close https://github.com/pingcap/tispark/issues/840

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
+++ b/core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
@@ -39,6 +39,7 @@ import org.apache.spark.sql.catalyst.plans.logical._
 import org.apache.spark.sql.execution._
 import org.apache.spark.sql.execution.datasources.LogicalRelation
 import org.apache.spark.sql.internal.SQLConf
+import org.joda.time.{DateTime, DateTimeZone}
 
 import scala.collection.JavaConverters._
 import scala.collection.mutable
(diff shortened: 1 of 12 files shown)


core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -82,13 +83,17 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
...
     sqlConf.getConfString(TiConfigConst.COPROCESS_STREAMING, "false").toLowerCase.toBoolean
 
-  private def timeZoneOffset(): Int =
-    sqlConf
-      .getConfString(
-        TiConfigConst.KV_TIMEZONE_OFFSET,
-        String.valueOf(ZonedDateTime.now.getOffset.getTotalSeconds)
-      )
-      .toInt
+  private lazy val timeZoneOffset: Int = {
```
zhexuany: what is the unit of this?

birdstorm: seconds. (my understanding)

core/src/test/scala/com/pingcap/tispark/convert/ToTimestampSuite.scala
```diff
@@ -87,15 +86,16 @@ class ToTimestampSuite extends BaseDataSourceTest("test_data_type_convert_to_tim
...
   }
 
-  // TODO: to support
-  ignore("Test Convert from java.sql.Date to TIMESTAMP") {
+  test("Test Convert from java.sql.Date to TIMESTAMP") {
     // success
     // java.sql.Date -> TIMESTAMP
     compareTiDBWriteWithJDBC {
-      case (writeFunc, _) =>
+      case (writeFunc, "tidbWrite") =>
```
birdstorm: what does this change stand for?

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -82,13 +82,14 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
...
+    val tz = DateTimeZone.getDefault
```
birdstorm: I wonder whether this is a correct behavior for `timeZoneOffset` because `timeZoneOffset` is not a lazy val, which means it may not follow the timezone changes if user use `DateTimeZone.setDefault`.

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
marsishandsome: seconds

core/src/test/scala/com/pingcap/tispark/convert/ToTimestampSuite.scala
marsishandsome: only test: write using batch write 
skip test: write using spark jdbc

because spark jdbc write does not use local time zone, which means it's result is wrong!

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
marsishandsome: agree

PR #845, opened 2019-06-20
Fix behavior of estimateTableSize

### LINKED ISSUE #844: Tables might be incorrectly broadcasted if statistics fail to calculate table size
Table size estimation might fail if statistics information is incorrectly read/written, causing large tables be broadcasted incorrectly.

birdstorm:
Close #844 
When estimated count is 0, it is possible that statistics information is not read correctly. Just set its estimate count to `Long.Max` since we do not want to broadcast this table.


birdstorm:
already merged.

Issue #770, opened 2019-05-28
`org.joda.time.DateTime` cannot hold mysql type `TIMESTAMP`


marsishandsome:
in `class AbstractDateTimeType`, we use `org.joda.time.DateTime`  to  store mysql type `TIMESTAMP`

```
protected void encodeKey(CodecDataOutput cdo, Object value) {
    DateTime dt = Converter.convertToDateTime(value);
    DateTimeCodec.writeDateTimeFully(cdo, dt, getTimezone());
  }
```

but `org.joda.time.DateTime` can only hold `ms`,  while  `TIMESTAMP` can hold `ns`.

PR #853, opened 2019-06-21
extend joda's DateTime to support micro second in order to hold Timestamp

marsishandsome:
let joda's DateTime hold micro second

close https://github.com/pingcap/tispark/issues/770

### DIFF
--- a/core/src/test/scala/com/pingcap/tispark/convert/ToDateTimeSuite.scala
+++ b/core/src/test/scala/com/pingcap/tispark/convert/ToDateTimeSuite.scala
@@ -30,20 +30,20 @@ class ToDateTimeSuite extends BaseDataSourceTest("test_data_type_convert_to_date
 
     val readA: java.sql.Timestamp = java.sql.Timestamp.valueOf("2019-11-11 11:11:11")
     val readB: java.sql.Timestamp = java.sql.Timestamp.valueOf("1990-01-01 01:01:01.999")
-    val readC: java.sql.Timestamp = java.sql.Timestamp.valueOf("1995-05-01 21:21:21.6")
+    val readC: java.sql.Timestamp = java.sql.Timestamp.valueOf("1995-05-01 21:21:21.666666")
 
     readRow1 = Row(1, null, null, null)
     readRow2 = Row(2, readA, readB, readC)
   }
 
-  test("Test Convert from java.lang.Long to DATETIME") {
+  ignore("Test Convert from java.lang.Long to DATETIME") {
     // success
     // java.lang.Long -> DATETIME
     compareTiDBWriteWithJDBC {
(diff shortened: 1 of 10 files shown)


marsishandsome:
@birdstorm @zhexuany PTAL

core/src/test/scala/com/pingcap/tispark/convert/ToDateTimeSuite.scala
```diff
@@ -30,20 +30,20 @@ class ToDateTimeSuite extends BaseDataSourceTest("test_data_type_convert_to_date
...
     val readB: java.sql.Timestamp = java.sql.Timestamp.valueOf("1990-01-01 01:01:01.999")
-    val readC: java.sql.Timestamp = java.sql.Timestamp.valueOf("1995-05-01 21:21:21.6")
+    val readC: java.sql.Timestamp = java.sql.Timestamp.valueOf("1995-05-01 21:21:21.666666")
 
     readRow1 = Row(1, null, null, null)
     readRow2 = Row(2, readA, readB, readC)
   }
 
-  test("Test Convert from java.lang.Long to DATETIME") {
+  ignore("Test Convert from java.lang.Long to DATETIME") {
```
birdstorm: why is this test ignored?

tikv-client/src/main/java/com/pingcap/tikv/types/DateTimeType.java
```diff
@@ -56,7 +57,8 @@ protected Object doConvertToTiDBType(Object value)
       throws ConvertNotSupportException {
     java.sql.Timestamp result;
     if (value instanceof Long) {
-      result = new java.sql.Timestamp((Long) value);
+      throw new ConvertNotSupportException(value.getClass().getName(), this.getClass().getName());
```
birdstorm: I think this logic should be something like `someExtendedTimestamp.toTimestamp()`?

core/src/test/scala/com/pingcap/tispark/convert/ToDateTimeSuite.scala
marsishandsome: 1. do not sure what long means? maybe microseconds or milliseconds from 1970-01-01
2. should timezone be considered?

will support after investigate

zhexuany: what's the behavior in tidb side?

marsishandsome: after investigate mysql do not support insert long type to DATETIME

birdstorm: I believe we can delete this test or check if it is not supported.

PR #862, opened 2019-06-24
Fix incorrect timestamp of `tidbMapDatabase`

### LINKED ISSUE #861: Incorrect timestamp setting in `tidbMapDatabase()`
If the timestamp was get when tidbMapDatabase was called, it may be outdated when trying to read new data afterwards. The timestamp should be get only when planning.

birdstorm:
close #861


marsishandsome:
line 162 also delete?

birdstorm:
> line 162 also delete?

Yes! Thanks for pointing out!

Issue #868, opened 2019-06-26
region and store should get from same task


marsishandsome:
in class `DAGIterator`, region and store should get from different task
```
TiRegion region = task.getRegion();
 Metapb.Store store = regionTask.getStore();
```

PR #869, opened 2019-06-26
fix bug: get region and store from same RegionTask

marsishandsome:
close https://github.com/pingcap/tispark/issues/868


marsishandsome:
@zhexuany @windtalker PTAL

windtalker:
LGTM

zhexuany:
Good Catch.

PR #874, opened 2019-06-26
fix java 9 build and pump grpc to 1.19.0

ilovesoup:
Fix java 9 build by adding missing annotation dep. Also pump version of grpc to 1.19 as TiKV.

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/tispark/874)


ilovesoup:
It seems other problems still block this. Close this for now.

PR #875, opened 2019-06-27
Remove splitTasks as a temporary fix for unstable index read

birdstorm:
Temporarily fix index unstable issue over large dataset by removing splitTasks. It may be added back later, but we need to check where went wrong first.


marsishandsome:
@zhexuany PTAL

Issue #877, opened 2019-07-01
count error when using double read and task split


marsishandsome:
there's may be bugs if `regionTasks's size > 1` 

```
public static CoprocessIterator<Row> getRowIterator(
      TiDAGRequest req, List<RegionTask> regionTasks, TiSession session)
```

PR #878, opened 2019-07-01
fix count error, if advanceNextResponse is empty, we should read next region

marsishandsome:
closing https://github.com/pingcap/tispark/issues/877

In DATIterator, if advanceNextResponse is empty, we should read next region instead of just return false.

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/DAGIterator.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/DAGIterator.java
@@ -131,14 +131,16 @@ public abstract class DAGIterator<T> extends CoprocessIterator<T> {
    * @return
    */
   private boolean readNextRegionChunks() {
-    while (true) {
-      if (eof || regionTasks == null || taskIndex >= regionTasks.size()) {
-        return false;
-      }
+    while (hasNextRegionTask()) {
       if (doReadNextRegionChunks()) {
         return true;
       }
     }
+    return false;
+  }
+
+  private boolean hasNextRegionTask() {
+    return !(eof || regionTasks == null || taskIndex >= regionTasks.size());
   }
 
   private boolean doReadNextRegionChunks() {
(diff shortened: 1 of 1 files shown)


tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/DAGIterator.java
```diff
@@ -143,7 +143,11 @@ private boolean readNextRegionChunks() {
     }
 
     taskIndex++;
-    return advanceNextResponse();
+    if (advanceNextResponse()) {
```
zhexuany: Please add some comments. This is very hard to catch.

birdstorm:
Good job! I will try to reproduce it.

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/DAGIterator.java
marsishandsome: good idea

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/DAGIterator.java
```diff
@@ -124,6 +124,21 @@ private boolean advanceNextResponse() {
...
+      } // else {
+      // if doReadNextRegionChunks returns false
+      // readNextRegionChunks should not just return false
+      // readNextRegionChunks should read next region chunks
+      // }
+    }
+  }
+
+  private boolean doReadNextRegionChunks() {
     if (eof || regionTasks == null || taskIndex >= regionTasks.size()) {
```
zhexuany: Do we need this if? It seems we did this predicate in `regionNextRegionChunks`

marsishandsome: yes, can delete

marsishandsome: done

PR #879, opened 2019-07-01
Revert "Remove splitTasks as a temporary fix for unstable index read"

marsishandsome:
Reverts pingcap/tispark#875

cause we find the bug: https://github.com/pingcap/tispark/pull/878


tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/CoprocessIterator.java
```diff
@@ -35,7 +35,7 @@
 public abstract class CoprocessIterator<T> implements Iterator<T> {
   protected final TiSession session;
   protected final List<RegionTask> regionTasks;
-  protected final DAGRequest dagRequest;
+  protected DAGRequest dagRequest;
```
birdstorm: maybe we can leave this change.

marsishandsome: maybe in other PR to add `final`

PR #899, opened 2019-07-04
fix count error, if advanceNextResponse is empty, we should read next region

marsishandsome:
cherry pick https://github.com/pingcap/tispark/pull/878


marsishandsome:
@birdstorm PTAL

PR #901, opened 2019-07-04
release 1.2.1

marsishandsome:
close https://github.com/pingcap/tispark/issues/900

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -2,7 +2,7 @@ from setuptools import setup
 setup(
     name = 'pytispark',
     packages = ['pytispark'],
-    version = '1.2.1',
+    version = '1.1',
     description = 'TiSpark support for python',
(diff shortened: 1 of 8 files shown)


python/setup.py
```diff
@@ -2,7 +2,7 @@
 setup(
     name = 'pytispark',
     packages = ['pytispark'],
-    version = '1.1',
+    version = '1.2.1',
```
birdstorm: if you update pytispark version, it should be released on pyti

marsishandsome: shall we release pytispark?

birdstorm: we may need to release pytispark when changes involve in logic or spark version.

marsishandsome: reverted

marsishandsome:
@birdstorm PTAL

Issue #904, opened 2019-07-04
integrate delay-clean-table-lock


marsishandsome:
see https://github.com/pingcap/tidb/pull/11038

The logic of TiSpark to use the table lock is:

1. lock table
2. prewrite data.
3. check whether the locked table session is closed.
4. if the session doesn't close, do commit.
5. if the session was closed by unexpected, do a rollback.

The problem is, What if the session was closed by unexpected between step 3 and 4, It will cause some problem.

TiSpark need to delayed release table lock before do commit.

PR #905, opened 2019-07-05
add delay clean table lock

marsishandsome:
close https://github.com/pingcap/tispark/issues/904

The logic of TiSpark to use the table lock is:

1. lock table
2. prewrite data.
3. check whether the locked table session is closed.
4. if the session doesn't close, do commit.
5. if the session was closed by unexpected, do a rollback.

The problem is, What if the session was closed by unexpected between step 3 and 4, It will cause some problem.

TiSpark need to delayed release table lock before do commit.

### DIFF
--- a/core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
+++ b/core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
@@ -42,8 +42,8 @@ import scala.collection.mutable
 
 object TiBatchWrite {
   // Milliseconds
-  private val MIN_DELAY_CLEAN_TABLE_LOCK = 10000
-  private val DELAY_CLEAN_TABLE_LOCK_AND_COMMIT_BACKOFF_DELTA = 4000
+  private val MIN_DELAY_CLEAN_TABLE_LOCK = 60000
+  private val DELAY_CLEAN_TABLE_LOCK_AND_COMMIT_BACKOFF_DELTA = 30000
   private val PRIMARY_KEY_COMMIT_BACKOFF = MIN_DELAY_CLEAN_TABLE_LOCK - DELAY_CLEAN_TABLE_LOCK_AND_COMMIT_BACKOFF_DELTA
 
   type SparkRow = org.apache.spark.sql.Row
--- a/core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
+++ b/core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
@@ -41,6 +41,11 @@ import scala.collection.JavaConverters._
 import scala.collection.mutable
 
 object TiBatchWrite {
+  // Milliseconds
+  private val MIN_DELAY_CLEAN_TABLE_LOCK = 10000
+  private val DELAY_CLEAN_TABLE_LOCK_AND_COMMIT_BACKOFF_DELTA = 4000
+  private val PRIMARY_KEY_COMMIT_BACKOFF = MIN_DELAY_CLEAN_TABLE_LOCK - DELAY_CLEAN_TABLE_LOCK_AND_COMMIT_BACKOFF_DELTA
+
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -146,7 +151,26 @@ class TiBatchWrite(@transient val df: DataFrame,
 
     // lock table
     tiDBJDBCClient = new TiDBJDBCClient(TiDBUtils.createConnectionFactory(options.url)())
-    isEnableTableLock = tiDBJDBCClient.isEnableTableLock
+    isEnableTableLock = {
+      if (tiDBJDBCClient.isEnableTableLock) {
+        if (tiDBJDBCClient.getDelayCleanTableLock >= MIN_DELAY_CLEAN_TABLE_LOCK) {
```
birdstorm: Since we have logic to check “delay clean table lock” here, I think it’s better to make MIN_DELAY_CLEAN_TABLE_LOCK configurable.

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -324,7 +348,7 @@ class TiBatchWrite(@transient val df: DataFrame,
         s"invalid transaction tso with startTs=$startTs, commitTs=$commitTs"
       )
     }
-    val commitPrimaryBackoff = ConcreteBackOffer.newCustomBackOff(BackOffer.BATCH_COMMIT_BACKOFF)
+    val commitPrimaryBackoff = ConcreteBackOffer.newCustomBackOff(PRIMARY_KEY_COMMIT_BACKOFF)
```
birdstorm: Please keep in mind that it does not mean backOff will be limited in PRIMARY_KEY_COMMIT_BACKOFF. It is still possible that commit duration exceeds PRIMARY_KEY_COMMIT_BACKOFF because we can only check the time duration when we receive a response.

birdstorm: Please keep in mind that it does not mean backOff will be limited in PRIMARY_KEY_COMMIT_BACKOFF. It is still possible that commit duration exceeds PRIMARY_KEY_COMMIT_BACKOFF because we can only check the time duration when we receive a response.

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
marsishandsome: pls give a reason, why make it configurable?

birdstorm: Because delayCleanTableLock is configurable in TiDB, the user may want to reduce table lock duration by TiDB configuration.

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
marsishandsome: increase backoff time as discussed

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
marsishandsome: not necessary as discussed

tikv-client/src/main/java/com/pingcap/tikv/TiDBJDBCClient.java
```diff
@@ -45,6 +47,20 @@ public boolean isEnableTableLock() throws IOException, SQLException {
     return (Boolean) enableTableLock;
   }
 
+  /**
+   * get delay clean table lock
```
zhexuany: This comment seems wired.

marsishandsome: changed

marsishandsome:
@birdstorm PTAL again

Issue #906, opened 2019-07-05
forbid write to table with generated column


marsishandsome:
cause currently tispark does not support

PR #907, opened 2019-07-05
forbid write to table with generated column

marsishandsome:
close https://github.com/pingcap/tispark/issues/906


birdstorm:
Consider also update the document about what scenario do we support and what we do not?

marsishandsome:
> Consider also update the document about what scenario do we support and what we do not?

yes, i will update the document before release. (we also do not support partition table which will also be included in document.)

PR #909, opened 2019-07-10
implement split index region

zhexuany:
In this PR, we add support of splitting index region if schema having any index.

### DIFF
--- a/core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
+++ b/core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
@@ -723,6 +723,14 @@ class TiBatchWrite(@transient val df: DataFrame,
     tiTableInfo.getId
   }
 
+  private def estimateRegionSplitNumForIndex(wrappedRowRdd: RDD[WrappedRow],
+                                             tiIndexInfo: TiIndexInfo) = {
+    //TODO refine this https://github.com/pingcap/tispark/issues/891
+    val rowSize = tiIndexInfo.getIndexColumnSize
+    //TODO: replace 96 with actual value read from pd https://github.com/pingcap/tispark/issues/890
+    (wrappedRowRdd.count() * rowSize) / (96 * 1024 * 1024)
+  }
+
   private def estimateRegionSplitNum(wrappedRowRdd: RDD[WrappedRow]) = {
     //TODO refine this https://github.com/pingcap/tispark/issues/891
     val rowSize = tiTableInfo.getEstimatedRowSizeInByte
@@ -732,25 +740,35 @@ class TiBatchWrite(@transient val df: DataFrame,
 
   private def splitIndexRegion(wrappedRowRdd: RDD[WrappedRow]) = {
     if (options.enableRegionSplit && isEnableSplitRegion) {
-      val regionSplitNum = if (options.regionSplitNum != 0) {
-        options.regionSplitNum
-      } else {
-        estimateRegionSplitNum(wrappedRowRdd)
-      }
-
-      val sampledDataList =
-        wrappedRowRdd.takeSample(withReplacement = false, num = regionSplitNum.toInt).toList
       val indices = tiTableInfo.getIndices.asScala
-      indices.foreach { index =>
+      val regionSplitNums = indices.map { index =>
+        if (options.regionSplitNum != 0) {
+          options.regionSplitNum
+        } else {
+          estimateRegionSplitNumForIndex(wrappedRowRdd, index)
+        }
+      }
+      val sampledDataRdds =
+        regionSplitNums.map { num =>
+          wrappedRowRdd.takeSample(withReplacement = false, num = num.toInt)
+        }.toList
+
+      indices.zipWithIndex.foreach { indexWithIdx =>
+        val index = indexWithIdx._1
+        val idx = indexWithIdx._2
         val indexCols = index.getIndexColumns
-        val splitIndicesList = sampledDataList.map { value =>
-          val colBuffer = mutable.ListBuffer.empty[String]
-          for (i <- 0 until indexCols.size()) {
-            val col = indexCols.get(i)
-            colBuffer += value.row.get(col.getOffset, null).toString
+
+        val splitIndicesList = sampledDataRdds(idx)
+          .map { value =>
+            val colBuffer = mutable.ListBuffer.empty[String]
+            for (i <- 0 until indexCols.size()) {
+              val col = indexCols.get(i)
+              colBuffer += value.row.get(col.getOffset, null).toString
+            }
+            colBuffer.toList.asJava
           }
-          colBuffer.toList.asJava
-        }.asJava
+          .toList
+          .asJava
 
         tiDBJDBCClient
           .splitIndexRegion(
--- a/core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
+++ b/core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
@@ -779,7 +779,6 @@ class TiBatchWrite(@transient val df: DataFrame,
           )
       }
     }
-
   }
 
   // when data to be inserted is too small to do region split, we check is user set region split num.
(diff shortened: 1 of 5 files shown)


tikv-client/src/main/java/com/pingcap/tikv/TiDBJDBCClient.java
```diff
@@ -110,6 +110,44 @@ public boolean splitTableRegion(
...
   }
 
+  public boolean splitIndexRegion(
+      String dbName, String tblName, String idxName, List<List<String>> splitPoints)
+      throws SQLException {
+    // SPLIT TABLE t1 INDEX idx4 ("a", "2000-01-01 00:00:01"), ("b", "2019-04-17 14:26:19"), ("c",
+    // "");
+    StringBuilder sb = new StringBuilder();
+    sb.append("split table ")
+        .append(dbName)
```
birdstorm: If you wish to compose the query via string, do not forget to wrap **`** around your table and database names.

tikv-client/src/main/java/com/pingcap/tikv/TiBatchWriteUtils.java
```diff
@@ -15,15 +15,31 @@
...
+  public static List<TiRegion> getRegionByIndex(
+      TiSession session, TiTableInfo table, TiIndexInfo index) {
+    ArrayList<TiRegion> regionList = new ArrayList<>();
+    Key min = IndexKey.toIndexKey(table.getId(), index.getId());
+    Key max = min.nextPrefix();
+
+    while (min.compareTo(max) < 0) {
+      TiRegion region = session.getRegionManager().getRegionByKey(min.toByteString());
+      regionList.add(region);
+      min = Key.toRawKey(region.getEndKey());
```
birdstorm: somewhat dangerous if region splits or changes...

zhexuany: This only used for testing, I believe.

tikv-client/src/main/java/com/pingcap/tikv/TiDBJDBCClient.java
zhexuany: I will fix this.

zhexuany: DONE

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -710,9 +711,51 @@ class TiBatchWrite(@transient val df: DataFrame,
...
+    //TODO: replace 96 with actual value read from pd https://github.com/pingcap/tispark/issues/890
+    (wrappedRowRdd.count() * rowSize) / (96 * 1024 * 1024)
+  }
+
+  private def splitIndexRegion(wrappedRowRdd: RDD[WrappedRow]) = {
+    if (options.enableRegionSplit && isEnableSplitRegion) {
+      val regionSplitNum = if (options.regionSplitNum != 0) {
+        options.regionSplitNum
+      } else {
+        estimateRegionSplitNum(wrappedRowRdd)
```
marsishandsome: current implementation of `estimateRegionSplitNum` is only suitable for `splitTableRegion`, better to implement a special version for `splitIndexRegion`

because `index data size` is usually much smaller than `table data size`

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -710,9 +711,51 @@ class TiBatchWrite(@transient val df: DataFrame,
...
+      }
+
+      val sampledDataList =
+        wrappedRowRdd.takeSample(withReplacement = false, num = regionSplitNum.toInt).toList
```
marsishandsome: is it possible that 
1. calculate `min` and `max`
2. evenly divided
just like region split for table?

zhexuany: It is not easy to do that right now. We have to know the min max of each column which we have not implemented yet.

tikv-client/src/main/java/com/pingcap/tikv/TiBatchWriteUtils.java
zhexuany: `getRegionByIndex` and `getRegionByTable` is almost the same. Should we move them to `core/src/test` ?

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
zhexuany: It is not feasible to do that at this point, we have to first implement the min and max of each column type.

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
zhexuany: I had the same concern, let me figure out a way to resolve this.

tikv-client/src/main/java/com/pingcap/tikv/TiBatchWriteUtils.java
marsishandsome: as u said, `getRegionByIndex` is only used for testing, if we put here, other developer may use it as a mistake

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
marsishandsome: can we use spark's `min` and `max` ？

tikv-client/src/main/java/com/pingcap/tikv/TiBatchWriteUtils.java
zhexuany: My point is that `getRegionByIndex` is identical `getRegionByTable`, what should I do for `getRegionByTable`. `getRegionByTable` is used in `TiBatchWrite`.

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
zhexuany: Each column type has it is own min and max. For example, if we have a time column, the min of this column is ` -838:59:59` and the max is `838:59:59`. Using spark's min and max is not relevant here.

marsishandsome: user interface is DataFrame, which means all user provided data is spark sql supported type, which means we can use spark `min` and `max`.

zhexuany: That requires implementing orders which is something we do not have right now.

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -710,9 +711,51 @@ class TiBatchWrite(@transient val df: DataFrame,
...
+  private def estimateRegionSplitNum(wrappedRowRdd: RDD[WrappedRow]) = {
+    //TODO refine this https://github.com/pingcap/tispark/issues/891
+    val rowSize = tiTableInfo.getEstimatedRowSizeInByte
+    //TODO: replace 96 with actual value read from pd https://github.com/pingcap/tispark/issues/890
+    (wrappedRowRdd.count() * rowSize) / (96 * 1024 * 1024)
+  }
+
+  private def splitIndexRegion(wrappedRowRdd: RDD[WrappedRow]) = {
+    if (options.enableRegionSplit && isEnableSplitRegion) {
+      val regionSplitNum = if (options.regionSplitNum != 0) {
```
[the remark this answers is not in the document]
zhexuany: Resolved

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
zhexuany: resolved.

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
marsishandsome: can we calculate `min` and `max` after encoding?

PR #911, opened 2019-07-11
ignore case when compare column name

marsishandsome:
close https://github.com/pingcap/tispark/issues/910


marsishandsome:
@zhexuany @birdstorm PTAL

PR #913, opened 2019-07-11
add lock ttl as parameter

marsishandsome:
close https://github.com/pingcap/tispark/issues/912

1. add lock ttl as user parameter
2. fix bug: lock ttl time unit is millisecond instead of second
3. change skipCommitSecondaryKey default value to false (if primary key is commit, but secondary key is not commit and ttl is not reached, tidb cannot resolve the lock)

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/TwoPhaseCommitter.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/TwoPhaseCommitter.java
@@ -96,8 +96,8 @@ public class TwoPhaseCommitter {
    */
   private static final int TXN_COMMIT_BATCH_SIZE = 768 * 1024;
 
-  /** unit is second */
-  private static final long DEFAULT_BATCH_WRITE_LOCK_TTL = 3600;
+  /** unit is millisecond */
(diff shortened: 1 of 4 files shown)


tikv-client/src/main/java/com/pingcap/tikv/TwoPhaseCommitter.java
```diff
@@ -96,8 +96,8 @@ public BytePairWrapper(byte[] key, byte[] value) {
    */
   private static final int TXN_COMMIT_BATCH_SIZE = 768 * 1024;
 
-  /** unit is second */
-  private static final long DEFAULT_BATCH_WRITE_LOCK_TTL = 3000;
+  /** unit is millisecond */
+  private static final long DEFAULT_BATCH_WRITE_LOCK_TTL = 3600000;
```
birdstorm: 3.6 seconds? what does it stand for?

marsishandsome: 3600000ms = 3600s = 1h

birdstorm: alright then.

Issue #914, opened 2019-07-11
add parameter `writeConcurrency` to control write concurrency


marsishandsome:
too much write concurrency will cause tikv busy

PR #915, opened 2019-07-11
add write concurrency parameter

marsishandsome:
close https://github.com/pingcap/tispark/issues/914


marsishandsome:
> Mostly LGTM except PR description does not agree with the code.

cause this PR is based on https://github.com/pingcap/tispark/pull/913

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -765,7 +765,7 @@ class TiRegionPartitioner(regions: List[TiRegion]) extends Partitioner {
       val region = regions(i)
       val range = KeyRangeUtils.makeRange(region.getStartKey, region.getEndKey)
       if (range.contains(rawKey)) {
-        return i
+        return if (writeConcurrency <= 0) i else i % writeConcurrency
```
birdstorm: why not `return i % numPartitions`

marsishandsome: good point

PR #922, opened 2019-07-12
Fix improper response with region error

### LINKED ISSUE #921: Some requests missed to deal with unretryable RegionError
It is possible that some region error was ignored when an unretryable Error encounters in response, resulting in an incorrect empty result rather than error thrown.

Root cause for #918

birdstorm:
Close #921

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
@@ -244,7 +244,7 @@ public class RegionStoreClient extends AbstractGRPCClient<TikvBlockingStub, Tikv
               resp -> resp.hasRegionError() ? resp.getRegionError() : null);
       ScanResponse resp = callWithRetry(backOffer, TikvGrpc.METHOD_KV_SCAN, request, handler);
       if (isScanSuccess(backOffer, resp)) {
-        return doScan(backOffer, resp);
+        return doScan(resp);
       }
     }
   }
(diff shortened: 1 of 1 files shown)


melin:
修复的代码，还是会出现下面问题：

tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
```diff
@@ -122,6 +123,24 @@ public ByteString get(BackOffer backOffer, ByteString key, long version)
...
   }
 
+  private boolean checkLockError(BackOffer backOffer, Kvrpcpb.KeyError error) {
+    if (error.hasLocked()) {
+      Lock lock = new Lock(error.getLocked());
+      boolean ok =
+          lockResolverClient.resolveLocks(
+              backOffer, new ArrayList<>(Collections.singletonList(lock)));
+      if (!ok) {
+        // if not resolve all locks, we wait and retry
```
marsishandsome: `all locks`?  maybe there's only one lock

birdstorm: Maybe it is better to change to `if any lock unresolved`

tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
```diff
@@ -206,51 +215,57 @@ private boolean isGetSuccess(BackOffer backOffer, GetResponse resp)
...
       }
 
-      // TODO: we should retry
-      // fix me
-    }
-
-    if (resp.hasRegionError()) {
-      // TODO, we should redo the split and redo the batchGet
-      throw new RegionException(resp.getRegionError());
+      // FIXME: we should retry
```
zhexuany: Do we have a issue for this `FIXME` ?

birdstorm: I will create an issue for these lines.

birdstorm:
@marsishandsome @zhexuany PTAL thx.

tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
birdstorm: #927 @zhexuany

PR #938, opened 2019-07-19
adding shard row id bits test

zhexuany:
This PR adds a test suite to ensure that we can read/write against a table with `shard_row_id_bits`. 

A new test suite is added. In the test suite, we first create a table with `shard_row_id_bits` and then insert via jdbc. Later, we use out batch write api to insert some data. The way to check the correctness is select data from TiKV. 

### Check List 

Tests 
 - Integration test

Side effects
No side effects.

### DIFF
--- a/core/src/test/scala/com/pingcap/tispark/datasource/ShardRowIDBitsSuite.scala
+++ b/core/src/test/scala/com/pingcap/tispark/datasource/ShardRowIDBitsSuite.scala
@@ -1,0 +1,29 @@
+package com.pingcap.tispark.datasource
+
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.types.{IntegerType, StructField, StructType}
+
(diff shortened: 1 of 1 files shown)


core/src/test/scala/com/pingcap/tispark/datasource/ShardRowIDBitsSuite.scala
```diff
@@ -0,0 +1,29 @@
...
+  private val row1 = Row(1)
+  private val row2 = Row(2)
+  private val row3 = Row(3)
+  private val schema = StructType(
+    List(
+      StructField("a", IntegerType)
+    )
+  )
+
+  test("reading from a table with shard_row_id_bits") {
```
marsishandsome:
```suggestion
  test("reading and writing a table with shard_row_id_bits") {
```

Issue #889, opened 2019-07-02
udf print more info


marsishandsome:
tikv version
tidb version
tispark version
spark support version
tidb read support or not
tidb write support not support

marsishandsome:
cannot get the following info without tidb address
```
tikv version
tidb version
tidb read support or not
tidb write support not support
```

PR #943, opened 2019-07-22
add spark version to tispark udf ti_version

marsishandsome:
add spark version to tispark udf ti_version, like
image

close https://github.com/pingcap/tispark/issues/889


marsishandsome:
@zhexuany @birdstorm PTAL

PR #948, opened 2019-07-24
Add tests for primary key

birdstorm:
Add tests for primary key.

Add constraints on primary and unique keys.

### Check List 

Tests 

 - Integration test

Side effects

None

Related changes

None


core/src/test/scala/org/apache/spark/sql/test/generator/Schema.scala
```diff
@@ -49,6 +45,24 @@ case class Schema(database: String,
 
   assert(indexInfo.count(_.isPrimary) <= 1, "more than one primary key exist in schema")
 
+  private val pkIndexInfo = indexInfo.filter(_.isPrimary)
```
zhexuany: primary key defined in index implies primary key is not handle. Have we considered primary key is handle case here?

zhexuany: Never mind. I mislead by the name of `indexInfo`.

PR #954, opened 2019-07-26
add multi-column tests

birdstorm:
add multi column tests

tests

### Check List 

Tests 

 - Integration test

Side effects

 - Possible test performance regression

### DIFF
--- a/core/src/test/scala/org/apache/spark/sql/BaseTestGenerationSpec.scala
+++ b/core/src/test/scala/org/apache/spark/sql/BaseTestGenerationSpec.scala
@@ -19,17 +19,17 @@ package org.apache.spark.sql
 
 trait BaseTestGenerationSpec {
 
-  protected val rowCount: Int = 50
+  protected val rowCount: Int
 
-  private def toString(dataTypes: Seq[String]) = dataTypes.map(_.head).mkString("_")
+  protected val preDescription: String = "Generating Data for "
 
-  protected def getTableName(dataTypes: String*): String = s"test_${toString(dataTypes)}"
+  def getTableName(dataTypes: String*): String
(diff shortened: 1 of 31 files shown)


core/src/test/scala/org/apache/spark/sql/BaseTestGenerationSpec.scala
```diff
@@ -1,6 +1,6 @@
 /*
  *
- * Copyright 2017 PingCAP, Inc.
+ * Copyright 2019 PingCAP, Inc.
```
marsishandsome: what's the rule about when to update the `Copyright year`?

core/src/test/scala/org/apache/spark/sql/MysqlDataTypeSuite.scala
```diff
@@ -1,4 +1,5 @@
 /*
+ *
```
marsishandsome: should we add `license check` in scalafmt?

core/src/test/scala/org/apache/spark/sql/BaseTestGenerationSpec.scala
birdstorm: I forgot to use the right copyright year when I copy-pasted it from another file.

core/src/test/scala/org/apache/spark/sql/MysqlDataTypeSuite.scala
birdstorm: sure.

birdstorm:
Also updates a graph of current tests(should be less messy than before:D) @marsishandsome @zhexuany 

image

Issue #957, opened 2019-07-30
[BUG] read from RANGE Partition throw UnsupportedSyntaxException


marsishandsome:
**Describe the bug**

 read from Range Partition should not throw UnsupportedSyntaxException

**What did you do**
```
test("part pruning on mod function") {
    enablePartitionForTiDB()
    tidbStmt.execute("DROP TABLE IF EXISTS `pt3`")
    tidbStmt.execute("""
                       |CREATE TABLE `pt3` (
                       |  `id` int(11) DEFAULT NULL,
                       |  `name` varchar(50) DEFAULT NULL,
                       |  `purchased` date DEFAULT NULL,
                       |  index `idx_id`(`id`)
                       |) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin
                       |PARTITION BY RANGE (mod(year(purchased), 4)) (
                       |  PARTITION p0 VALUES LESS THAN (1),
                       |  PARTITION p1 VALUES LESS THAN (2),
                       |  PARTITION p2 VALUES LESS THAN (3),
                       |  PARTITION p3 VALUES LESS THAN (MAXVALUE)
                       |)
                     """.stripMargin)

    tidbStmt.execute("insert into `pt3` values(1, 'name', '2019-01-01')")
    refreshConnections()

    spark
    // expected part info only contains one part which is p.
      .sql("select * from pt3")
      .show(200, false)

    spark
    // expected part info only contains one part which is p.
      .sql("select * from pt3 where name = 'name'")
      .show(200, false)
  }
```

```
[504 487 203]: it is not supported right now
com.pingcap.tikv.exception.UnsupportedSyntaxException: [504 487 203]: it is not supported right now
	at com.pingcap.tikv.parser.AstBuilder.visitMathExpressionAtom(AstBuilder.java:212)
	at com.pingcap.tikv.parser.AstBuilder.visitMathExpressionAtom(AstBuilder.java:21)
	at com.pingcap.tikv.parser.MySqlParser$MathExpressionAtomContext.accept(MySqlParser.java:6403)
	at org.antlr.v4.runtime.tree.AbstractParseTreeVisitor.visitChildren(AbstractParseTreeVisitor.java:46)
	at com.pingcap.tikv.parser.MySqlParserBaseVisitor.visitExpressionAtomPredicate(MySqlParserBaseVisitor.java:335)
[...]

PR #960, opened 2019-07-30
fix range partition throw UnsupportedSyntaxException error

marsishandsome:
close https://github.com/pingcap/tispark/issues/957

read from Range Partition should not throw UnsupportedSyntaxException when using unsupported partition function like `mod`, instead tispark should use full table scan

### DIFF
--- a/core/src/test/scala/org/apache/spark/sql/PartitionTableSuite.scala
+++ b/core/src/test/scala/org/apache/spark/sql/PartitionTableSuite.scala
@@ -21,6 +21,34 @@ import org.apache.spark.sql.execution.{CoprocessorRDD, RegionTaskExec}
 class PartitionTableSuite extends BaseTiSparkTest {
   def enablePartitionForTiDB(): Boolean = tidbStmt.execute("set @@tidb_enable_table_partition = 1")
 
+  test("test read from range partition and partition function (mod) is not supported by tispark") {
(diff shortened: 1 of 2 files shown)


core/src/test/scala/org/apache/spark/sql/PartitionTableSuite.scala
```diff
@@ -21,6 +21,34 @@ import org.apache.spark.sql.execution.{CoprocessorRDD, RegionTaskExec}
...
+                     """.stripMargin)
+
+    tidbStmt.execute("insert into `pt` values(1, 'name', '1995-10-10')")
+    refreshConnections()
+
+    judge("select * from pt")
+    judge("select * from pt where name = 'name'")
+    judge("select * from pt where name != 'name'")
+    judge("select * from pt where purchased = date'1995-10-10'")
+    judge("select * from pt where purchased != date'1995-10-10'")
```
birdstorm: I don't think these tests used the range partition information? All of the logical plans fall back to full table scan eventually.

marsishandsome: tispark currently do not support `mod` function, it should use full table scan.

birdstorm: Misread the example. My fault.

PR #970, opened 2019-07-31
Add multiple-column PK tests

birdstorm:
Add multiple-column PK tests

- Add Date/Timestamp random value generation
- Add 6w+ tests concerning primary keys with multiple columns 

### Check List 

Tests 

 - Integration test


.ci/integration_test.groovy
```diff
@@ -90,12 +90,13 @@ def call(ghprbActualCommit, ghprbCommentBody, ghprbPullId, ghprbPullTitle, ghprb
                         sh """
                         cp -R /home/jenkins/git/tispark/. ./
                         git checkout -f ${ghprbActualCommit}
-                        find core/src -name '*Suite*' > test
+                        find core/src -name '*Suite*' | grep -v 'MultiColumnPKDataTypeSuite' > test
+                        shuf test -o  test2
+                        mv test2 test
+                        find core/src -name '*MultiColumnPKDataTypeSuite*' >> test
```
marsishandsome: why add `MultiColumnPKDataTypeSuite` to the end?

birdstorm: We use round robin to split all tests evenly (especially `MultiColumnPKDataTypeSuite`, ordering them together ensures that 36 suites are separated evenly in 18 test containers), also they run longer and should be executed later to minimize test waiting period.

core/src/test/scala/org/apache/spark/sql/test/generator/TestDataGenerator.scala
```diff
@@ -279,8 +283,11 @@ object TestDataGenerator {
   }
 
   def hash(value: Any): String = value match {
-    case null           => "null"
-    case b: Array[Byte] => b.mkString("[", ",", "]")
+    case null                  => "null"
+    case b: Array[Byte]        => b.mkString("[", ",", "]")
+    case t: java.sql.Timestamp =>
+      // timestamp was indexed as Integer when treated as unique key
```
zhexuany: on what situations?

core/src/test/scala/org/apache/spark/sql/BaseTiSparkTest.scala
```diff
@@ -166,12 +164,6 @@ class BaseTiSparkTest extends QueryTest with SharedSQLContext {
     loadTestData()
   }
 
-  protected def initializeTimeZone(): Unit = {
```
zhexuany: why do we remove initializeTimeZone ?

zhexuany: Never mind, I just thought you move to another file.

birdstorm: added in SharedSQLContext

core/src/test/scala/org/apache/spark/sql/test/generator/TestDataGenerator.scala
birdstorm: unique key, as commented

Issue #971, opened 2019-08-01
[BUG] throw java.lang.IncompatibleClassChangeError when submit with spark-2.4.3


marsishandsome:
**Describe the bug**

 throw java.lang.IncompatibleClassChangeError when submit with spark-2.4.3

**What did you do**
```
./bin/spark-shell --jars tispark/assembly/target/tispark-assembly-2.3.0-SNAPSHOT.jar
spark.sql("use tpch_test");
spark.sql("select count(*) from customer").show()
```
```

java.lang.IncompatibleClassChangeError: Found interface org.apache.spark.sql.catalyst.catalog.ExternalCatalog, but class was expected                                                                                                                            
  at org.apache.spark.sql.catalyst.catalog.TiConcreteSessionCatalog.databaseExists(TiConcreteSessionCatalog.scala:38)
  at org.apache.spark.sql.catalyst.catalog.TiCompositeSessionCatalog$$anonfun$catalogOf$1$$anonfun$apply$1.apply(TiCompositeSessionCatalog.scala:96)                                                                                                             
  at org.apache.spark.sql.catalyst.catalog.TiCompositeSessionCatalog$$anonfun$catalogOf$1$$anonfun$apply$1.apply(TiCompositeSessionCatalog.scala:96)                                                                                                             
  at scala.collection.LinearSeqOptimized$class.find(LinearSeqOptimized.scala:113)
  at scala.collection.immutable.List.find(List.scala:84)
  at org.apache.spark.sql.catalyst.catalog.TiCompositeSessionCatalog$$anonfun$catalogOf$1.apply(TiCompositeSessionCatalog.scala:96)                                                                                                                              
  at org.apache.spark.sql.catalyst.catalog.TiCompositeSessionCatalog$$anonfun$catalogOf$1.apply(TiCompositeSessionCatalog.scala:91)                                                                                                                              
  at scala.Option.map(Option.scala:146)
  at org.apache.spark.sql.catalyst.catalog.TiCompositeSessionCatalog.catalogOf(TiCompositeSessionCatalog.scala:90)
  at org.apache.spark.sql.catalyst.catalog.TiCompositeSessionCatalog.setCurrentDatabase(TiCompositeSessionCatalog.scala:130)
  at org.apache.spark.sql.execution.command.TiSetDatabaseCommand.run(databases.scala:28)
  at org.apache.spark.sql.execution.command.ExecutedCommandExec.sideEffectResult$lzycompute(commands.scala:70)
[...]

PR #972, opened 2019-08-01
fix  java.lang.IncompatibleClassChangeError when submit with spark-2.4.3

marsishandsome:

close https://github.com/pingcap/tispark/issues/971

tispark throws java.lang.IncompatibleClassChangeError when submit with spark-2.4.3

the reason is that:
1. tispark is compiled with spark-2.3.3
2. but running with spark-2.4.3
3. some scala code in spark although is compatible between spark-2.3.3 and spark-2.4.3
4. but it's not compatible on the level of jvm byte code

use reflection and call the incompatible code in spark-wrapper-2.3 and spark-wrapper-2.4


core/src/main/scala/org/apache/spark/sql/extensions/rules.scala
```diff
@@ -57,7 +58,7 @@ case class TiResolutionRule(getOrCreateTiContext: SparkSession => TiContext)(
       )(sqlContext)
       // Use SubqueryAlias so that projects and joins can correctly resolve
       // UnresolvedAttributes in JoinConditions, Projects, Filters, etc.
-      SubqueryAlias(tableName, LogicalRelation(tiDBRelation))
+      newSubqueryAlias(tableName, LogicalRelation(tiDBRelation))
```
zhexuany: How about change prefix `new` to `Ti`?

marsishandsome: but function name should start with `lowercase`? any better suggestions? maybe `tiSubqueryAlias`?

birdstorm:
After discussion, we might need to move our tests to a new project to test its stability when depending on different spark version for the same assembly jar.

marsishandsome:
> After discussion, we might need to move our tests to a new project to test its stability when depending on different spark version for the same assembly jar.

see https://github.com/pingcap/tispark/issues/974

core/src/main/scala/org/apache/spark/sql/extensions/rules.scala
zhexuany: Sounds pretty nice

PR #978, opened 2019-08-02
use spark-2.4.3 to run ut

marsishandsome:
tispark-core is compiled with spark-2.3.3, but support running with spark-2.4.3

so we should test tispark-core with spark-2.4.3 in order to ensure compatibility

create a new sub project `tispark-core-test` 
1. `tispark-core-test` depend on spark-2.4.3
2. copy test code from `tispark-core` to `tispark-core-test` before running ci

### DIFF
--- a/core/scripts/fetch-test-data.sh
+++ b/core/scripts/fetch-test-data.sh
@@ -1,3 +1,7 @@
 #!/usr/bin/env bash
-cd ..
+
+CURRENT_DIR=`pwd`
+TISPARK_HOME="$(cd "`dirname "$0"`"/../..; pwd)"
+cd $TISPARK_HOME/core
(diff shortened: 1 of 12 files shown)


core/scripts/fetch-test-data.sh
```diff
@@ -1,3 +1,7 @@
 #!/usr/bin/env bash
-cd ..
-git submodule update --init --recursive
\ No newline at end of file
+
+CURRENT_DIR=`pwd`
+TISPARK_HOME="$(cd "`dirname "$0"`"/../..; pwd)"
```
birdstorm: consider move this `TISPARK_HOME` to an `_env.sh` so that all bash scripts can source the environment settings

Issue #987, opened 2019-08-06
Refactor RegionStoreClient Logic


birdstorm:
Propose a better design for RegionStoreClient.

1. RegionStoreClient should be inaccessible directly.
2. All apis of RegionStoreClient should not provide retry aside from callWithRetry, if a request needs to be retried because of an un-retryable cause, e.g., keys need to be re-split across regions/stores, region info outdated, e.t.c., we should retry it in an upper client logic (KVClient, TxnClient, e.t.c.)

@marsishandsome @zhexuany PTAL

marsishandsome:
nice & LGTM

PR #989, opened 2019-08-06
Refactor RegionStoreClient logic

### LINKED ISSUE #988: LockResolverTest may be incorrect
https://github.com/pingcap/tispark/blob/93128b8ad66e0cc02df4b8574a29523ef18e6afe/tikv-client/src/test/java/com/pingcap/tikv/txn/LockResolverTest.java#L407-L443

Line 422 locks a key by pre-write the key to TiKV without committing. It should result in a KeyIsLocked Exception in Line 428.

birdstorm:
An alternative enhancement against #986.
Provide an example solution for #987. 
Fix #988

**IMPORTANT: SEE TODO IN COMMENTS**

Try to change the RegionStoreClient interface so that it should only deal with retryable errors in one region and one store(as it is named). This PR does part of the job by first making `RegionStoreClient.get()` throw an error whenever an un-retryable exception is encountered, so that it will not need to fetch the region again.

We introduce a new API for all those who want to continue to use RegionStoreClient directly: KVClient. In upper logics such as KVClient and TxnClient, developers should implement the logic of resolving un-retryable exceptions, for instance, re-splitting keys across regions. It will make the error handling more concise for reviewers.

The PR also moves ResolveLock logic to ErrorHandler so that the ErrorHandler should be integrated to handle all kinds of manageable errors.

Changes are made to ScanIterator because it should deal with more circumstances such as scanning a range of keys, and resolve locks when they were present during scan.

Logic fixed in LockResolverTest, the test was incorrectly set according to #988. @zhexuany We may discuss it again later.

### Check List 

Tests 

 - Unit test

Related changes

 - Need to cherry-pick to the release branch (should be decided after discussion)

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
@@ -1,0 +1,227 @@
+/*
+ *
+ * Copyright 2019 PingCAP, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.pingcap.tikv;
+
+import com.google.protobuf.ByteString;
+import com.pingcap.tikv.exception.TiKVException;
+import com.pingcap.tikv.operation.iterator.ConcreteScanIterator;
+import com.pingcap.tikv.region.RegionStoreClient;
+import com.pingcap.tikv.region.RegionStoreClient.*;
+import com.pingcap.tikv.region.TiRegion;
+import com.pingcap.tikv.util.BackOffFunction;
+import com.pingcap.tikv.util.BackOffer;
+import com.pingcap.tikv.util.ConcreteBackOffer;
+import java.util.*;
+import java.util.concurrent.*;
+import java.util.stream.Collectors;
+import org.apache.log4j.Logger;
+import org.tikv.kvproto.Kvrpcpb;
+import org.tikv.kvproto.Kvrpcpb.KvPair;
+
+public class KVClient implements AutoCloseable {
+  private final RegionStoreClientBuilder clientBuilder;
+  private final TiConfiguration conf;
+  private final ExecutorCompletionService<List<KvPair>> completionService;
+  private static final Logger logger = Logger.getLogger(KVClient.class);
+
+  private static final int BATCH_GET_SIZE = 16 * 1024;
+
+  public KVClient(TiConfiguration conf, RegionStoreClientBuilder clientBuilder) {
+    Objects.requireNonNull(conf, "conf is null");
+    Objects.requireNonNull(clientBuilder, "clientBuilder is null");
+    this.conf = conf;
+    this.clientBuilder = clientBuilder;
+    //    ExecutorService executors = Executors.newFixedThreadPool(conf.getKVClientConcurrency());
+    ExecutorService executors = Executors.newFixedThreadPool(20);
+    this.completionService = new ExecutorCompletionService<>(executors);
+  }
+
+  @Override
+  public void close() {}
+
+  /**
+   * Get a raw key-value pair from TiKV if key exists
+   *
+   * @param key raw key
+   * @return a ByteString value if key exists, ByteString.EMPTY if key does not exist
+   */
+  public ByteString get(ByteString key, long version) {
+    BackOffer backOffer = ConcreteBackOffer.newGetBackOff();
+    while (true) {
+      RegionStoreClient client = clientBuilder.build(key);
+      try {
+        return client.get(backOffer, key, version);
+      } catch (final TiKVException e) {
+        backOffer.doBackOff(BackOffFunction.BackOffFuncType.BoRegionMiss, e);
+      }
+    }
+  }
+
+  /**
+   * Get a set of key-value pair by keys from TiKV
+   *
+   * @param keys keys
+   */
+  public List<KvPair> batchGet(List<ByteString> keys, long version) {
+    return batchGet(ConcreteBackOffer.newBatchGetMaxBackOff(), keys, version);
+  }
+
+  private List<KvPair> batchGet(BackOffer backOffer, List<ByteString> keys, long version) {
+    Set<ByteString> set = new HashSet<>(keys);
+    return batchGet(backOffer, set, version);
+  }
+
+  private List<KvPair> batchGet(BackOffer backOffer, Set<ByteString> keys, long version) {
+    Map<TiRegion, List<ByteString>> groupKeys = groupKeysByRegion(keys);
+    List<Batch> batches = new ArrayList<>();
+
+    for (Map.Entry<TiRegion, List<ByteString>> entry : groupKeys.entrySet()) {
+      appendBatches(batches, entry.getKey(), entry.getValue(), BATCH_GET_SIZE);
+    }
+    return sendBatchGet(backOffer, batches, version);
+  }
+
+  /**
+   * Scan raw key-value pairs from TiKV in range [startKey, endKey)
+   *
+   * @param startKey raw start key, inclusive
+   * @param endKey raw end key, exclusive
+   * @return list of key-value pairs in range
+   */
+  public List<Kvrpcpb.KvPair> scan(ByteString startKey, ByteString endKey, long version) {
+    Iterator<Kvrpcpb.KvPair> iterator =
+        scanIterator(conf, clientBuilder, startKey, endKey, version);
+    List<Kvrpcpb.KvPair> result = new ArrayList<>();
+    iterator.forEachRemaining(result::add);
+    return result;
+  }
+
+  /**
+   * Scan raw key-value pairs from TiKV in range [startKey, endKey)
+   *
+   * @param startKey raw start key, inclusive
+   * @param limit limit of key-value pairs
+   * @return list of key-value pairs in range
+   */
+  public List<Kvrpcpb.KvPair> scan(ByteString startKey, int limit) {
+    Iterator<Kvrpcpb.KvPair> iterator = scanIterator(conf, clientBuilder, startKey, limit);
+    List<Kvrpcpb.KvPair> result = new ArrayList<>();
+    iterator.forEachRemaining(result::add);
+    return result;
+  }
+
+  /** A Batch containing the region, a list of keys to send */
+  private static final class Batch {
+    private final TiRegion region;
+    private final List<ByteString> keys;
+
+    Batch(TiRegion region, List<ByteString> keys) {
+      this.region = region;
+      this.keys = keys;
+    }
+  }
+
+  /**
+   * Append batch to list and split them according to batch limit
+   *
+   * @param batches a grouped batch
+   * @param region region
+   * @param keys keys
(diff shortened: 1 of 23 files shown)


tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
```diff
@@ -0,0 +1,227 @@
...
+  private static final Logger logger = Logger.getLogger(KVClient.class);
+
+  private static final int BATCH_GET_SIZE = 16 * 1024;
+
+  public KVClient(TiConfiguration conf, RegionStoreClientBuilder clientBuilder) {
+    Objects.requireNonNull(conf, "conf is null");
+    Objects.requireNonNull(clientBuilder, "clientBuilder is null");
+    this.conf = conf;
+    this.clientBuilder = clientBuilder;
+    //    ExecutorService executors = Executors.newFixedThreadPool(conf.getKVClientConcurrency());
```
zhexuany: Is this dead code?

tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
```diff
@@ -0,0 +1,227 @@
...
+   * @return list of key-value pairs in range
+   */
+  public List<Kvrpcpb.KvPair> scan(ByteString startKey, int limit) {
+    Iterator<Kvrpcpb.KvPair> iterator = scanIterator(conf, clientBuilder, startKey, limit);
+    List<Kvrpcpb.KvPair> result = new ArrayList<>();
+    iterator.forEachRemaining(result::add);
+    return result;
+  }
+
+  /** A Batch containing the region, a list of keys to send */
```
zhexuany: s/A Batch containing the region, a list of keys to send/A Batch containing the region and a list of keys to send to the region

tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
birdstorm: Yeah, i should mark it as todo.

tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+    try {
+      List<KvPair> result = new ArrayList<>();
+      for (int i = 0; i < batches.size(); i++) {
+        result.addAll(
+            completionService.take().get(BackOffer.BATCH_GET_MAX_BACKOFF, TimeUnit.SECONDS));
+      }
+      return result;
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
```
zhexuany: I do not understand why we need interrupt thread here.

tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
birdstorm: ok

tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
birdstorm: do you have any good ideas to deal with InterruptedException?

zhexuany:
```
try {
  Thread.sleep(100);
} catch (InterruptedException ex) {
  throw new RuntimeException(ex);
}
```
Above is what I found on stackoverflow

birdstorm: Well, in case you do not want to do anything to interrupt exception, I think it is better to restore the interrupted status. What do you think, @marsishandsome ?

tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
```diff
@@ -0,0 +1,227 @@
...
+    Objects.requireNonNull(clientBuilder, "clientBuilder is null");
+    this.conf = conf;
+    this.clientBuilder = clientBuilder;
+    //    ExecutorService executors = Executors.newFixedThreadPool(conf.getKVClientConcurrency());
+    ExecutorService executors = Executors.newFixedThreadPool(20);
+    this.completionService = new ExecutorCompletionService<>(executors);
+  }
+
+  @Override
+  public void close() {}
```
marsishandsome: should we close the `ExecutorService`?

tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
```diff
@@ -0,0 +1,227 @@
...
+
+  /**
+   * Get a raw key-value pair from TiKV if key exists
+   *
+   * @param key raw key
+   * @return a ByteString value if key exists, ByteString.EMPTY if key does not exist
+   */
+  public ByteString get(ByteString key, long version) {
```
marsishandsome: better to add `which Exceptions this function throws` to the function signature

tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
birdstorm: yes, nice catch

tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
birdstorm: sure

tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
```diff
@@ -0,0 +1,234 @@
...
+    }
+  }
+
+  /**
+   * Get a raw key-value pair from TiKV if key exists
+   *
+   * @param key raw key
+   * @return a ByteString value if key exists, ByteString.EMPTY if key does not exist
+   */
+  public ByteString get(ByteString key, long version) throws GrpcException {
```
marsishandsome: missed BackOffTimeOut Exception?

tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
```diff
@@ -0,0 +1,234 @@
...
+      }
+    }
+  }
+
+  /**
+   * Get a set of key-value pair by keys from TiKV
+   *
+   * @param keys keys
+   */
+  public List<KvPair> batchGet(List<ByteString> keys, long version) {
```
marsishandsome: as a base function called by others, better to add `throws` what exceptions

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ConcreteScanIterator.java
```diff
@@ -0,0 +1,97 @@
...
+      TiConfiguration conf,
+      RegionStoreClientBuilder builder,
+      ByteString startKey,
+      ByteString endKey,
+      long version) {
+    super(conf, builder, startKey, endKey, Integer.MAX_VALUE);
+    this.version = version;
+  }
+
+  TiRegion loadCurrentRegionToCache() throws Exception {
```
marsishandsome: what `Exception` will be thrown?

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ScanIterator.java
```diff
@@ -18,133 +18,121 @@
...
-    if (isCacheDrained() && !loadCache()) {
+    if (isCacheDrained() && cacheLoadFails()) {
       endOfScan = true;
       return false;
     }
     return true;
   }
 
-  private Kvrpcpb.KvPair getCurrent() {
+  Kvrpcpb.KvPair getCurrent() {
```
marsishandsome: protected?

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ScanIterator.java
```diff
@@ -18,133 +18,121 @@
...
     } catch (Exception e) {
       throw new TiClientInternalException("Error scanning data from region.", e);
     }
-    return true;
+    return false;
   }
 
-  private boolean isCacheDrained() {
-    return currentCache == null || index >= currentCache.size() || index == -1;
+  boolean isCacheDrained() {
```
marsishandsome: protected?

tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
birdstorm: when backoff timeout, it throws an GrpcException

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ScanIterator.java
birdstorm: package private

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ScanIterator.java
birdstorm: same as above

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ScanIterator.java
```diff
@@ -18,133 +18,122 @@
...
+    this.limit = limit;
+    this.conf = conf;
+    this.builder = builder;
   }
 
-  // return false if current cache is not loaded or empty
-  private boolean loadCache() {
-    if (endOfScan) {
-      return false;
+  abstract TiRegion loadCurrentRegionToCache() throws GrpcException;
```
zhexuany: I am confused by the name of the method.

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ScanIterator.java
```diff
@@ -18,133 +18,122 @@
...
       }
       index = 0;
+      Key lastKey = Key.EMPTY;
       // Session should be single-threaded itself
       // so that we don't worry about conf change in the middle
       // of a transaction. Otherwise below code might lose data
-      if (currentCache.size() < session.getConf().getScanBatchSize()) {
-        // Current region done, start new batch from next region
-        startKey = region.getEndKey();
+      if (currentCache.size() < conf.getScanBatchSize()) {
```
zhexuany: Why do we need this if check?

tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
```diff
@@ -234,12 +180,13 @@ private boolean isBatchGetSuccess(BackOffer bo, BatchGetResponse resp) {
     if (!locks.isEmpty()) {
       boolean ok = lockResolverClient.resolveLocks(bo, locks);
       if (!ok) {
-        // if not resolve all locks, we wait and retry
-        bo.doBackOff(BoTxnLockFast, new KeyException((resp.getPairsList().get(0).getError())));
-        return false;
+        // resolveLocks already retried, just throw error to upper logic.
+        throw new TiKVException("locks not resolved, retry");
```
zhexuany: may be "locks cannot be resolved"?

birdstorm: any examples?

tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
```diff
@@ -0,0 +1,235 @@
...
+import com.pingcap.tikv.util.BackOffer;
+import com.pingcap.tikv.util.ConcreteBackOffer;
+import java.util.*;
+import java.util.concurrent.*;
+import java.util.stream.Collectors;
+import org.apache.log4j.Logger;
+import org.tikv.kvproto.Kvrpcpb;
+import org.tikv.kvproto.Kvrpcpb.KvPair;
+
+public class KVClient implements AutoCloseable {
```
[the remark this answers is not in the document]
birdstorm: sure

PR #990, opened 2019-08-07
use tispark self-made m2 cache file

marsishandsome:
use tispark self-made m2 cache file instead of tiflash's cache file


marsishandsome:
> do we have a job to upload cache?

yes, tispark_update_m2_cache_daily on jenkins

PR #991, opened 2019-08-07
add spark sql document for batch write

marsishandsome:
add spark sql document for batch write


docs/datasource_api_userguide.md
```diff
@@ -216,6 +216,54 @@ df.write
   .save()
 ```
 
+## Use Data Source API in SparkSQL
+Config tidb/pd address and enable write through SparkSQL in `conf/spark-defaults.conf` as follows:
+
+```
+spark.tispark.pd.addresses pd0:2379
```
birdstorm: pd0?

marsishandsome: fixed

PR #995, opened 2019-08-07
fix index scan bug

zhexuany:
When we do index scan, we should not add required index data type. 

Remove related code in double read case.

### Check List 

Tests 

Related changes


zhexuany:
@birdstorm @marsishandsome PTAL

marsishandsome:
cherry pick to release-2.1?

PR #999, opened 2019-08-08
refactor batch write region pre-split

marsishandsome:
close https://github.com/pingcap/tispark/issues/891

refine row size estimation when calculate how many regions need to be split.

1. use data's real byte size to calculate `region split number`
2. use `SPLIT TABLE t INDEX BETWEEN` instead of `SPLIT TABLE t INDEX BY` 
3. add config `spark.tispark.tikv.region_split_size_in_mb`
4. fix bug: table and database should be wrapped with `

### Check List 

Tests 

 - Unit test

### DIFF
--- a/core/src/test/scala/com/pingcap/tispark/TiBatchWriteSuite.scala
+++ b/core/src/test/scala/com/pingcap/tispark/TiBatchWriteSuite.scala
@@ -40,17 +40,18 @@ class TiBatchWriteSuite extends BaseTiSparkTest {
     database = tpchDBName
     setCurrentDatabase(database)
     for (table <- tables) {
-      tidbStmt.execute(s"drop table if exists `${batchWriteTablePrefix}_$table`")
-      tidbStmt.execute(s"create table if not exists `${batchWriteTablePrefix}_$table` like $table ")
+      val tableToWrite = s"${batchWriteTablePrefix}_$table"
+      tidbStmt.execute(s"drop table if exists `$tableToWrite`")
+      tidbStmt.execute(s"create table if not exists `$tableToWrite` like $table ")
     }
   }
 
   test("ti batch write") {
     for (table <- tables) {
-      println(table)
+      val tableToWrite = s"${batchWriteTablePrefix}_$table"
 
       // select
-      refreshConnections(TestTables(database, s"${batchWriteTablePrefix}_$table"))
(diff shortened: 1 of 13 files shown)


core/src/test/scala/com/pingcap/tispark/TiBatchWriteSuite.scala
```diff
@@ -33,15 +33,15 @@ class TiBatchWriteSuite extends BaseTiSparkTest {
...
+  private val batchWriteTablePrefix = "BATCH.WRITE"
 
   override def beforeAll(): Unit = {
     super.beforeAll()
     database = tpchDBName
     setCurrentDatabase(database)
     for (table <- tables) {
-      tidbStmt.execute(s"drop table if exists ${batchWriteTablePrefix}_$table")
-      tidbStmt.execute(s"create table if not exists ${batchWriteTablePrefix}_$table like $table ")
+      tidbStmt.execute(s"drop table if exists `${batchWriteTablePrefix}_$table`")
```
birdstorm: how about we make `${batchWriteTablePrefix}_$table` constant?

tikv-client/src/main/java/com/pingcap/tikv/TiDBJDBCClient.java
```diff
@@ -107,14 +107,40 @@ public void splitTableRegion(
...
     } catch (Exception ignored) {
       logger.warn("failed to split table region");
     }
   }
 
+  /**
+   * split index region by calling tidb jdbc command `SPLIT TABLE`, e.g. SPLIT TABLE t INDEX idx
+   * BETWEEN (-9223372036854775808) AND (9223372036854775807) REGIONS 16;
+   *
+   * @param dbName
```
zhexuany: fill the desc info?

core/src/test/scala/com/pingcap/tispark/TiBatchWriteSuite.scala
marsishandsome: nice

tikv-client/src/main/java/com/pingcap/tikv/TiDBJDBCClient.java
marsishandsome: nice

PR #1004, opened 2019-08-08
prohibit agg or groupby pushdown on double read

zhexuany:
This solves an index bug.
We prohibit agg or groupby pushdown on double read case. 

### Check List 

Tests 

Related changes

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
@@ -463,6 +463,7 @@ public class TiDAGRequest implements Serializable {
     }
 
     if (!getGroupByItems().isEmpty() || !getAggregates().isEmpty()) {
+      // only allow table scan or covering index scan push down groupby and agg
       if (!isIndexDoubleScan || (isGroupByCoveredByIndex() && isAggregateCoveredByIndex())) {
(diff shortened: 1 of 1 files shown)


birdstorm:
table scan should also allow aggregate push down.

tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
```diff
@@ -463,58 +464,75 @@ else if (col.getColumnInfo().isPrimaryKey() && tableInfo.isPkHandle()) {
...
-            .forEach(
-                tiExpr ->
-                    aggregationBuilder.addAggFunc(
-                        ProtoConverter.toProto(tiExpr, colOffsetInFieldMap)));
-        executorBuilder.setTp(ExecType.TypeAggregation);
-        dagRequestBuilder.addExecutors(executorBuilder.setAggregation(aggregationBuilder));
-        executorBuilder.clear();
-        addPushDownGroupBys();
-        addPushDownAggregates();
+        pushDownAggAndGroupBy(dagRequestBuilder, executorBuilder, colOffsetInFieldMap);
```
birdstorm: How about we add some comments for better understanding?

zhexuany: indded

PR #1010, opened 2019-08-12
Fix scala compiler version

### LINKED ISSUE #1009: Could not find artifact org.scala-lang:scala-compiler:2.11
**Describe the bug**

`scala-compiler` was not built because it uses `2.11` as version, whereas it actually supports only syntax such as `2.11.x`.

birdstorm:
Fix #1009 

set scala version to `2.11.0` instead of `2.11`.

### Check List 

Tests 

 - No code

Related changes

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
         <spark.version.compile>2.3.3</spark.version.compile>
         <spark.version.test>2.4.3</spark.version.test>
         <scala.binary.version>2.11</scala.binary.version>
-        <scala.version>2.11.0</scala.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -67,7 +67,7 @@
         <spark.version.compile>2.3.3</spark.version.compile>
         <spark.version.test>2.4.3</spark.version.test>
         <scala.binary.version>2.11</scala.binary.version>
-        <scala.version>2.11</scala.version>
+        <scala.version>2.11.0</scala.version>
```
marsishandsome: 2.11.12？

PR #1014, opened 2019-08-13
adding unique indices test for batch write

zhexuany:
Improve the coverage of batch-write on unique indices.

In this PR, we are trying to improve the coverage of batch-write on unique indices. 

In order to test the correctness of batch-write, we have to create different schemas with different numbers of the index on the schema. Additionally, we have to prepare data and use batch-write api to perform the insertion. 

After discussion with @birdstorm, we decided we should generate the schema, so does the data. 

To generate different schemas, we create an array of `dataTypes`. Our schema generator enumerates over the `dataTypes` and chooses some of them to become unique indices. 

Once we have schema, we generate data randomly. 

With schema and data, it is ready to test our batch-write api.

### Check List 

Tests 
 - Integration test

### DIFF
--- a/core/src/test/scala/com/pingcap/tispark/datasource/BaseDataSourceTest.scala
+++ b/core/src/test/scala/com/pingcap/tispark/datasource/BaseDataSourceTest.scala
@@ -43,8 +43,10 @@ class BaseDataSourceTest(val table: String,
     TiSession.clearCache()
   }
 
-  protected def dropTable(tblName: String): Unit =
+  protected def dropTable(tblName: String): Unit = {
     jdbcUpdate(s"drop table if exists $database.$tblName")
+  }
+
   protected def tidbWriteWithTable(rows: List[Row],
                                    schema: StructType,
                                    tblName: String,
@@ -197,16 +199,20 @@ class BaseDataSourceTest(val table: String,
 
   protected def compareTiDBSelectWithJDBCWithTable_V2(tblName: String,
                                                       sortCol: String = "i"): Unit = {
-    val sql = s"select * from $database.$tblName order by $sortCol"
+    val sql = s"select * from `$database`.`$tblName` order by $sortCol"
 
     // check jdbc result & data source result
     val jdbcResult = queryTiDBViaJDBC(sql)
     val df = queryDatasourceTiDBWithTable(sortCol, tableName = tblName)
     val tidbResult = seqRowToList(df.collect(), df.schema)
 
-    assert(
-      compResult(jdbcResult, tidbResult)
-    )
+    if (compResult(jdbcResult, tidbResult)) {
+      assert(true)
+    } else {
+      println(s"failed on $tblName")
+      println(tidbResult)
+      assert(false)
+    }
   }
 
   protected def compareTiDBSelectWithJDBC_V2(sortCol: String = "i"): Unit = {
--- a/core/src/test/scala/com/pingcap/tispark/datasource/BaseDataSourceTest.scala
+++ b/core/src/test/scala/com/pingcap/tispark/datasource/BaseDataSourceTest.scala
@@ -44,7 +44,7 @@ class BaseDataSourceTest(val table: String,
   }
 
   protected def dropTable(tblName: String): Unit = {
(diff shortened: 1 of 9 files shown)


core/src/test/scala/com/pingcap/tispark/datasource/BaseDataSourceTest.scala
```diff
@@ -26,6 +26,10 @@ class BaseDataSourceTest(val table: String,
   override def beforeAll(): Unit = {
     enableTidbConfigPropertiesInjectedToSpark = _enableTidbConfigPropertiesInjectedToSpark
     super.beforeAllWithoutLoadData()
+    if (!database.equals("tispark_test")) {
+      tidbStmt.execute(s"drop database if exists $database")
+      tidbStmt.execute(s"create database $database")
+    }
```
marsishandsome: why recreate the database?

core/src/test/scala/com/pingcap/tispark/datasource/BaseDataSourceTest.scala
```diff
@@ -34,19 +38,27 @@ class BaseDataSourceTest(val table: String,
 
   protected def dropTable(): Unit = jdbcUpdate(s"drop table if exists $dbtable")
 
-  protected def tidbWrite(rows: List[Row],
-                          schema: StructType,
-                          param: Option[Map[String, String]] = None): Unit = {
+  protected def dropTable(tblName: String): Unit =
+    jdbcUpdate(s"drop table if exists $database.$tblName")
```
marsishandsome: `
```suggestion
    jdbcUpdate(s"drop table if exists $dbtable")
```

core/src/test/scala/org/apache/spark/sql/test/generator/Schema.scala
```diff
@@ -17,7 +17,9 @@
 
 package org.apache.spark.sql.test.generator
 
+import com.pingcap.tispark.utils.{TiConverter, TiUtil}
 import org.apache.spark.sql.test.generator.DataType._
+import org.apache.spark.sql.types.{StructField, StructType}
```
marsishandsome: unused import

core/src/test/scala/com/pingcap/tispark/datasource/BaseDataSourceTest.scala
marsishandsome: what if database = tpch_test?

core/src/test/scala/com/pingcap/tispark/datasource/BaseDataSourceTest.scala
zhexuany: Nope. I was using a different table name.

marsishandsome: add ` before and after $database and $tblName

core/src/test/scala/com/pingcap/tispark/datasource/BaseDataSourceTest.scala
zhexuany: It is not recreating database but creates a new database. In our BaseDataSourceSuite, the default database is `tispark_test`

core/src/test/scala/com/pingcap/tispark/datasource/BaseDataSourceTest.scala
zhexuany: ok

core/src/test/scala/org/apache/spark/sql/test/generator/Schema.scala
zhexuany: ok

core/src/test/scala/com/pingcap/tispark/datasource/BaseDataSourceTest.scala
birdstorm: rather strange here...

marsishandsome: what if `ANewTest extends BaseDataSourceTest` and `override` `$database`?

PR #1016, opened 2019-08-13
Delete TiSession.create()

birdstorm:

Delete a confusing method to create TiSession since we have `getInstance()` already.

### Check List 

Tests 

 - Unit test

Code changes


birdstorm:
moved to #1021

PR #1021, opened 2019-08-14
Remove create in tisession

zhexuany:
In this PR, we propose remove `create()` in `TiSession` since we already have `getInstance()`

### DIFF
--- a/core/src/test/scala/com/pingcap/tispark/convert/ToUnsignedSuite.scala
+++ b/core/src/test/scala/com/pingcap/tispark/convert/ToUnsignedSuite.scala
@@ -1,6 +1,8 @@
 package com.pingcap.tispark.convert
 
 import com.pingcap.tikv.TiSession
+import com.pingcap.tikv.allocator.RowIDAllocator
+import com.pingcap.tispark.TiDBUtils
 import com.pingcap.tispark.datasource.BaseDataSourceTest
 import org.apache.spark.sql.Row
 import org.apache.spark.sql.types._
@@ -14,10 +16,6 @@ import org.apache.spark.sql.types._
  * 5. BIGINT UNSIGNED
  */
 class ToUnsignedSuite extends BaseDataSourceTest("test_data_type_convert_to_unsigned") {
-  override def beforeAll() = {
-    super.beforeAll()
-    TiSession.clearCache()
-  }
   private val readSchema = StructType(
     List(
       StructField("i", IntegerType),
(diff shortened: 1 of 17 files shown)


core/src/test/scala/com/pingcap/tispark/convert/ToUnsignedSuite.scala
```diff
@@ -418,10 +422,9 @@ class ToUnsignedSuite extends BaseDataSourceTest("test_data_type_convert_to_unsi
...
 
-  override def afterAll(): Unit =
-    try {
-      dropTable()
-    } finally {
-      super.afterAll()
-    }
+  override def afterAll(): Unit = {
+    super.afterAll()
+    dropTable()
```
marsishandsome: why change the order?

core/src/test/scala/com/pingcap/tispark/datasource/BaseDataSourceTest.scala
```diff
@@ -37,6 +37,7 @@ class BaseDataSourceTest(val table: String,
   protected def tidbWrite(rows: List[Row],
                           schema: StructType,
                           param: Option[Map[String, String]] = None): Unit = {
+    refreshConnections()
```
marsishandsome: better to add `clear catalog cache` after each `drop table`

core/src/test/scala/com/pingcap/tispark/convert/ToUnsignedSuite.scala
zhexuany: It was already undo in my local.

core/src/test/scala/com/pingcap/tispark/datasource/BaseDataSourceTest.scala
zhexuany: DONE

core/src/test/scala/org/apache/spark/sql/catalyst/plans/BasePlanTest.scala
```diff
@@ -159,24 +159,4 @@ class BasePlanTest extends BaseTiSparkTest {
     assert(estimatedRowCount === answer)
   }
 
-  test("baseline test") {
```
birdstorm: Why do we move this test?

zhexuany: I think `test` should belong to a `suite` rather than `BasexxxTest`

PR #1031, opened 2019-08-16
set tikv region size from 96M to 1M

marsishandsome:
Some bugs will only occur when data in TiKV meets certain characteristics.

This PR sets TiKV region size from `96M` to `1M`, so that bugs are easier to appear.

### DIFF
--- a/config/tikv.toml
+++ b/config/tikv.toml
@@ -148,6 +148,7 @@ address = "pushgateway:9091"
 # bit smaller.
 # region-max-size = "144MB"
 # region-split-size = "96MB"
+region-split-size = "1MB"
 
 [rocksdb]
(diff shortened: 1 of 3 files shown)


config/tikv.toml
```diff
@@ -148,6 +148,7 @@ address = "pushgateway:9091"
 # bit smaller.
 # region-max-size = "144MB"
 # region-split-size = "96MB"
+region-split-size = "1MB"
```
zhexuany: why do we do the change?

marsishandsome: as discussed yesterday

This PR sets TiKV region size from 96M to 1M, so that bugs are easier to appear.

PR #1037, opened 2019-08-16
fix reflection bug: pass in different arguments for different version of same function

marsishandsome:
close https://github.com/pingcap/tispark/issues/1006

this PR https://github.com/pingcap/tispark/pull/1017 introduces a new bug:

```
Caused by: java.lang.IllegalArgumentException: wrong number of arguments
  at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
  at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  at java.lang.reflect.Method.invoke(Method.java:498)
  at com.pingcap.tispark.utils.ReflectionUtil$ReflectionMapPartitionWithIndexInternal.invoke(ReflectionUtil.scala:117)
  at org.apache.spark.sql.execution.CoprocessorRDD$$anonfun$doExecute$1.apply(CoprocessorRDD.scala:82)
  at org.apache.spark.sql.execution.CoprocessorRDD$$anonfun$doExecute$1.apply(CoprocessorRDD.scala:78)
  at scala.collection.immutable.List.map(List.scala:273)
```

### How to fix
pass in different arguments for a different version of the same function

### Cherry Pick
to master branch

### DIFF
--- a/core/src/main/scala/com/pingcap/tispark/utils/ReflectionUtil.scala
+++ b/core/src/main/scala/com/pingcap/tispark/utils/ReflectionUtil.scala
@@ -15,8 +15,6 @@
 
 package com.pingcap.tispark.utils
 
-import java.lang.reflect.Method
-
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.UnsafeRow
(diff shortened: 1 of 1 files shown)


birdstorm:
a bit messy... perhaps we can ignore spark version and try these methods naively?

core/src/main/scala/com/pingcap/tispark/utils/ReflectionUtil.scala
```diff
@@ -43,85 +41,89 @@ object ReflectionUtil {
...
+  ) {
+    // Spark HDP Release may not compatible with official Release
+    // see https://github.com/pingcap/tispark/issues/1006
+    def invoke(): RDD[InternalRow] =
+      spark_version match {
+        case "2.3.0" | "2.3.1" =>
+          try {
+            mapPartitionsWithIndexInternalV1(rdd, internalRowToUnsafeRowWithIndex)
+          } catch {
+            case _: Throwable =>
```
zhexuany: We catch all exception here. What if the exception thrown by spark is not what we want?

marsishandsome: nice! fixed

marsishandsome:
> a bit messy... perhaps we can ignore spark version and try these methods naively?

Right, I can simply try these methods naively, like

```
try {
  use spark-2.3 function
} catch {
  use spark-2.4 function
}
```
When using spark-2.3, we will always succeed in running `try`.
But when using spark-2.4, we will always fail in running `try`, which we can avoid.

PR #1044, opened 2019-08-19
adding batch write pk insertion test

zhexuany:
This PR adds new test cases. 
In this PR, we generate different schemas with pk. Later, we generate data based on the schema. 
With data and schema, we can perform insert and select with our code. If the data we insert into tikv agreeing with generated data, then we consider our code is working as expected.

### DIFF
--- a/core/src/test/scala/com/pingcap/tispark/datasource/BaseDataSourceTest.scala
+++ b/core/src/test/scala/com/pingcap/tispark/datasource/BaseDataSourceTest.scala
@@ -202,6 +202,7 @@ class BaseDataSourceTest(val table: String,
     val df = queryDatasourceTiDBWithTable(sortCol, tableName = tblName)
     val tidbResult = seqRowToList(df.collect(), df.schema)
 
+    println(s"running test on table $tblName")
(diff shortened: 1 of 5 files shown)


core/src/test/scala/com/pingcap/tispark/datasource/BaseDataSourceTest.scala
```diff
@@ -202,6 +202,7 @@ class BaseDataSourceTest(val table: String,
     val df = queryDatasourceTiDBWithTable(sortCol, tableName = tblName)
     val tidbResult = seqRowToList(df.collect(), df.schema)
 
+    println(s"running test on table $tblName")
```
marsishandsome: use log?

zhexuany: We use `println` only in our test codebase. It should be ok. Printing this out can help us know whether the test is run or not. In addition, with this info, it helps us identify test failing on which table.

marsishandsome: image

better to use log4j

Issue #1039, opened 2019-08-16
[BUG] Distinct without alias throws NullPointerException


birdstorm:
**Describe the bug**

Distinct without alias fails.

**What did you do**

```
create table t(c1 int);
insert into t values (2), (3);
select distinct(c1), 1 as w from t;
select distinct(c1) as d, 1 as w from t;
```

**What do you expect**

Both select should be ok.

**What happens instead**

First select fails while the second one succeeds.
```
ERROR Executor: Exception in task 0.0 in stage 1.0 (TID 1)
java.lang.NullPointerException: at index 2
	at com.google.common.collect.ObjectArrays.checkElementNotNull(ObjectArrays.java:225)
	at com.google.common.collect.ObjectArrays.checkElementsNotNull(ObjectArrays.java:215)
	at com.google.common.collect.ObjectArrays.checkElementsNotNull(ObjectArrays.java:209)
	at com.google.common.collect.ImmutableList.construct(ImmutableList.java:327)
	at com.google.common.collect.ImmutableList.copyOf(ImmutableList.java:244)
	at com.pingcap.tikv.operation.transformer.RowTransformer.<init>(RowTransformer.java:86)
	at com.pingcap.tikv.operation.transformer.RowTransformer.<init>(RowTransformer.java:41)
	at com.pingcap.tikv.operation.transformer.RowTransformer$Builder.build(RowTransformer.java:52)
	at com.pingcap.tikv.operation.SchemaInfer.buildTransform(SchemaInfer.java:83)
	at com.pingcap.tikv.operation.SchemaInfer.<init>(SchemaInfer.java:53)
	at com.pingcap.tikv.operation.SchemaInfer.create(SchemaInfer.java:46)
	at com.pingcap.tikv.operation.SchemaInfer.create(SchemaInfer.java:42)
	at com.pingcap.tikv.operation.iterator.CoprocessIterator.getRowIterator(CoprocessIterator.java:78)
	at com.pingcap.tikv.Snapshot.tableRead(Snapshot.java:97)
	at org.apache.spark.sql.tispark.TiRowRDD$$anon$1.<init>(TiRowRDD.scala:69)
	at org.apache.spark.sql.tispark.TiRowRDD.compute(TiRowRDD.scala:59)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:324)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:288)
[...]

PR #1054, opened 2019-08-21
fix distinct without alias bug: disable pushdown aggregate with alias

marsishandsome:
close https://github.com/pingcap/tispark/issues/1039

Distinct without alias fails, e.g.
```
select distinct(c1), 1 as w from t;
```

disable pushdown aggregate with alias

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
+++ b/core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
@@ -468,7 +468,8 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
       filters.forall(TiUtil.isSupportedFilter(_, source, blacklist)) &&
       groupingExpressions.forall(TiUtil.isSupportedGroupingExpr(_, source, blacklist)) &&
       aggregateExpressions.forall(TiUtil.isSupportedAggregate(_, source, blacklist)) &&
-      !aggregateExpressions.exists(_.isDistinct)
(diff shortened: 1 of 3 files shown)


birdstorm:
This is a temporary fix, because we have not found a best way to fix it.

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -468,7 +468,8 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
       filters.forall(TiUtil.isSupportedFilter(_, source, blacklist)) &&
       groupingExpressions.forall(TiUtil.isSupportedGroupingExpr(_, source, blacklist)) &&
       aggregateExpressions.forall(TiUtil.isSupportedAggregate(_, source, blacklist)) &&
-      !aggregateExpressions.exists(_.isDistinct)
+      !aggregateExpressions.exists(_.isDistinct) &&
```
zhexuany: are planning a better fix? If so, we may add a todo here.

marsishandsome: fixed

marsishandsome: > are planning a better fix? If so, we may add a todo here.

fixed

zhexuany:
LGTM

Issue #1056, opened 2019-08-22
[BUG] ci failed: test show databases/tables


marsishandsome:
https://internal.pingcap.net/idc-jenkins/blue/rest/organizations/jenkins/pipelines/tispark_regression_test_daily/runs/50/nodes/731/steps/974/log/?start=0

```
[2019-08-21T19:09:08.172Z] - test show databases/tables *** FAILED ***
[2019-08-21T19:09:08.172Z]   java.lang.UnsupportedOperationException:
[2019-08-21T19:09:08.172Z]   at com.google.common.collect.UnmodifiableIterator.remove(UnmodifiableIterator.java:46)
[2019-08-21T19:09:08.172Z]   at java.util.Collection.removeIf(Collection.java:415)
[2019-08-21T19:09:08.172Z]   at com.pingcap.tikv.catalog.Catalog$CatalogCache.listTables(Catalog.java:89)
[2019-08-21T19:09:08.172Z]   at com.pingcap.tikv.catalog.Catalog.listTables(Catalog.java:185)
[2019-08-21T19:09:08.172Z]   at com.pingcap.tispark.MetaManager.getTables(MetaManager.scala:36)
[2019-08-21T19:09:08.172Z]   at org.apache.spark.sql.catalyst.catalog.TiDirectExternalCatalog.listTables(TiDirectExternalCatalog.scala:56)
[2019-08-21T19:09:08.172Z]   at org.apache.spark.sql.catalyst.catalog.TiDirectExternalCatalog.listTables(TiDirectExternalCatalog.scala:60)
[2019-08-21T19:09:08.172Z]   at org.apache.spark.sql.catalyst.catalog.SessionCatalog.listTables(SessionCatalog.scala:765)
[2019-08-21T19:09:08.172Z]   at org.apache.spark.sql.catalyst.catalog.TiCompositeSessionCatalog.listTables(TiCompositeSessionCatalog.scala:219)
[2019-08-21T19:09:08.172Z]   at org.apache.spark.sql.catalyst.catalog.TiCompositeSessionCatalog.listTables(TiCompositeSessionCatalog.scala:204)
```

zhexuany:
It seems we have concurrent access issue here.

zhexuany:
After some investigation, it is not a concurrent issue. The unsupported operation exception was thrown because of the implementation of remove of `UnmodifiableIterator`

PR #1057, opened 2019-08-22
using stream rather removeIf

zhexuany:

This PR closes https://github.com/pingcap/tispark/issues/1056

We use stream rather than removeIf.

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/catalog/Catalog.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/catalog/Catalog.java
@@ -86,7 +86,7 @@ public class Catalog implements AutoCloseable {
         tableMap = loadTables(db);
       }
       Collection<TiTableInfo> tables = tableMap.values();
-      return tables.stream().filter(TiTableInfo::isView).collect(Collectors.toList());
+      return tables.stream().filter(TiTableInfo::isNotView).collect(Collectors.toList());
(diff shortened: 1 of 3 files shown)


tikv-client/src/main/java/com/pingcap/tikv/meta/TiTableInfo.java
```diff
@@ -94,6 +94,10 @@ public TiTableInfo(
     primaryKeyColumn = primaryKey;
   }
 
+  public boolean isNotView() {
+    return this.viewInfo == null;
```
birdstorm: better use `!isView()` so that we don't need to modify multiple lines in future.

tikv-client/src/main/java/com/pingcap/tikv/catalog/Catalog.java
```diff
@@ -86,8 +86,7 @@ public TiDBInfo getDatabase(String name) {
         tableMap = loadTables(db);
       }
       Collection<TiTableInfo> tables = tableMap.values();
-      tables.removeIf(TiTableInfo::isView);
-      return ImmutableList.copyOf(tables);
+      return tables.stream().filter(TiTableInfo::isNotView).collect(Collectors.toList());
```
birdstorm: or you can use `filter(x => !x.isView)`

PR #1064, opened 2019-08-25
adding recreate flag when create tisession

zhexuany:
After we remove `TiSession.create()` from codebase, our ci randomly reports no-such-table exception. In this PR, we add it back in a different way.

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
@@ -61,16 +61,7 @@ public class TiSession implements AutoCloseable {
   // Since we create session as singleton now, configuration change will not
   // reflect change
   public static TiSession getInstance(TiConfiguration conf) {
-    synchronized (sessionCachedMap) {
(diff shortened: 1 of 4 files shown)


tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
```diff
@@ -45,12 +45,10 @@
 
   private static final Map<String, TiSession> sessionCachedMap = new HashMap<>();
 
-  // Since we create session as singleton now, configuration change will not
-  // reflect change
-  public static TiSession getInstance(TiConfiguration conf) {
+  public static TiSession getInstance(TiConfiguration conf, boolean recreate) {
```
birdstorm: When recreate is true, it seems that sessionCacheMap is useless. Mind if we change the method's signature?

zhexuany: It will update the session with the same key in sessionMap.

birdstorm: ic

Issue #1076, opened 2019-08-29
[BUG] Meet "object Predef does not have a member classOf" error when use tispark in Livy


yantzu:
**Describe the bug**
When use tispark in livy thrift server, I add tispark jar with livy.session.conf.spark.jars config properties. The spark application will start, but will be dead soon.

**What did you do**
Just submit spark application using:
```
beeline -n hive -p -u "jdbc:hive2://localhost:10010" --hiveconf livy.session.conf.spark.sql.extensions=org.apache.spark.sql.TiExtensions --hiveconf livy.session.conf.spark.tispark.pd.addresses=localhost:2379 --hiveconf livy.session.conf.spark.jars=hdfs:///tispark-core-2.1.4-spark_2.4-jar-with-dependencies.jar
```
**What do you expect**
Spark application should start normally, and I can access tidb in spark

**What happens instead**
Spark application stopped with error:
```
scala.reflect.internal.FatalError: object Predef does not have a member classOf
	at scala.reflect.internal.Definitions$DefinitionsClass.scala$reflect$internal$Definitions$DefinitionsClass$$fatalMissingSymbol(Definitions.scala:1182)
	at scala.reflect.internal.Definitions$DefinitionsClass.getMember(Definitions.scala:1199)
	at scala.reflect.internal.Definitions$DefinitionsClass.getMemberMethod(Definitions.scala:1234)
	at scala.reflect.internal.Definitions$DefinitionsClass$RunDefinitions.Predef_classOf$lzycompute(Definitions.scala:1465)
	at scala.reflect.internal.Definitions$DefinitionsClass$RunDefinitions.Predef_classOf(Definitions.scala:1465)
	at scala.reflect.internal.Definitions$DefinitionsClass$RunDefinitions.isPredefClassOf(Definitions.scala:1455)
	at scala.tools.nsc.typechecker.Typers$Typer.typedIdent$2(Typers.scala:4912)
	at scala.tools.nsc.typechecker.Typers$Typer.typedIdentOrWildcard$1(Typers.scala:4935)
	at scala.tools.nsc.typechecker.Typers$Typer.typedInAnyMode$1(Typers.scala:5367)
	at scala.tools.nsc.typechecker.Typers$Typer.typed1(Typers.scala:5387)
	at scala.tools.nsc.typechecker.Typers$Typer.runTyper$1(Typers.scala:5423)
[...]

PR #1075, opened 2019-08-28
Remove useless dependencies

birdstorm:
### What problem does this PR solve? 
Close #1073 and close #1074 
Some dependencies are useless in TiSpark assembly jar.

### What is changed and how it works?
I find useless `scala-lang` dependency in core, and useless `jackson` dependency in tikv-client.
Both dependencies can be fetched from spark. 

Assembly Jar size reduce from 29.4 MB to 22 MB (-25.2%).

### Check List 

Tests 

 - No code

Code changes

 - Has dependency change

PR #1077, opened 2019-08-29
Remove scala-lang from assembly jar to fix bug when use tispark in Livy

yantzu:
https://github.com/pingcap/tispark/issues/1076

The bug reason is tispark include scala-lang class which is not compatible with livy. While it is not necessary to include scala-lang class in assembly jar, so I remove them all.

Similar issue: https://social.msdn.microsoft.com/Forums/azure/en-US/fe7bd433-bc38-4513-9bb6-a87641ccdcfe/import-external-jars-when-using-jupyter-notebook-with-spark-kernel?forum=hdinsight

### Check List 

Tests 
Have tested manually

Code changes
build script changes

Side effects
May need to include scala-lang dependency if not using spark

### DIFF
--- a/tikv-client/pom.xml
+++ b/tikv-client/pom.xml
@@ -359,6 +359,11 @@
                                     <shadedPattern>shade.com.google.common</shadedPattern>
                                 </relocation>
                             </relocations>
+                            <artifactSet>
+                                <excludes>
+                                    <exclude>org.scala-lang:*</exclude>
+                                </excludes>
+                            </artifactSet>
                         </configuration>
                     </execution>
                 </executions>


yantzu:
Realize PR #1075 has fix the issue, so I just close this PR. We work on it almost on the same time.

marsishandsome:
hi  @yantzu  tispark-2.1.5 is released, please download here https://github.com/pingcap/tispark/releases/tag/v2.1.5

yantzu:
Thanks a lot, I will have a try. Impressed by your efficiency of the work.

PR #1085, opened 2019-08-30
Support read from TiFlash

zhexuany:
Add flag to support read from TiFlash.

### DIFF
--- a/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
+++ b/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
@@ -29,6 +29,7 @@ import org.apache.spark.sql.catalyst.expressions.aggregate._
 import org.apache.spark.sql.catalyst.expressions.{AttributeReference, Expression, Literal, NamedExpression}
 import org.apache.spark.sql.execution.SparkPlan
 import org.apache.spark.sql.execution.aggregate.SortAggregateExec
+import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.types.{MetadataBuilder, StructField, StructType}
 import org.tikv.kvproto.Kvrpcpb.{CommandPri, IsolationLevel}
 
@@ -95,6 +96,9 @@ object TiUtil {
 
     true
   }
+
+  def allowTiFlashRead(sqlConf: SQLConf) =
+    sqlConf.getConfString(TiConfigConst.USE_TIFLASH, "false").toLowerCase.toBoolean
 
   def isSupportedFilter(expr: Expression,
                         source: TiDBRelation,
--- a/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
+++ b/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
@@ -192,6 +192,10 @@ object TiUtil {
       tiConf.setTikvRegionSplitSizeInMB(conf.get(TiConfigConst.TIKV_REGION_SPLIT_SIZE_IN_MB).toInt)
     }
 
+    if (conf.contains(TiConfigConst.ENABLE_TIFLASH_TEST)) {
+      tiConf.setUseTiFlash(conf.get(TiConfigConst.ENABLE_TIFLASH_TEST).toBoolean)
+    }
+
     tiConf
   }
 
--- a/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
+++ b/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
@@ -196,10 +196,6 @@ object TiUtil {
       tiConf.setTikvRegionSplitSizeInMB(conf.get(TiConfigConst.TIKV_REGION_SPLIT_SIZE_IN_MB).toInt)
     }
 
-    if (conf.contains(TiConfigConst.ENABLE_TIFLASH_TEST)) {
-      tiConf.setUseTiFlash(conf.get(TiConfigConst.ENABLE_TIFLASH_TEST).toBoolean)
-    }
-
     tiConf
   }
 
--- a/core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
+++ b/core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
@@ -80,6 +80,10 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
   private def allowIndexRead(): Boolean =
     sqlConf.getConfString(TiConfigConst.ALLOW_INDEX_READ, "true").toLowerCase.toBoolean
 
+  private def allowTiFlashRead(): Boolean = {
+    sqlConf.getConfString(TiConfigConst.USE_TIFLASH, "false").toLowerCase.toBoolean
+  }
+
   private def useStreamingProcess(): Boolean =
     sqlConf.getConfString(TiConfigConst.COPROCESS_STREAMING, "false").toLowerCase.toBoolean
 
@@ -254,6 +258,7 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
 
     scanBuilder.buildTiDAGReq(
       allowIndexRead(),
+      allowTiFlashRead(),
       tiColumns.map { _.getColumnInfo }.asJava,
       tiFilters.asJava,
       source.table,
--- a/core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
+++ b/core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
@@ -80,9 +80,6 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
   private def allowIndexRead(): Boolean =
     sqlConf.getConfString(TiConfigConst.ALLOW_INDEX_READ, "true").toLowerCase.toBoolean
 
-  private def allowTiFlashRead(): Boolean =
-    sqlConf.getConfString(TiConfigConst.USE_TIFLASH, "false").toLowerCase.toBoolean
-
   private def useStreamingProcess(): Boolean =
     sqlConf.getConfString(TiConfigConst.COPROCESS_STREAMING, "false").toLowerCase.toBoolean
 
@@ -257,7 +254,7 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
 
     scanBuilder.buildTiDAGReq(
       allowIndexRead(),
-      allowTiFlashRead(),
+      TiUtil.allowTiFlashRead(sqlConf),
       tiColumns.map { _.getColumnInfo }.asJava,
       tiFilters.asJava,
       source.table,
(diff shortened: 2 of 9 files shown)


zhexuany:
@zanmato1984 PTAL

zhexuany:
@birdstorm @marsishandsome should we merge this PR into master? From my point of view, it is harmless as long as we disable tiflash test in `tidb_config.properties`

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -80,6 +80,10 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
   private def allowIndexRead(): Boolean =
     sqlConf.getConfString(TiConfigConst.ALLOW_INDEX_READ, "true").toLowerCase.toBoolean
 
+  private def allowTiFlashRead(): Boolean = {
+    sqlConf.getConfString(TiConfigConst.USE_TIFLASH, "true").toLowerCase.toBoolean
```
birdstorm: `Why` is it default as true?
What is its difference with `TiRowRDD.scala:43`?

zhexuany: It should be false.

core/src/test/scala/org/apache/spark/sql/test/SharedSQLContext.scala
```diff
@@ -386,6 +386,8 @@ object SharedSQLContext extends Logging {
       tpchDBName = getOrElse(prop, TPCH_DB_NAME, "tpch_test")
       tpcdsDBName = getOrElse(prop, TPCDS_DB_NAME, "")
 
+      val enableTiFlashTest = getOrElse(prop, ENABLE_TIFLASH_TEST, "false")
```
birdstorm: can we just make it a global property for all tests?

core/src/main/scala/org/apache/spark/sql/tispark/TiRowRDD.scala
```diff
@@ -41,6 +38,59 @@ class TiRowRDD(override val dagRequest: TiDAGRequest,
                @transient private val sparkSession: SparkSession)
     extends TiRDD(dagRequest, physicalId, tiConf, tableRef, session, sparkSession) {
 
+  protected val logger: Logger = log
+
+  def allowTiFlashRead(): Boolean =
```
birdstorm: check duplicate here.

core/src/test/scala/org/apache/spark/sql/BaseTiSparkTest.scala
```diff
@@ -377,6 +384,18 @@ class BaseTiSparkTest extends QueryTest with SharedSQLContext {
       }
     }
 
+    // test tiflash's result
+    if (spark.conf.get(TiConfigConst.ENABLE_TIFLASH_TEST).toBoolean) {
+      // get result from TiFlash
+      try {
+        spark.conf.set(TiConfigConst.USE_TIFLASH, "true")
```
birdstorm: why do we need to set this configuration to true?

core/src/test/scala/org/apache/spark/sql/BaseTiSparkTest.scala
```diff
@@ -377,6 +384,18 @@ class BaseTiSparkTest extends QueryTest with SharedSQLContext {
       }
     }
 
+    // test tiflash's result
+    if (spark.conf.get(TiConfigConst.ENABLE_TIFLASH_TEST).toBoolean) {
```
birdstorm: make this a global variable for tests.

core/src/test/scala/org/apache/spark/sql/BaseTiSparkTest.scala
```diff
@@ -377,6 +384,18 @@ class BaseTiSparkTest extends QueryTest with SharedSQLContext {
       }
     }
 
+    // test tiflash's result
```
birdstorm: when `TiFlash` is tested, you should disable all other tests, because we only want to test if `TiFlash` is correct, regardless of TiSpark.

zhexuany: TiFlash test is just an if branch, it does not affect other tests.

core/src/test/scala/org/apache/spark/sql/BaseTiSparkTest.scala
zhexuany: Nope. TiFlash test currently is not supported by CI we have to maintain this config.

core/src/test/scala/org/apache/spark/sql/BaseTiSparkTest.scala
zhexuany: This flag is used to control code behavior. 

The ENABLE_TIFLASH_TEST is to control test behavior.

core/src/main/scala/org/apache/spark/sql/tispark/TiRowRDD.scala
zhexuany: Could you be more specific?

birdstorm: You have two `allowTiFlashRead()` existing in code. I'm not sure if it is necessary.

core/src/test/scala/org/apache/spark/sql/BaseTiSparkTest.scala
birdstorm: well you don't have to read configuration from spark.conf *every time* when `runTest()` is called

core/src/test/scala/org/apache/spark/sql/BaseTiSparkTest.scala
birdstorm: I mean that when you try to test TiFlash, you should disable tests concerning TiSpark. Or else the result may not be correct because we only need one of the three tests to succeed in order to mark the test as passed, while in fact we need TiFlash test to pass strictly.

core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
```diff
@@ -192,6 +192,10 @@ object TiUtil {
       tiConf.setTikvRegionSplitSizeInMB(conf.get(TiConfigConst.TIKV_REGION_SPLIT_SIZE_IN_MB).toInt)
     }
```
zanmato1984: No setting "USE_TIFLASH"? We may want to use tiflash for various usage in our local environment, not only for testing.

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -80,6 +80,10 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
   private def allowIndexRead(): Boolean =
     sqlConf.getConfString(TiConfigConst.ALLOW_INDEX_READ, "true").toLowerCase.toBoolean
 
+  private def allowTiFlashRead(): Boolean = {
```
zanmato1984: `{}` is not neccesary, just make it of similar format in this `allow*` member category.

tikv-client/src/main/java/com/pingcap/tikv/TiConfiguration.java
```diff
@@ -72,6 +75,11 @@
   private boolean writeWithoutLockTable = DEF_WRITE_WITHOUT_LOCK_TABLE;
   private int tikvRegionSplitSizeInMB = DEF_TIKV_REGION_SPLIT_SIZE_IN_MB;
 
+  private boolean useTiFlash = DEF_USE_TIFLASH;
+  private String tiFlashLabelKey = DEF_TIFLASH_LABEL_KEY;
+
```
zanmato1984: Blank line?

core/src/test/scala/org/apache/spark/sql/BaseTiSparkTest.scala
birdstorm: understood that a fail() is called when comparison fails.

core/src/test/scala/org/apache/spark/sql/BaseTiSparkTest.scala
```diff
@@ -377,6 +384,26 @@ class BaseTiSparkTest extends QueryTest with SharedSQLContext {
...
+          fail(
+            s"""Failed with
+               |TiSpark:\t\t${listToString(r1)}
+               |Spark With TiFlash:${listToString(r2)}
+               |TiDB:\t\t\t${listToString(r3)}""".stripMargin
+          )
+        }
+      } catch {
+        case e: Throwable =>
+          logger.warn(s"TiSpark over TiFlash failed when executing: $qJDBC", e) // JDBC failed
```
birdstorm: Do we need to add a `fail()` here?

core/src/test/scala/org/apache/spark/sql/BaseTiSparkTest.scala
birdstorm: Is there any case that code behavior should be different from test behavior? Could you please describe them in comments for better understanding?

zhexuany:
@birdstorm @zanmato1984 PTAL

birdstorm:
Nice! I'm still wondering why there are two sets of configuration about TiFlash tests. Does it mean that enabling TiFlash code is not equivalent to enabling TiFlash tests?

core/src/main/scala/org/apache/spark/sql/tispark/TiRowRDD.scala
```diff
@@ -41,6 +38,53 @@ class TiRowRDD(override val dagRequest: TiDAGRequest,
...
 
+  protected val logger: Logger = log
+
+  override protected def getPartitions: Array[Partition] = {
+    val partitions = super.getPartitions
+
+    if (!dagRequest.getUseTiFlash)
+      return partitions
+
+    logger.info("testing tiflash logic")
```
zanmato1984: You should not assume this code path is for "testing" ONLY.

core/src/test/scala/org/apache/spark/sql/test/SharedSQLContext.scala
birdstorm: `enableTiFlashTest` can be used as a flag in test only. It does not need to be contained in `spark.conf`

zhexuany: done

core/src/main/scala/org/apache/spark/sql/tispark/TiRowRDD.scala
zhexuany: done

birdstorm:
LGTM

core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
```diff
@@ -192,6 +196,10 @@ object TiUtil {
...
 
+    if (conf.contains(TiConfigConst.ENABLE_TIFLASH_TEST)) {
```
zanmato1984: Wrong.

PR #1087, opened 2019-09-02
update document for release 2.2.0

marsishandsome:
update document for release 2.2.0

### DIFF
--- a/docs/datasource_api_userguide.md
+++ b/docs/datasource_api_userguide.md
@@ -2,7 +2,7 @@
 
 The TiDB Connector for Spark enables you to use TiDB as the data source of Apache Spark, similar to other data sources (PostgreSQL, HDFS, S3, etc.).
 
-The TiDB Connector supports Spark 2.3.0+.
+The TiDB Connector supports Spark 2.3.0+ and Spark 2.4.0+.
 
 ## Interaction Between TiDB and Spark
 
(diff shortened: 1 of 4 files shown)


docs/datasource_api_userguide.md
```diff
@@ -308,7 +300,9 @@ The following table shows the TiDB-specific options, which can be passed in thro
 
 ## TiDB Version and Configuration for Write
 
-TiDB's version must be 3.0 or later.
+TiDB's version must be 4.0 or later.
+
+**IMPORTANT: currently TiDB-4.0 is not released yet, but you can use code on tidb's master branch.**
```
zhexuany: s/can use code on tidb's master branch./can use tidb's master branch. 

Additionally, any requirement of TiKV and pd?

docs/datasource_api_userguide.md
```diff
@@ -324,7 +318,7 @@ delay-clean-table-lock: 60000
 split-table: true
 ```
 
-If your TiDB's version is earlier than 3.0, set `spark.tispark.write.without_lock_table` to `true` to enable write, but ACID is **not** guaranteed.
+If your TiDB's version is earlier than 4.0, set `spark.tispark.write.without_lock_table` to `true` to enable write, but ACID is **not** guaranteed.
```
zhexuany: Do we need to fill more info here? @birdstorm

docs/datasource_api_userguide.md
marsishandsome: fixed.

requirement of TiKV and pd is that: compatible with tidb master branch

docs/datasource_api_userguide.md
birdstorm: We may add the detailed info when we throw exception in batch write?

birdstorm: We may add the detailed info when we throw exception in batch write?

PR #1095, opened 2019-09-05
rename tispark core maven name

marsishandsome:
When user upgrades from tispark-2.1.x to tispark-2.2.x, they may still use `tispark-core`, but they should use `tispark-assembly`.

In this PR, we renamed tispark core maven name from `tispark-core` to `tispark-core-internal`.

close https://github.com/pingcap/tispark/issues/1096


birdstorm:
do we really need to deploy tispark-core now?

PR #1101, opened 2019-09-12
add field type on each expr

zhexuany:
TiFlash's side is doing a stricter check. TiFlash requires expr, pushed down to it, providing FieldType. This PR addresses the problem.
Adding FieldType for each expr. 

Tests 


guliangliangatpingcap:
@zhexuany should release in tispark-2.2.1 ?

zhexuany:
@guliangliangatpingcap Yep. It should be.

Issue #1102, opened 2019-09-16
[BUG] Data in TiDB does not update when selecting from Structured Streaming


guliangliangatpingcap:
see https://asktug.com/t/structured-streaming-sql/1080

When selecting data from tidb using spark structured streaming, the data selecting from tidb will never be updated.

The reason is that the `CoprocessorRDD's timestamp` will not be updated.

code is as follows:
```
import java.sql.*;

/**
 * test spark!
 */
public class App {


    public static void main(String[] args) {
        final Connection conn = getConnect();

        SparkConf conf = new SparkConf().setAppName("StructuredStream")
                .setMaster("spark://192.168.10.201:7077")
                .set("spark.driver.cores","1")
                .set("spark.driver.memory","1G")
                .set("spark.deploy.defaultCores","3")
                .set("spark.cores.max","3")
                .set("spark.executor.cores","1")
                .set("spark.executor.memory","2G");


        SparkSession spark = SparkSession
                .builder()
                .config(conf)
                .getOrCreate();
        spark.sparkContext().setLogLevel("WARN");
        SQLContext sqlContext = spark.sqlContext();

        Dataset<Row> df = spark.readStream()
                .format("kafka")
                .option("kafka.bootstrap.servers", "192.168.10.201:9092,192.168.10.202:9092,192.168.10.203:9092,192.168.10.204:9092")
                .option("subscribe","kafka_obinlog")
                .option("startingOffsets","latest")
                .load();
        Dataset<Row> data = df.select(functions.col("value"),functions.col("timestamp"));

        Encoder<BinglogTest> rowEncoder = Encoders.bean(BinglogTest.class);
        Dataset<Row> binlog = data.map(line->{
            BinlogOuterClass.Binlog binlogText = BinlogOuterClass.Binlog.parseFrom((byte[])line.get(0));
            BinlogOuterClass.Table table = binlogText.getDmlData().getTables(0);
            BinlogOuterClass.Row row = table.getMutations(0).getRow();
            return new BinglogTest((int)row.getColumns(0).getInt64Value(),row.getColumns(1).getStringValue());
        },rowEncoder).toDF();

        binlog.createOrReplaceGlobalTempView("binlog_test");
        Dataset<Row> d = sqlContext.sql("select * from global_temp.binlog_test a left join hydee.dm_test b on a.id = b.id");
        sqlContext.tables().show();
        StreamingQuery streamingQuery = d.writeStream()
                .outputMode("append")
                .format("console")
[...]

PR #1104, opened 2019-09-16
Support reading data from TiDB in Spark Structured Streaming 

guliangliangatpingcap:
close https://github.com/pingcap/tispark/issues/1102

When selecting data from tidb using spark structured streaming, the data selecting from tidb will never be updated.

The reason is that the `CoprocessorRDD's timestamp` will not be updated.

Update `CoprocessorRDD's timestamp` if in Streaming Context and apply on `Root LogicalPlan`.

### Check List 

Tests 

https://github.com/pingcap/tispark-test/blob/master/tispark-examples/src/main/scala/com/pingcap/tispark/examples/StructuredStreamingReadTiDB.scala

Related changes


core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -44,6 +44,20 @@ import org.joda.time.{DateTime, DateTimeZone}
...
 import scala.collection.mutable
 
+object TiStrategy {
+  private val assignedTSPlanCache = new mutable.WeakHashMap[LogicalPlan, Boolean]()
+
+  private def hasAssignedTS(plan: LogicalPlan): Boolean = {
+    assignedTSPlanCache.get(plan).isDefined
+  }
+
+  private def markAssignedTS(plan: LogicalPlan): Unit = {
```
zhexuany: markTSAssigned?

guliangliangatpingcap: nice

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -115,7 +130,16 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
 
   override def apply(plan: LogicalPlan): Seq[SparkPlan] = {
     val ts = tiContext.tiSession.getTimestamp
-    plan foreachUp applyStartTs(ts)
+
+    if (plan.isStreaming) {
```
zhexuany: I feel like we need some comment to explain why we do this.

guliangliangatpingcap: nice!

PR #1132, opened 2019-09-26
[DNM] Test Netty OOM

birdstorm:
Test Netty OOM problem


birdstorm:
I will delete this branch because the reason of OOM lies in Netty.

PR #1169, opened 2019-10-25
Support read from tiflash

zhexuany:
This PR allows TiSpark can access tiflash.

### DIFF
--- a/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
+++ b/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
@@ -29,6 +29,7 @@ import org.apache.spark.sql.catalyst.expressions.aggregate._
 import org.apache.spark.sql.catalyst.expressions.{AttributeReference, Expression, Literal, NamedExpression}
 import org.apache.spark.sql.execution.SparkPlan
 import org.apache.spark.sql.execution.aggregate.SortAggregateExec
+import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.types.{MetadataBuilder, StructField, StructType}
 import org.tikv.kvproto.Kvrpcpb.{CommandPri, IsolationLevel}
 
@@ -95,6 +96,9 @@ object TiUtil {
 
     true
   }
+
+  def allowTiFlashRead(sqlConf: SQLConf) =
+    sqlConf.getConfString(TiConfigConst.USE_TIFLASH, "false").toLowerCase.toBoolean
 
   def isSupportedFilter(expr: Expression,
                         source: TiDBRelation,
--- a/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
+++ b/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
@@ -97,8 +97,7 @@ object TiUtil {
     true
   }
 
-  def allowTiFlashRead(sqlConf: SQLConf) =
-    sqlConf.getConfString(TiConfigConst.USE_TIFLASH, "false").toLowerCase.toBoolean
+  def allowTiFlashRead(tiConf: TiConfiguration) = tiConf.isUseTiFlash
 
   def isSupportedFilter(expr: Expression,
                         source: TiDBRelation,
--- a/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
+++ b/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
@@ -97,8 +97,6 @@ object TiUtil {
     true
   }
 
-  def allowTiFlashRead(tiConf: TiConfiguration) = tiConf.isUseTiFlash
-
   def isSupportedFilter(expr: Expression,
                         source: TiDBRelation,
                         blacklist: ExpressionBlacklist): Boolean =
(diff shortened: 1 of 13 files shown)


core/src/main/scala/org/apache/spark/sql/execution/CoprocessorRDD.scala
```diff
@@ -75,8 +75,13 @@ trait LeafExecRDD extends LeafExecNode {
       b.append(s"with dag request: $dagRequest")
       b.toString()
     } else {
-      s"TiDB $nodeName{$dagRequest}" +
-        s"${TiUtil.getReqEstCountStr(dagRequest)}"
+      if (dagRequest.getUseTiFlash) {
```
birdstorm: If the difference is just TiFlash/TiKV, we can preserve the rest to avoid duplicate code.

core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
```diff
@@ -96,6 +97,9 @@ object TiUtil {
     true
   }
 
+  def allowTiFlashRead(sqlConf: SQLConf) =
```
birdstorm: it seems weird that we have to read config through SQLConf...

core/src/main/scala/org/apache/spark/sql/tispark/TiRowRDD.scala
```diff
@@ -41,6 +40,47 @@ class TiRowRDD(override val dagRequest: TiDAGRequest,
...
+        .map(task => {
+          val learnerList = task.getRegion.getLearnerList
+          val learnerStore = learnerList
+            .collectFirst {
+              case peer =>
+                val store = regionMgr.getStoreById(peer.getStoreId)
+                if (store.getLabelsList
+                      .exists(
+                        label =>
+                          label.getKey == tiConf.getTiFlashLabelKey && label.getValue == tiConf.getTiFlashLabelValue
```
birdstorm: are we going to hard-code the label config?

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/ProtoConverter.java
```diff
@@ -251,22 +264,26 @@ protected Expr visit(ColumnRef node, Object context) {
...
-      return builder.build();
     } else {
-      DataType type = node.getType();
       builder.setTp(type.getProtoExprType());
       CodecDataOutput cdo = new CodecDataOutput();
       type.encode(cdo, EncodeType.PROTO, node.getValue());
       builder.setVal(cdo.toByteString());
     }
+    if (type.getType() == MySQLType.TypeTimestamp) {
+      type = DataTypeFactory.of(MySQLType.TypeDatetime);
```
birdstorm: very nice! can you add some comments for these lines? (or provide an issue link perhaps?)

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/ProtoConverter.java
```diff
@@ -308,16 +326,36 @@ protected Expr visit(IsNull node, Object context) {
...
+    builder.setFieldType(toPBFieldType(getType(node)));
     return builder.build();
   }
 
   @Override
   protected Expr visit(Not node, Object context) {
-    ScalarFuncSig protoSig = ScalarFuncSig.UnaryNot;
+    ScalarFuncSig protoSig = null;
+    DataType dataType = getType(node);
+    switch (dataType.getType()) {
```
birdstorm: how about we add a link to tipb?

tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
```diff
@@ -198,6 +206,7 @@ public long mask(long flags) {
 
   private TiTableInfo tableInfo;
   private List<TiPartitionDef> prunedParts;
+  private boolean useTiFlash;
```
birdstorm: is it possible that we can make it final?

core/src/main/scala/org/apache/spark/sql/tispark/TiRowRDD.scala
zhexuany: Yes

tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
zhexuany: We have to set the value from TiStrategy.

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/ProtoConverter.java
zhexuany: This is some code I forget to remove. :(

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/ProtoConverter.java
zhexuany: The difference is caused by upgrading tipb.

birdstorm:
@zanmato1984 PTAL

PR #1174, opened 2019-10-29
fix view

zhexuany:
This PR fixes TiSpark cannot parse view table's json.

ViewCols could be null. If such case happens, we do not need parse viewCols.

Tests 
 - Integration test

### DIFF
--- a/core/src/test/scala/org/apache/spark/sql/ViewTestSuite.scala
+++ b/core/src/test/scala/org/apache/spark/sql/ViewTestSuite.scala
@@ -4,17 +4,14 @@ class ViewTestSuite extends BaseTiSparkTest {
   private val table = "test_view"
 
   test("Test View") {
-    tidbStmt.execute(s"drop table if exists $table")
+    tidbStmt.execute(s"create table $table(qty INT, price INT);")
+    tidbStmt.execute(s"INSERT INTO $table VALUES(3, 50);")
+
     try {
-      tidbStmt.execute("drop view if exists v")
+      tidbStmt.execute(s"CREATE VIEW v AS SELECT qty, price, qty*price AS value FROM $table;")
     } catch {
       case _: Exception => cancel
     }
-
-    tidbStmt.execute(s"create table $table(qty INT, price INT);")
-
-    tidbStmt.execute(s"INSERT INTO $table VALUES(3, 50);")
-    tidbStmt.execute(s"CREATE VIEW v AS SELECT qty, price, qty*price AS value FROM $table;")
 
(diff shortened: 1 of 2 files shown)


core/src/test/scala/org/apache/spark/sql/ViewTestSuite.scala
```diff
@@ -4,23 +4,29 @@ class ViewTestSuite extends BaseTiSparkTest {
   private val table = "test_view"
 
   test("Test View") {
-    tidbStmt.execute(s"drop table if exists $table")
```
marsishandsome: what if there's some dirty data (e.g. a table named `test_view`) in database?

core/src/test/scala/org/apache/spark/sql/ViewTestSuite.scala
```diff
@@ -4,23 +4,29 @@ class ViewTestSuite extends BaseTiSparkTest {
...
+    tidbStmt.execute(s"create table $table(qty INT, price INT);")
+    tidbStmt.execute(s"INSERT INTO $table VALUES(3, 50);")
+
     try {
-      tidbStmt.execute("drop view if exists v")
```
marsishandsome: what if there's some dirty data (e.g. a view named `v`) in database?

core/src/test/scala/org/apache/spark/sql/ViewTestSuite.scala
zhexuany: afterAll will drop it.

core/src/test/scala/org/apache/spark/sql/ViewTestSuite.scala
zhexuany: afterAll will drop it.

core/src/test/scala/org/apache/spark/sql/ViewTestSuite.scala
marsishandsome: but we should still drop it before each test

marsishandsome: see this
image

marsishandsome:
@birdstorm would u please add more information to the commit message?

core/src/test/scala/org/apache/spark/sql/ViewTestSuite.scala
zhexuany: Is it partitionTableSuite? If you take a look at `afterAll`, you will find that table `t` is not dropped at `afterAll`.

PR #1192, opened 2019-11-07
Add columnar execution for TiSpark 

zhexuany:
In this PR, we add columnar execution for TiSaprk.


tikv-client/src/main/java/com/pingcap/tikv/chunk/ChunkWriter.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.Objects;
+import org.apache.arrow.vector.BigIntVector;
+import org.apache.arrow.vector.DateDayVector;
+import org.apache.arrow.vector.DecimalVector;
+import org.apache.arrow.vector.FieldVector;
+import org.apache.arrow.vector.Float8Vector;
+import org.apache.arrow.vector.TimeStampMicroTZVector;
+import org.apache.arrow.vector.ValueVector;
+import org.apache.arrow.vector.VarBinaryVector;
+import org.apache.arrow.vector.VarCharVector;
```
ilovesoup: Why we directly rely on Apache Arrow which is heavy and slow?
Please take a look on the implementation of TiFlash's version. I don't think this is what we want.

PR #1195, opened 2019-11-07
Optimize tests and import

birdstorm:
optimize some import in TiSpark and fix incorrect test output.


core/src/main/scala/com/pingcap/tispark/TiDBRelation.scala
```diff
@@ -93,9 +93,9 @@ case class TiDBRelation(session: TiSession,
       NamedExpression.newExprId
     )
 
-    val sortAgg = TiUtil
+    val agg = TiUtil
```
birdstorm: FYI, this is not a sort aggregation.

core/src/test/scala/org/apache/spark/sql/BaseTiSparkTest.scala
```diff
@@ -363,14 +361,33 @@ class BaseTiSparkTest extends QueryTest with SharedSQLContext {
...
       try {
         r1 = queryViaTiSpark(qSpark)
+        sparkPlan = getSparkPlan(qSpark)
       } catch {
         case e: Throwable =>
-          logger.error(s"TiSpark failed when executing: $qSpark")
-          fail(e)
+          try {
+            logger.error(s"TiSpark failed when executing: $qSpark")
+            logger.warn("failure detected: \n", sparkPlan)
```
birdstorm: will automatically output explain result when query fails.

core/src/test/scala/org/apache/spark/sql/BaseTiSparkTest.scala
```diff
@@ -401,9 +418,10 @@ class BaseTiSparkTest extends QueryTest with SharedSQLContext {
         if (!compSqlResult(qSpark, r1, r4, checkLimit)) {
           fail(
             s"""Failed with
-               |TiSpark:\t\t${listToString(r1)}
+               |TiFlash:\t\t${listToString(r4)}
```
birdstorm: Previous output result was not correct @zhexuany

core/src/main/scala/com/pingcap/tispark/TiDBRelation.scala
```diff
@@ -83,28 +82,10 @@ case class TiDBRelation(session: TiSession,
...
-    val sortAgg = TiUtil
-      .planAggregateWithoutPartial(
-        Seq(handlePlan.attributeRef.head), // group by region id
-        Seq(aggExpr),
-        Seq(handlePlan.output.head, aggExpr.resultAttribute), // output <region, handleList>
-        handlePlan
-      )
-      .children
-      .head
+    // TODO: we may optimize by partitioning the result by region.
```
zhexuany: we need create an issue to track this. It is also better to have the issue link.

PR #1207, opened 2019-11-13
Fix task retry if tikv is down

marsishandsome:
close https://github.com/pingcap/tispark/issues/1208

Retry fails if TiKV is down during spark task running

 if TiKV is down during spark task running, tispark will refetch leader from PD.

### Check List 

Tests 

1. set breakpoint on RegionStoreClient.java line 713
2. start a tidb cluster with 3 TiKV instance
3. select data from tidb
4. stop on breakpoint on RegionStoreClient.java line 713
5. kill a tikv instance
6. continue running

Related changes

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/operation/KVErrorHandler.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/operation/KVErrorHandler.java
@@ -248,6 +248,8 @@ public class KVErrorHandler<RespT> implements ErrorHandler<RespT> {
         BackOffFunction.BackOffFuncType.BoTiKVRPC,
         new GrpcException(
             "send tikv request error: " + e.getMessage() + ", try next peer later", e));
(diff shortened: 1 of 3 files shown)


tikv-client/src/main/java/com/pingcap/tikv/operation/KVErrorHandler.java
```diff
@@ -248,6 +248,6 @@ public boolean handleRequestError(BackOffer backOffer, Exception e) {
         BackOffFunction.BackOffFuncType.BoTiKVRPC,
         new GrpcException(
             "send tikv request error: " + e.getMessage() + ", try next peer later", e));
-    return true;
+    return false;
```
birdstorm: mind if we add a comment for this change?

marsishandsome: fixed

PR #1231, opened 2019-11-21
fix output offsets

zhexuany:
This PR adds field type to Constant and ColumnRef when encode proto. 

With this, tikv can switch to vectorize execution which can improve the performance. 

When we push down a timestamp literal, tidb will convert it to datetime type.  TiSpark should adopt this behavior. 

Related changes

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
@@ -276,17 +276,20 @@ public class TiDAGRequest implements Serializable {
   }
 
   public DAGRequest buildIndexScan() {
-    DAGRequest.Builder builder = buildScan(true);
-    return buildRequest(builder);
+    List<Integer> outputOffsets = new ArrayList<>();
+    DAGRequest.Builder builder = buildScan(true, outputOffsets);
+    return buildRequest(builder, outputOffsets);
   }
 
   public DAGRequest buildTableScan() {
+    List<Integer> outputOffsets = new ArrayList<>();
     boolean isCoveringIndex = isCoveringIndexScan();
-    DAGRequest.Builder builder = buildScan(isCoveringIndex);
-    return buildRequest(builder);
+    DAGRequest.Builder builder = buildScan(isCoveringIndex, outputOffsets);
+    return buildRequest(builder, outputOffsets);
   }
 
-  private DAGRequest buildRequest(DAGRequest.Builder dagRequestBuilder) {
+  private DAGRequest buildRequest(
+      DAGRequest.Builder dagRequestBuilder, List<Integer> outputOffsets) {
     checkNotNull(startTs, "startTs is null");
     checkArgument(startTs.getVersion() != 0, "timestamp is 0");
     DAGRequest request =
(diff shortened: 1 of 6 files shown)


tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
```diff
@@ -238,6 +238,8 @@ public long mask(long flags) {
   private final PushDownType pushDownType;
   private IdentityHashMap<Expression, DataType> typeMap;
   private double estimatedCount = -1;
+  private List<Integer> outputOffsetsToBeAdded = new ArrayList<>();
+  //  private TLongArrayList indexScanOutputOffsetsToBeAdded = new TLongArrayList();
```
marsishandsome: delete this line?

zhexuany: of course.

tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
```diff
@@ -385,7 +388,8 @@ private DAGRequest buildRequest(DAGRequest.Builder dagRequestBuilder) {
           throw new DAGRequestException("Incorrect index scan with zero column count");
         }
 
-        dagRequestBuilder.addOutputOffsets(colCount - 1);
+        outputOffsetsToBeAdded.add(colCount - 1);
+        //        dagRequestBuilder.addOutputOffsets(colCount - 1);
```
zhexuany: will delete it.

tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
```diff
@@ -313,6 +315,7 @@ private DAGRequest buildRequest(DAGRequest.Builder dagRequestBuilder) {
    * @return final DAGRequest built
    */
   private DAGRequest.Builder buildScan(boolean buildIndexScan) {
+    outputOffsetsToBeAdded.clear();
```
birdstorm: is there any reason that this outputOffsets should be used externally since it is local? will there be any concurrent issue?

tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
```diff
@@ -467,6 +474,7 @@ else if (col.getColumnInfo().isPrimaryKey() && tableInfo.isPkHandle()) {
         executorBuilder.clear();
         addPushDownFilters();
       } else {
+        dagRequestBuilder.addAllOutputOffsets(outputOffsetsToBeAdded);
```
birdstorm: how about we return a dagRequest without offsets and add the offsets afterwards? so we won't forget to add it when new logic is added.

zhexuany: done

tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
zhexuany: done

birdstorm:
format

PR #1232, opened 2019-11-21
Push decimal

ilovesoup:
In order to make decimal pushable to TiFlash cop, we need to remove checkoverflow and promote_precision. Both TiFlash and TiKV's cop has its own logic of promoting and checkoverflow we should just trust them for their own part.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/tispark/BasicExpression.scala
+++ b/core/src/main/scala/org/apache/spark/sql/tispark/BasicExpression.scala
@@ -13,7 +13,6 @@
  * limitations under the License.
  */
 
-
 package org.apache.spark.sql.tispark
 
 import java.sql.Timestamp
--- a/core/src/main/scala/org/apache/spark/sql/tispark/BasicExpression.scala
+++ b/core/src/main/scala/org/apache/spark/sql/tispark/BasicExpression.scala
@@ -65,10 +65,10 @@ object BasicExpression {
 
   def sameCopType(lhs: DataType, rhs: DataType): Boolean = {
     (lhs, rhs) match {
-      case (_: IntegralType, _: IntegralType) => true
-      case (_: DecimalType, _: DecimalType) => true
+      case (_: IntegralType, _: IntegralType)                           => true
+      case (_: DecimalType, _: DecimalType)                             => true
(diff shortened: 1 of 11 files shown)


birdstorm:
do we also need to remove `MakeDecimal` and `UnscaledValue`?

ilovesoup:
> do we also need to remove `MakeDecimal` and `UnscaledValue`?

when they will be added?

birdstorm:
> > do we also need to remove `MakeDecimal` and `UnscaledValue`?
> 
> when they will be added?

seems that it might be optimized in `sum(some_decimal)` when `some_decimal` can be optimized to `long`?

birdstorm:
I haven't tried it yet but it is stated in logic of `DecimalAggregates`

ilovesoup:
> I haven't tried it yet but it is stated in logic of `DecimalAggregates`

I'd rather first fix as it is. Can you please create a jira issue that MakeDecimal preventing pushdown?

birdstorm:
> > I haven't tried it yet but it is stated in logic of `DecimalAggregates`
> 
> I'd rather first fix as it is. Can you please create a jira issue that MakeDecimal preventing pushdown?

sure.

birdstorm:
TISPARK-420

birdstorm:
should format code using `mvn clean install -DskipTests=true`

core/src/main/scala/org/apache/spark/sql/tispark/BasicExpression.scala
```diff
@@ -130,6 +137,20 @@ object BasicExpression {
       case Like(BasicExpression(lhs), BasicExpression(rhs)) =>
         Some(StringRegExpression.like(lhs, rhs))
 
+      // Coprocessor has its own behavior of type promoting and overflow check
+      // so we simply remove it from expression and let cop handle it
+      case CheckOverflow(BasicExpression(expr), _) =>
```
zhexuany: should we remove MakeDecimal and UnscaledValue?

birdstorm: see discussions above....

zhexuany: ok.

PR #1242, opened 2019-11-25
disable set, enum and bit pushed down

zhexuany:
This PR disables set, enum and bit pushed down. 

The reason we do that is because TiKV does not support these type's pushdown. 

This PR closes https://github.com/pingcap/tispark/issues/1244

### DIFF
--- a/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
+++ b/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
@@ -28,8 +28,6 @@ import org.apache.spark.SparkConf
 import org.apache.spark.sql.DataFrame
 import org.apache.spark.sql.catalyst.expressions.aggregate._
 import org.apache.spark.sql.catalyst.expressions.{AttributeReference, Expression, Literal, NamedExpression}
-import org.apache.spark.sql.execution.SparkPlan
-import org.apache.spark.sql.execution.aggregate.SortAggregateExec
 import org.apache.spark.sql.types.{MetadataBuilder, StructField, StructType}
 import org.tikv.kvproto.Kvrpcpb.{CommandPri, IsolationLevel}
 
@@ -75,10 +73,9 @@ object TiUtil {
 
     if (expr.children.isEmpty) {
       expr match {
-        // bit/duration type is not allowed to be pushed down
+        // bit, duration, set, and enum type is not allowed to be pushed down
         case attr: AttributeReference if nameTypeMap.contains(attr.name) =>
-          val head = nameTypeMap.get(attr.name).head
-          return !head.isInstanceOf[BitType]
+          return nameTypeMap.get(attr.name).head.isPushDownSupported
         // TODO:Currently we do not support literal null type push down
         // when Constant is ready to support literal null or we have other
         // options, remove this.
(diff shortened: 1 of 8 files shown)


marsishandsome:
please add `why disables set, enum and bit pushed down?` in PR description

core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
```diff
@@ -75,10 +73,11 @@ object TiUtil {
 
     if (expr.children.isEmpty) {
       expr match {
-        // bit/duration type is not allowed to be pushed down
+        // bit, duration, set, and enum type is not allowed to be pushed down
         case attr: AttributeReference if nameTypeMap.contains(attr.name) =>
           val head = nameTypeMap.get(attr.name).head
           return !head.isInstanceOf[BitType]
+          return nameTypeMap.get(attr.name).head.isPushDownSupported
```
marsishandsome: two `returns`?

zhexuany: The second return is the real return. I'll fix it.

zhexuany: The second return is the real return. I'll fix it.

birdstorm: two replies and two returns... lmao

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -308,8 +308,13 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
     sortOrder: Seq[SortOrder]
   ): SparkPlan = {
     val request = new TiDAGRequest(pushDownType(), timeZoneOffsetInSeconds())
-    request.setLimit(limit)
-    addSortOrder(request, sortOrder)
+    val filteredSortOrder =
+      sortOrder.filter(order => TiUtil.isSupportedOrderBy(order, source, blacklist))
+    if (filteredSortOrder.nonEmpty) {
```
birdstorm: does this mean that limit will not be pushed down if there isn't an `order by` statement? It seems quite strange.

zhexuany: yes.

tikv-client/src/main/java/com/pingcap/tikv/types/DataType.java
```diff
@@ -322,6 +322,13 @@ protected boolean isPrefixIndexSupported() {
    */
   public abstract Object getOriginDefaultValueNonNull(String value, long version);
 
+  /*
+   * return true if this type can be pushed down to TiKV or TiFLASH
+   * */
```
marsishandsome: pls fix the format of the comment

zhexuany:
`/run-all-tests

tikv-client/src/main/java/com/pingcap/tikv/types/DataType.java
zhexuany: DONE

PR #1270, opened 2019-12-02
rewrite the logic of TiSpark reading data from TiFlash

zhexuany:
This PR rewrite the logic of TiSpark reading data from TiFlash. Original approach did not consider outdated region cache.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
+++ b/core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
@@ -103,6 +103,9 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
   private def isEnableChunk(): Boolean =
     sqlConf.getConfString(TiConfigConst.ENABLE_CHUNK, "true").toLowerCase.toBoolean
 
+  private def isUseTiFlash(): Boolean =
+    sqlConf.getConfString(TiConfigConst.USE_TIFLASH, "false").toLowerCase.toBoolean
+
   private def timeZoneOffsetInSeconds(): Int = {
     val tz = DateTimeZone.getDefault
     val instant = DateTime.now.getMillis
@@ -300,7 +303,7 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
 
     scanBuilder.buildTiDAGReq(
       allowIndexRead(),
-      tiContext.tiConf.isUseTiFlash,
+      isUseTiFlash(),
       tiColumns.map { _.getColumnInfo }.asJava,
       tiFilters.asJava,
       source.table,
--- a/tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/DAGIterator.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/DAGIterator.java
@@ -22,6 +22,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.tikv.kvproto.Coprocessor;
 import org.tikv.kvproto.Metapb;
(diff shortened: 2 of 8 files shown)


core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -103,6 +103,9 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
   private def isEnableChunk(): Boolean =
     sqlConf.getConfString(TiConfigConst.ENABLE_CHUNK, "true").toLowerCase.toBoolean
 
+  private def isUseTiFlash(): Boolean =
+    sqlConf.getConfString(TiConfigConst.USE_TIFLASH, "false").toLowerCase.toBoolean
+
```
marsishandsome: delete unused same config in `class TiConfiguration`?

```
  public boolean isUseTiFlash() {
    return useTiFlash;
  }
```

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/DAGIterator.java
```diff
@@ -218,6 +238,20 @@ private SelectResponse process(RangeSplitter.RegionTask regionTask) {
     return SelectResponse.newBuilder().addAllChunks(resultChunk).setEncodeType(encodeType).build();
   }
 
+  private Store getTiFlashStore(TiRegion region) {
```
marsishandsome: what if there are two replicas for tiflash? shall we return a `List<Store>`?

tikv-client/src/main/java/com/pingcap/tikv/region/TiRegion.java
```diff
@@ -87,8 +87,7 @@ public Peer getLeader() {
   }
 
   public List<Peer> getLearnerList() {
-    int peerCnt = getMeta().getPeersCount();
-    List<Peer> peers = new ArrayList<Peer>();
+    List<Peer> peers = new ArrayList<>();
```
marsishandsome: `getLearnerList` is not used, maybe we should delete this function

marsishandsome:
there is an unused import in `CoprocessorRDD.scala`  on line 36

marsishandsome:
in `CoprocessorRDD.scala` on line 47:
instead of `UnknownPartitioning`, can we use the partition in from `tiRDDs`?

tikv-client/src/main/java/com/pingcap/tikv/region/TiRegion.java
zhexuany: We use this function to get learner store.

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/DAGIterator.java
zhexuany: DONE.

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
zhexuany: We may use it in future.

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/DAGIterator.java
zhexuany: This will be addressed in another PR. It is about load balancing.

tikv-client/src/main/java/com/pingcap/tikv/region/TiRegion.java
zhexuany: It is used in CoprocessorIterator.process.

Issue #1272, opened 2019-12-02
check timezone on each spark node


marsishandsome:
check timezone on each spark node, throw error if timezone is different on different spark node

PR #1275, opened 2019-12-03
Add timezone check

marsishandsome:
close https://github.com/pingcap/tispark/issues/1272

if timezone is different on different machine, there will be some problems (e.g. https://github.com/pingcap/tispark/issues/1273) with timezone related data type, e.g. `date`、`timestamp`.

Add timezone check, if driver's timezone is different with executor's timezone, tispark will throw exception.

### Check List 

Tests 

 - No code

Related changes

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/tispark/TiHandleRDD.scala
+++ b/core/src/main/scala/org/apache/spark/sql/tispark/TiHandleRDD.scala
@@ -55,7 +55,7 @@ class TiHandleRDD(val dagRequest: TiDAGRequest,
       if (!tiConf.checkLocalTimeZone()) {
         throw new TiInternalException(
           "timezone are different! dirver: " + tiConf.getLocalTimeZone + " executor:" + Converter.getLocalTimezone +
-          " please set spark.driver.extraJavaOptions=-Duser.timezone=GMT+8 and spark.executor.extraJavaOptions -Duser.timezone=GMT+8"
+            " please set user.timezone in spark.driver.extraJavaOptions and spark.executor.extraJavaOptions"
         )
       }
 
--- a/core/src/main/scala/org/apache/spark/sql/tispark/TiHandleRDD.scala
+++ b/core/src/main/scala/org/apache/spark/sql/tispark/TiHandleRDD.scala
@@ -52,7 +52,7 @@ class TiHandleRDD(val dagRequest: TiDAGRequest,
 
   override def compute(split: Partition, context: TaskContext): Iterator[Row] =
     new Iterator[Row] {
-      if (!tiConf.checkLocalTimeZone()) {
+      if (!tiConf.getLocalTimeZone.equals(Converter.getLocalTimezone)) {
         throw new TiInternalException(
           "timezone are different! dirver: " + tiConf.getLocalTimeZone + " executor:" + Converter.getLocalTimezone +
             " please set user.timezone in spark.driver.extraJavaOptions and spark.executor.extraJavaOptions"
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/apache/spark/sql/tispark/TiHandleRDD.scala
```diff
@@ -50,6 +52,13 @@ class TiHandleRDD(val dagRequest: TiDAGRequest,
 
   override def compute(split: Partition, context: TaskContext): Iterator[Row] =
     new Iterator[Row] {
+      if (!tiConf.checkLocalTimeZone()) {
+        throw new TiInternalException(
+          "timezone are different! dirver: " + tiConf.getLocalTimeZone + " executor:" + Converter.getLocalTimezone +
+          " please set spark.driver.extraJavaOptions=-Duser.timezone=GMT+8 and spark.executor.extraJavaOptions -Duser.timezone=GMT+8"
```
birdstorm: I don't think +8 is the only choice for all users

tikv-client/src/main/java/com/pingcap/tikv/TiConfiguration.java
```diff
@@ -71,6 +74,14 @@ public static TiConfiguration createDefault(String pdAddrsStr) {
     return PDUtils.addrsToUrls(addrs);
   }
 
+  public DateTimeZone getLocalTimeZone() {
+    return localTimeZone;
+  }
+
+  public boolean checkLocalTimeZone() {
```
birdstorm: any reason why these functions should be in `TiConfiguration`?

marsishandsome: we want to add `driver's timezone` to Spark's Task, then serialize and send to spark executor. 

`TiConfiguration` is already added to Spark's Task.

core/src/main/scala/org/apache/spark/sql/tispark/TiHandleRDD.scala
zhexuany: +1

marsishandsome: error message is an example

marsishandsome: change to `spark.driver.extraJavaOptions=-Duser.timezone=${some time zone}` ?

birdstorm: I think suggesting "please set user.timezone in spark.driver.extraJavaOptions and spark.executor.extraJavaOptions" is enough

birdstorm: let the users decide what values to be set.

tikv-client/src/main/java/com/pingcap/tikv/TiConfiguration.java
birdstorm: I mean, there are many things that are sent to Spark executors, is there a better choice?

marsishandsome: why many things?

marsishandsome: add a `DateTimeZone` for each task

marsishandsome: removed

PR #1289, opened 2019-12-05
TISPARK-450: Fix tiflash task distributes as tikv

zanmato1984:
https://internal.pingcap.net/jira/browse/TISPARK-450

1. Add new structure `TiStoreType` identifying which store (TiKV/TiFlash) to read;
2. Let region task generation (by `RegionSplitter`) get store based on `TiStoreType`;
3. Point 2 also handles retrying on TiFlash region movement, as region task regeneration is still based on `TiStoreType`;
4. Hard code TiFlash engine label as every other project (TiKV/TiFlash/TiDB) does;
5. Remove some useless methods.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/region/RegionManager.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/region/RegionManager.java
@@ -202,13 +202,14 @@ public class RegionManager {
       Peer leader = region.getLeader();
       store = cache.getStoreById(leader.getStoreId());
     } else {
+      outerLoop:
       for (Peer peer : region.getLearnerList()) {
         Store s = getStoreById(peer.getStoreId());
         for (Metapb.StoreLabel label : s.getLabelsList()) {
           if (label.getKey().equals(storeType.getLabelKey())
               && label.getValue().equals(storeType.getLabelValue())) {
             store = s;
-            break;
+            break outerLoop;
           }
         }
       }
--- a/tikv-client/src/main/java/com/pingcap/tikv/region/RegionManager.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/region/RegionManager.java
@@ -36,6 +36,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 import org.apache.log4j.Logger;
+import org.tikv.kvproto.Metapb;
 import org.tikv.kvproto.Metapb.Peer;
 import org.tikv.kvproto.Metapb.Store;
 import org.tikv.kvproto.Metapb.StoreState;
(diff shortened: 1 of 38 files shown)


tikv-client/src/main/java/com/pingcap/tikv/region/RegionManager.java
```diff
@@ -184,26 +185,41 @@ public TiRegion getRegionById(long regionId) {
...
-    long storeId = leader.getStoreId();
-    return Pair.create(region, cache.getStoreById(storeId));
-  }
 
-  public Pair<TiRegion, Store> getRegionStorePairByRegionId(long id) {
-    TiRegion region = cache.getRegionById(id);
-    if (!region.isValid()) {
-      throw new TiClientInternalException("Region invalid: " + region.toString());
+    Store store = null;
+    if (storeType == null || storeType == TiStoreType.TiKV) {
```
birdstorm: when is it possible that `storeType` is null?

zanmato1984: Never. This is one thing I missed after several round of refining TiStoreType structure. Addressed.

tikv-client/src/main/java/com/pingcap/tikv/region/RegionManager.java
```diff
@@ -184,26 +185,41 @@ public TiRegion getRegionById(long regionId) {
...
+      Peer leader = region.getLeader();
+      store = cache.getStoreById(leader.getStoreId());
+    } else {
+      for (Peer peer : region.getLearnerList()) {
+        Store s = getStoreById(peer.getStoreId());
+        for (Metapb.StoreLabel label : s.getLabelsList()) {
+          if (label.getKey().equals(storeType.getLabelKey())
+              && label.getValue().equals(storeType.getLabelValue())) {
+            store = s;
+            break;
```
zhexuany: this `break` only breaks inner loop. Maybe you can use the following code 

```
outerloop:
 for (Peer peer : region.getLearnerList()) {
         Store s = getStoreById(peer.getStoreId());
         for (Metapb.StoreLabel label : s.getLabelsList()) {
           if (label.getKey().equals(storeType.getLabelKey())
               && label.getValue().equals(storeType.getLabelValue())) {
             store = s;
             break outerloop;
            }
         }
   }

```

zanmato1984: Nice catch. Addressed.

PR #1300, opened 2019-12-10
Add ci argument for region size

marsishandsome:
add a new argument for `/run-all-tests` to let tikv choose conf from
```
region-max-size = "144MB"
region-split-size = "96MB"
```
and
```
region-max-size = "2MB"
region-split-size = "1MB"
```

`/run-all-tests region=small` or `/run-all-tests region=normal`

default = `/run-all-tests region=normal`

### Check List 

Tests 

 - No code

### DIFF
--- a/config/tikv.toml
+++ b/config/tikv.toml
@@ -148,7 +148,8 @@ capacity = "8G"
 # bit smaller.
 # region-max-size = "144MB"
 # region-split-size = "96MB"
-region-split-size = "1MB"
+# region-max-size = "2MB"
+# region-split-size = "1MB"
 
 [rocksdb]
 # Maximum number of concurrent background jobs (compactions and flushes)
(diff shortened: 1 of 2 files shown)


config/tikv.toml
```diff
@@ -148,7 +148,8 @@ capacity = "8G"
 # bit smaller.
 # region-max-size = "144MB"
 # region-split-size = "96MB"
-region-split-size = "1MB"
+# region-max-size = "2MB"
```
zhexuany: should we remove commented code?

marsishandsome: will change it using `sed -i`

marsishandsome: sh "sed -i 's/\\# region-max-size = \\\"2MB\\\"/region-max-size = \\\"2MB\\\"/' config/tikv.toml"
                            sh "sed -i 's/\\# region-split-size = \\\"1MB\\\"/region-split-size = \\\"1MB\\\"/' config/tikv.toml"

PR #1301, opened 2019-12-10
unify common code in TiHandleRDD and TiRowRDD.

zhexuany:
This PR unifies some common code in `TiHandleRDD` and `TiRowRDD`.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/tispark/TiHandleRDD.scala
+++ b/core/src/main/scala/org/apache/spark/sql/tispark/TiHandleRDD.scala
@@ -55,12 +55,7 @@ class TiHandleRDD(override val dagRequest: TiDAGRequest,
 
   override def compute(split: Partition, context: TaskContext): Iterator[InternalRow] =
     new Iterator[InternalRow] {
-      if (!tiConf.getLocalTimeZone.equals(Converter.getLocalTimezone)) {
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/apache/spark/sql/tispark/TiHandleRDD.scala
```diff
@@ -55,12 +55,7 @@ class TiHandleRDD(override val dagRequest: TiDAGRequest,
 
   override def compute(split: Partition, context: TaskContext): Iterator[InternalRow] =
     new Iterator[InternalRow] {
-      if (!tiConf.getLocalTimeZone.equals(Converter.getLocalTimezone)) {
-        throw new TiInternalException(
-          "timezone are different! dirver: " + tiConf.getLocalTimeZone + " executor:" + Converter.getLocalTimezone +
-            " please set user.timezone in spark.driver.extraJavaOptions and spark.executor.extraJavaOptions"
-        )
-      }
+      checkTimezone
```
marsishandsome: i prefer to use `checkTimezone()` insteadof `checkTimezone`

it's easy to understand that `checkTimezone()` means `function call`

zhexuany: will do so.

PR #1304, opened 2019-12-11
cherry-pick using stream rather than removeIf

zhexuany:
This PR cherry pick #1057


birdstorm:
I think nothing is changed?

PR #1308, opened 2019-12-12
fix Residual Filter containing wrong info

zhexuany:
This PR fixes explain info having wrong information.

### DIFF
--- a/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
+++ b/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
@@ -255,10 +255,13 @@ object TiUtil {
         .tiRDDs
         .head
         .dagRequest
-    } else {
+    } else if(regionTaskExec.isDefined) {
       regionTaskExec.get
         .asInstanceOf[ColumnarRegionTaskExec]
         .dagRequest
+    } else {
+      throw new UnsupportedOperationException("cannot find ColumnarCoprocessorRDD or " +
+        "ColumnarRegionTaskExec in DataFrame.")
     }
   }
 }
--- a/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
+++ b/core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
@@ -248,11 +248,12 @@ object TiUtil {
   def extractDAGReq(df: DataFrame): TiDAGRequest = {
     val executedPlan = df.queryExecution.executedPlan
     val copRDD = executedPlan.find(e => e.isInstanceOf[ColumnarCoprocessorRDD])
-    val regionTaskExec = executedPlan.find(e => e.isInstanceOf[ColumnarCoprocessorRDD])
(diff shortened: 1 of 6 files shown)


marsishandsome:
need cherry-pick ?

birdstorm:
add some tests?

marsishandsome:
two `fix` on title ?

birdstorm:
consider move the test into PlanTestSuite?

core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
```diff
@@ -243,4 +244,20 @@ object TiUtil {
 
     mutableRow
   }
+
+  def extractDAGReq(df: DataFrame): TiDAGRequest = {
+    val executedPlan = df.queryExecution.executedPlan
+    val copRDD = executedPlan.find(e => e.isInstanceOf[ColumnarCoprocessorRDD])
+    val regionTaskExec = executedPlan.find(e => e.isInstanceOf[ColumnarCoprocessorRDD])
```
marsishandsome: `copRDD` is same as `regionTaskExec` ?

core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
```diff
@@ -243,4 +244,20 @@ object TiUtil {
 
     mutableRow
   }
+
+  def extractDAGReq(df: DataFrame): TiDAGRequest = {
```
marsishandsome: what if there are more than one or no `TiDAGRequest` in `df` ? show we return `List<TiDAGRequest>` ?

core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
marsishandsome: change to `val regionTaskExec = executedPlan.find(e => e.isInstanceOf[ColumnarRegionTaskExec])` ?

core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
zhexuany: We should only have one TiDAGRequest.

core/src/test/scala/org/apache/spark/sql/IssueTestSuite.scala
```diff
@@ -16,6 +16,7 @@
 package org.apache.spark.sql
 
 import com.pingcap.tispark.TiConfigConst
+import com.pingcap.tispark.utils.TiUtil
```
marsishandsome: unused import?

core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
marsishandsome: maybe zero ?

zhexuany: Originally, it is zero while IntelliJ gives me a warning.

PR #1318, opened 2019-12-14
rewrite get decimal from chunk

zhexuany:
This PR improves the performance of getting decimal.

MyDecimal uses a `wordBuf` to store 9 ints whose digits length is 10. We create a `BigDecinal` with `scale` by read the `wordBuf` as a `BigInteger`.

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
@@ -42,6 +42,7 @@ public class MyDecimal implements Serializable {
   private static final int ten9 = 1000000000;
   private static final int digMask = ten8;
   private static final int wordBase = ten9;
+  private static final BigInteger wordBaseBigInt = BigInteger.valueOf(ten9);
   private static final int wordMax = wordBase - 1;
   private static final int[] div9 =
       new int[] {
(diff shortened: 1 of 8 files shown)


tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
```diff
@@ -822,4 +756,89 @@ public void clear() {
...
+    int wordIdx = 0;
+    for (int i = this.digitsInt; i > 0; i -= digitsPerWord) {
+      x = x.multiply(BigInteger.valueOf(wordBase)).add(BigInteger.valueOf(this.wordBuf[wordIdx]));
+      wordIdx++;
+    }
+
+    int intWordIdx = wordIdx;
+    for (int i = this.digitsFrac; i > 0; i -= digitsPerWord) {
+      if (wordBuf[wordIdx] == 0) {
+        break;
```
windtalker: why

tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
```diff
@@ -280,8 +271,8 @@ private int countLeadingZeroes(int i, int word) {
     return leading;
   }
 
-  private int min(int a, int b) {
-    return Math.min(a, b);
+  private int min(int a) {
```
windtalker: `min` function with only one parameter looks confusing

tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
```diff
@@ -822,4 +756,89 @@ public void clear() {
...
+      x = x * powers10[digitsFrac];
+    }
+    if (negative) {
+      x = -x;
+    }
+    return x;
+  }
+
+  public BigDecimal toBigDecimal() {
+    if (digitsInt + digitsFrac < 19) {
```
ilovesoup: Please add comment for magic number.

tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
zhexuany: DONE

tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
zhexuany: DONE

tikv-client/src/main/java/com/pingcap/tikv/codec/MyDecimal.java
zhexuany: DONE

PR #1321, opened 2019-12-17
fix agg push down on count(*)

zhexuany:

We determine an expression can be pushed down to coprocessor by suing `isSupportedExpression` of `SupportedExpressionValidator`. After phase of `infer`, we forget to add `DataType` to typeMap. This PR fixes this.


zhexuany:
Closing this PR, since it will be fixed in #1327

PR #1333, opened 2019-12-24
Fix prefix index on blob

### LINKED ISSUE #1332: [BUG] Prefix Index may be incorrect when charset is not UTF8
https://github.com/pingcap/tispark/blob/0f5135c310b30e16827d7cbc59dc245621bee373/tikv-client/src/main/java/com/pingcap/tikv/types/Converter.java#L179-L188

The code did not consider the case when `prefixLength` is greater than length of `val`. So it might act abnormally when charset is not UTF8.

birdstorm:
Close #1332

### Check List 

Tests 

 - Integration test

Related changes


birdstorm:
@marsishandsome PTAL

Issue #182, opened 2017-12-29
Column pruning not handled properly


birdstorm:
<a href="https://github.com/Novemser"><img src="https://avatars1.githubusercontent.com/u/12718118?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [Novemser](https://github.com/Novemser)**
_Tue Nov 28 07:51:53 2017_
_Originally opened as https://github.com/pingcap/tikv-client-lib-java/issues/163_

----

SQL:
```sql
select id_dt from full_data_type_table where id_dt not in (select id_dt from full_data_type_table  where tp_decimal <> 1E10)
```

Plan:
```bash
== Physical Plan ==
*Project [id_dt#138L]
+- BroadcastNestedLoopJoin BuildRight, LeftAnti, ((id_dt#138L = id_dt#138L#258L) || isnull((id_dt#138L = id_dt#138L#258L)))
   :- TiDB CoprocessorRDD{[table: full_data_type_table] , Ranges: Start:[-9223372036854775808], End: [9223372036854775807], Columns: [id_dt], [tp_varchar], [tp_datetime], [tp_blob], [tp_binary], [tp_date], [tp_timestamp], [tp_year], [tp_bigint], [tp_decimal], [tp_double], [tp_float], [tp_int], [tp_mediumint], [tp_real], [tp_smallint], [tp_tinyint], [tp_char], [tp_nvarchar], [tp_longtext], [tp_mediumtext], [tp_text], [tp_tinytext], [tp_bit], [tp_time], [tp_enum], [tp_set]}
   +- BroadcastExchange IdentityBroadcastMode
      +- *Project [id_dt#138L AS id_dt#138L#258L]
         +- TiDB CoprocessorRDD{[table: full_data_type_table] , Ranges: Start:[-9223372036854775808], End: [9223372036854775807], Columns: [id_dt], [tp_decimal], Aggregates: Not(IsNull([tp_decimal])), <>([tp_decimal], 10000000000)}
```

Seem we didn't properly prune the plan and pushed down too many things.

birdstorm:
<a href="https://github.com/Novemser"><img src="https://avatars1.githubusercontent.com/u/12718118?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [Novemser](https://github.com/Novemser)**
_Tue Nov 28 08:05:08 2017_

----

@birdstorm After fixing this issue, regression testing is needed.

birdstorm:
[...]

PR #1337, opened 2019-12-27
support partition pruning on range columns partition 

zhexuany:
This PR wants to support partition pruning on range column. 

When a partition is range column, `lessThan` of `PartitionDefitions` will be at least one. The `expr` of `PartitionInfo` is an empty string. We cannot generate partition expression by only looking at `expr` like we said early: it is empty string. We need iterate all `columns` and `lessThan` to generate partition expression. 

Range Columns partition cannot accept expression but column's name. Additionally, the type of column is limited to int, or datetime related types. In order to build range, I change the logic of building range from building a long range to `typedKey` range. 

### Check List 
Tests 
 - Integration test

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/expression/RangePartitionPruner.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/expression/RangePartitionPruner.java
@@ -56,9 +56,10 @@ public class RangePartitionPruner {
   private List<TiPartitionDef> pruneRangeNormalPart(Expression cnfExpr) {
     Objects.requireNonNull(cnfExpr, "cnf expression cannot be null at pruning stage");
 
-    // we need rewrite filter expression. This step is designed to  deal with
-    // y < '1885-10-10' where y is a date type. Rewrite will apply partition expression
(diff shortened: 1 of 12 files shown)


zhexuany:
This PR cannot merge right now unless #1327 got merged first.

zhexuany:
This PR closes #1337 and #182

tikv-client/src/main/java/com/pingcap/tikv/expression/RangePartitionPruner.java
```diff
@@ -0,0 +1,154 @@
...
+      foundUnsupportedPartExpr = true;
+    }
+  }
+
+  private List<TiPartitionDef> pruneRangeNormalPart(Expression cnfExpr) {
+    Objects.requireNonNull(cnfExpr, "cnf expression cannot be null at pruning stage");
+
+    // we need rewrite filter expression. This step is designed to  deal with
+    // y < '1885-10-10' where y is a date type. Rewrite will apply partition expression
+    // on the constant part. After rewriting, it becomes y < 1995;
```
birdstorm: can you explain this comment more clearly?

PR #1351, opened 2020-01-03
Bump tipb version

zhexuany:
This PR upgrades tipb and kvproto's version. 

We first update commit hash in proto.sh. The second step is to resolving incompatible issue. 

### Check List 

Tests 

 - Integration test


tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/CoprocessorIterator.java
```diff
@@ -83,7 +83,8 @@
         session,
         SchemaInfer.create(dagRequest),
         dagRequest.getPushDownType(),
-        dagRequest.getStoreType()) {
+        dagRequest.getStoreType(),
+        dagRequest.getStartTs().getVersion()) {
```
birdstorm: question: can we always get the startts from dagrequest?

zhexuany: always.

tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
```diff
@@ -587,10 +588,6 @@ private void validateRequest(DAGRequest dagRequest) {
     // check encode type
     requireNonNull(dagRequest.getEncodeType());
 
-    // A DAG request must contain a valid timestamp
-    if (dagRequest.getStartTs() == 0) {
```
birdstorm: why is this line deleted?

zhexuany: because we do not need validate start ts any more?

birdstorm: okay, so where do we get the correct start ts in means of different version of TiKV?

birdstorm:
need cherry-pick?

tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
zhexuany: We set startTs in setStartsFallback if we meet an older version of TiKV.

zhexuany: We set startTs in setStartsFallback if we meet an older version of TiKV.

PR #1361, opened 2020-01-06
Optimize tispark performace

zhexuany:

This introduces a new codec format to improve the performance on TiFlash.


tikv-client/src/main/java/com/pingcap/tikv/expression/ArithmeticBinaryExpression.java
```diff
@@ -42,6 +42,11 @@ public static ArithmeticBinaryExpression minus(Expression left, Expression right
     return new ArithmeticBinaryExpression(left.dataType, MINUS, left, right);
   }
 
+  public static ArithmeticBinaryExpression multiply(
```
windtalker: what about other arithmetic binary expressions, do them need add similar function?

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -96,8 +96,17 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
   private def useStreamingProcess: Boolean =
     sqlConf.getConfString(TiConfigConst.COPROCESS_STREAMING, "false").toLowerCase.toBoolean
 
-  private def isEnableChunk: Boolean =
-    sqlConf.getConfString(TiConfigConst.ENABLE_CHUNK, "true").toLowerCase.toBoolean
+  // streaming only support TypeDefault. Even we enable chunk, we should still
+  // use TypeDefault.
+  private def getCodecFormat(): EncodeType = {
```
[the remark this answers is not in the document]
zhexuany: It is related. If the user  wants to use `TypeStreaming`,  we have to use `TypeDefault` even the user enable `Chunk` codec format.

birdstorm:
@zhexuany please resolve comments

zhexuany:
@birdstorm It seems we have concurrent issues. We are investigating it.

birdstorm:
Can you specify what concurrent issue you are encountering?

tikv-client/src/main/java/com/pingcap/tikv/expression/ArithmeticBinaryExpression.java
zhexuany: For now, we only find multiply and division need this.

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
```diff
@@ -96,8 +96,20 @@ case class TiStrategy(getOrCreateTiContext: SparkSession => TiContext)(sparkSess
...
+    val codecFormatStr =
+      sqlConf.getConfString(TiConfigConst.CODEC_FORMAT, "chblock").toLowerCase
+    if (useStreamingProcess) {
+      return EncodeType.TypeDefault
+    }
+    codecFormatStr match {
+      case "chunk"   => EncodeType.TypeChunk
+      case "chblock" => EncodeType.TypeCHBlock
```
windtalker: If the store type is tikv, we should not use chblock as the encode type

tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
```diff
@@ -162,7 +162,7 @@ public ExecutorService getThreadPoolForTableScan() {
         if (tableScanThreadPool == null) {
           tableScanThreadPool =
               Executors.newFixedThreadPool(
-                  conf.getTableScanConcurrency(),
+                  conf.getTableScanConcurrency() / conf.getPartitionPerSplit(),
```
windtalker: could you give some explanations of what problem do you want to solve by this change, and how/why it works?

core/src/main/scala/org/apache/spark/sql/TiStrategy.scala
zhexuany: TiKV will return data in `Default` codec. We decode data using codec type in resp.

tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
zhexuany: Okay, I  will add more comments.

PR #1409, opened 2020-04-20
BatchWrite: update primary key ttl

marsishandsome:
update primary key ttl

tidb support update ttl since v3.0.5: https://github.com/pingcap/tidb/pull/11979


marsishandsome:
@birdstorm PTAL

PR #1410, opened 2020-04-21
Batch Write: Optimize Key Partition

birdstorm:
@zhexuany 

A optimized version of key partition. Use binary search instead of linear search.


core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -916,7 +919,7 @@ case class WrappedEncodedRow(row: TiRow,
 
 // SimplifiedWrappedEncodedRow is used at last stage if write process.
```
marsishandsome, line 920:
```suggestion
// EncodedKVPair is used at last stage if write process.
```

core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -882,20 +872,33 @@ class TiBatchWrite(@transient val df: DataFrame,
...
+class TiRegionPartitioner(regions: util.List[TiRegion], writeConcurrency: Int) extends Partitioner {
+  def binarySearch(key: Key): Int = {
+    if (regions.get(0).contains(key)) {
+      return 0
+    }
+    var l = 0
+    var r = regions.size()
+    while (l < r) {
+      val mid = l + (r - l) / 2
+      val region = regions.get(mid)
```
marsishandsome, line 884: add following code?
```
if(region.contains(key)) 
    return mid
```

birdstorm: no need to return mid.

PR #1422, opened 2020-05-06
BatchWrite: fix split index

marsishandsome:

```
20/04/21 15:33:57 INFO TiDBJDBCClient: split index region: split table `tpch_test`.`BATCH.WRITE_LINEITEM` index idx_ship_date between (0) and (233906) regions 9
20/04/21 15:33:58 WARN TiDBJDBCClient: failed to split index region
com.mysql.jdbc.MysqlDataTruncation: Data truncation: Incorrect datetime value: '2023-39-6'
```

use values on `index column` instead of `handler`

https://pingcap.com/docs-cn/stable/reference/sql/statements/split-region/#split-index-region

### Check List 

Tests 

 - UnitTest

### DIFF
--- a/core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
+++ b/core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
@@ -15,6 +15,7 @@
 
 package com.pingcap.tispark
 
+import java.sql.SQLException
 import java.util
 
 import com.pingcap.tikv.allocator.RowIDAllocator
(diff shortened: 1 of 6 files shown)


core/src/main/scala/com/pingcap/tispark/TiBatchWrite.scala
```diff
@@ -857,22 +870,31 @@ class TiBatchWrite(@transient val df: DataFrame,
...
-                maxHandle,
+                minIndexValue,
+                maxIndexValue,
                 regionSplitNum
               )
-          } else {
-            logger.warn("region split is skipped")
+          } catch {
+            case e: SQLException =>
+              if (options.isTest) {
```
birdstorm, line 890: if this is a debug-only parameter, should it be caught and thrown somewhere else? e.g., in the test framework

PR #1424, opened 2020-05-08
BatchWrite: use snapshot.batchGet instead of snapshot.get

marsishandsome:
`snapshot.get` is too slow

use `snapshot.batchGet` instead of `snapshot.get`

### Check List 

Tests 

 - Unit test

### DIFF
--- a/core/src/test/scala/com/pingcap/tispark/TiBatchWriteSuite.scala
+++ b/core/src/test/scala/com/pingcap/tispark/TiBatchWriteSuite.scala
@@ -34,6 +34,9 @@ class TiBatchWriteSuite extends BaseTiSparkTest {
       Nil
 
   private val batchWriteTablePrefix = "BATCH.WRITE"
+  private val isPkHandlePrefix = "isPkHandle"
+  private val replacePKHandlePrefix = "replacePKHandle"
+  private val replaceUniquePrefix = "replaceUnique"
 
   override def beforeAll(): Unit = {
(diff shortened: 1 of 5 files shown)


core/src/test/scala/com/pingcap/tispark/TiBatchWriteSuite.scala
```diff
@@ -104,6 +550,21 @@ class TiBatchWriteSuite extends BaseTiSparkTest {
         val tableToWrite = s"${batchWriteTablePrefix}_$table"
         tidbStmt.execute(s"drop table if exists `$tableToWrite`")
       }
+
+      for (table <- tables) {
```
birdstorm, line 554: consider make these a function dropTable()...?

core/src/test/scala/com/pingcap/tispark/index/LineItemSuite.scala
```diff
@@ -63,19 +65,261 @@ class LineItemSuite extends BaseTiSparkTest {
...
+    assert(diff.size == dfCount2 * 2)
+  }
+
+  override def afterAll(): Unit = {
     try {
       setCurrentDatabase(database)
-      val tableToWrite = s"${batchWriteTablePrefix}_$table"
-      tidbStmt.execute(s"drop table if exists `$tableToWrite`")
+
+      {
```
birdstorm, line 300: same here

PR #1425, opened 2020-05-08
fix TiSpark decode error when using chunk encode type

windtalker:
fix TiSpark decode error when using chunk encode type

1. decode date/datetime/timestamp column correctly under chunk encode type
2. only use chunk/chblock encode type for `CoprocessorIterator<TiChunk>`

### Check List 

Tests 

 - Unit test
 - Integration test


tikv-client/src/main/java/com/pingcap/tikv/columnar/TiChunkColumnVector.java
```diff
@@ -152,7 +139,7 @@ private long getTime(int rowId) {
       // only return microsecond from epoch.
       Timestamp ts =
           new Timestamp(
-              year - 1900, month - 1, day, (int) hour, minute, second, (int) microsecond * 1000);
+              year - 1900, month - 1, day, hour, minute, second, (int) microsecond * 1000);
```
birdstorm, line 142: how do we consider about timezone? I think we should enable chunk encoding for tests.

windtalker: In chunk encode mode, the storage will return the data with timezone info, so we do not need to consider the timezone here. I already set the default encode type to chunk encode for tikv, so i think the tests are already running in chunk encode mode.

PR #1445, opened 2020-05-15
[DNM] TiFlash: Test CHBLOCK codec format

birdstorm:
Only a dummy PR that tests CHBLOCK codec format


birdstorm:
Test Result, following cases fail:

select _col1_ from t where _col2_ _op_ _val_;

- When _col1_ is BOOLEAN, BIT type, TiSpark's output is 0/1 rather than true/false.
- When _col1_ is **BLOB(PRIMARY KEY)** type, throws `getBinary not supported`.
- When _col1_ is DATE type, the result answer is INCORRECT.

- When _col2_ is BOOLEAN, BIT type, the result answer is INCORRECT.
- When _col2_ is BLOB type, throws `getBinary not supported`.

birdstorm:
Test complete, closing.

PR #1220, opened 2019-11-16
Support tidb chunk's data format and columnar execution in Spark 

zhexuany:
### What problem does this PR solve? 
In this PR, we support tidb chunk encoding. We also switch spark's execution model from row-wise to columnar.

PR #1459, opened 2020-05-22
Fix incorrect date parsing

birdstorm:

* Fix Incorrect date parsing in chunk and block encoding.
* Ignores some tests not wanted when using tiflash ci.

----

Cause: 

`java.time.LocalDate` should not be used, instead we should use `org.joda.time.LocalDate`.

Note: Spark uses Julian calendar before October 15, 1582, so 10 days will be skipped if we are using Gregorian calendar.

**MUST CHERRY-PICK WITH #1220**

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/columnar/TiChunkColumnVector.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/columnar/TiChunkColumnVector.java
@@ -22,7 +22,7 @@ import com.pingcap.tikv.util.JsonUtils;
 import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.sql.Timestamp;
-import java.time.LocalDate;
+import org.joda.time.LocalDate;
 
 /** An implementation of {@link TiColumnVector}. All data is stored in TiDB chunk format. */
 public class TiChunkColumnVector extends TiColumnVector {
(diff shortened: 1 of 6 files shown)


tikv-client/src/main/java/com/pingcap/tikv/columnar/TiChunkColumnVector.java
```diff
@@ -133,8 +133,8 @@ private long getTime(int rowId) {
       day = 1;
     }
     if (this.type instanceof DateType) {
-      // only return day from epoch
-      return LocalDate.of(year, month, day).toEpochDay();
+      LocalDate date = new LocalDate(year, month, day, null);
+      return date.toDate().getTime() / 24 / 3600 / 1000;
     } else if (type instanceof DateTimeType || type instanceof TimestampType) {
       // only return microsecond from epoch.
       Timestamp ts =
```
windtalker, line 140: is it ok to use Timestamp here?

birdstorm: yes, we are always using jvm time zone.

windtalker: Discussed offline, it is ok to use Timestamp here.

PR #1461, opened 2020-05-25
fix ci error when running in tispark + tiflash mode

windtalker:
fix ci error when running in tispark + tiflash mode

1. fix result verification logical to avoid false error
2. support `getBinary` in `TiBlockColumnVector`
3. return correct epoch day for date type
4. fix some data type error in dag request

### Check List 

Tests 

 - Integration test

### DIFF
--- a/core/src/test/scala/org/apache/spark/sql/QueryTest.scala
+++ b/core/src/test/scala/org/apache/spark/sql/QueryTest.scala
@@ -175,6 +175,7 @@ abstract class QueryTest extends SparkFunSuite {
       row.map {
         case null           => "null"
         case a: Array[Byte] => a.mkString("[", ",", "]")
+        case b: Boolean     => if (b) "1" else "0"
         case s              => s.toString
       }.mkString
     }
--- a/core/src/test/scala/org/apache/spark/sql/QueryTest.scala
+++ b/core/src/test/scala/org/apache/spark/sql/QueryTest.scala
@@ -187,20 +187,32 @@ abstract class QueryTest extends SparkFunSuite {
       try {
         if (lhs.length != rhs.length) {
           false
-        } else if (!isOrdered) {
-          comp(
-            lhs.sortWith((_1, _2) => sortComp(_1, _2)),
-            rhs.sortWith((_1, _2) => sortComp(_1, _2))
-          )
         } else {
-          implicit object NullableListOrdering extends Ordering[List[Any]] {
-            override def compare(p1: List[Any], p2: List[Any]): Int =
-              p1.contains(null).compareTo(p2.contains(null))
-          }
-          comp(
-            lhs.sortBy[List[Any]](x => x),
-            rhs.sortBy[List[Any]](x => x)
+          // the result order may be different in TiDB and TiSpark, so compare result as follows
+          // 1. compare directly
+          // 2. if 1 fails, use `NullableListOrdering` to sort the result, then compare
+          // 3. if 2 fails and the sql does not contains `order by`, use string compare the sort the result, then compare
+          var result = comp(
+            lhs,
+            rhs
           )
+          if (!result) {
+            implicit object NullableListOrdering extends Ordering[List[Any]] {
+              override def compare(p1: List[Any], p2: List[Any]): Int =
+                p1.contains(null).compareTo(p2.contains(null))
+            }
+            result = comp(
+              lhs.sortBy[List[Any]](x => x),
+              rhs.sortBy[List[Any]](x => x)
+            )
+          }
+          if (!result && !isOrdered) {
+            result = comp(
+              lhs.sortWith((_1, _2) => sortComp(_1, _2)),
+              rhs.sortWith((_1, _2) => sortComp(_1, _2))
+            )
+          }
+          result
         }
       } catch {
         // TODO:Remove this temporary exception handling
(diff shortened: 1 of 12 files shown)


core/src/test/scala/org/apache/spark/sql/QueryTest.scala
```diff
@@ -186,20 +187,32 @@ abstract class QueryTest extends SparkFunSuite {
...
         } else {
-          implicit object NullableListOrdering extends Ordering[List[Any]] {
-            override def compare(p1: List[Any], p2: List[Any]): Int =
-              p1.contains(null).compareTo(p2.contains(null))
-          }
-          comp(
-            lhs.sortBy[List[Any]](x => x),
-            rhs.sortBy[List[Any]](x => x)
+          // the result order may be different in TiDB and TiSpark, so compare result as follows
+          // 1. compare directly
```
birdstorm, line 192: if we know that TiDB/TiSpark result should be ordered, why should we compare directly?

tikv-client/src/main/java/com/pingcap/tikv/columnar/TiBlockColumnVector.java
```diff
@@ -178,13 +179,13 @@ private long getDateTime(int rowId) {
...
-    long ym = ymd >> 5;
+    long ymd = v >>> 41;
+    long ym = ymd >>> 5;
     int year = (int) (ym / 13);
     int month = (int) (ym % 13);
     int day = (int) (ymd & ((1 << 5) - 1));
-    LocalDate date = new LocalDate(year, month, day, null);
-    return date.toDate().getTime() / 24 / 3600 / 1000;
+    LocalDate date = new LocalDate(year, month, day);
+    return Math.floorDiv(date.toDate().getTime(), (3600 * 24 * 1000));
```
birdstorm, line 188: consider use MILLS_PER_DAY instead?

tikv-client/src/main/java/com/pingcap/tikv/columnar/TiChunkColumnVector.java
```diff
@@ -133,8 +133,8 @@ private long getTime(int rowId) {
       day = 1;
     }
     if (this.type instanceof DateType) {
-      LocalDate date = new LocalDate(year, month, day, null);
-      return date.toDate().getTime() / 24 / 3600 / 1000;
+      LocalDate date = new LocalDate(year, month, day);
+      return Math.floorDiv(date.toDate().getTime(), (3600 * 24 * 1000));
```
birdstorm, line 137: same here

tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/MetaResolver.java
```diff
@@ -56,7 +56,17 @@ protected Void visit(ComparisonBinaryExpression node, Expression parent) {
     // We may need add a expressionRewriter to address this.
     if (predicate != null) {
       visit(predicate.getColumnRef(), node);
-      predicate.getValue().setDataType(predicate.getColumnRef().getDataType());
+      // do not set the constant data type to the column ref data type because it may narrow the
+      // constant type, and cause wrong result.
+      // for example when the filter is `bit_col op long_constant`, set long_constant to bit type
+      // will
+      // truncated the long_constant, and may cause wrong result
```
birdstorm, line 63: please format these lines.

tikv-client/src/main/java/com/pingcap/tikv/types/DateType.java
```diff
@@ -99,6 +99,6 @@ protected Long decodeNotNull(int flag, CodecDataInput cdi) {
       return null;
     }
     // return how many days from EPOCH
-    return date.toDate().getTime() / (24 * 60 * 60 * 1000);
+    return Math.floorDiv(date.toDate().getTime(), (3600 * 24 * 1000));
```
birdstorm, line 102: same here.

tikv-client/src/main/java/com/pingcap/tikv/types/DataType.java
```diff
@@ -257,6 +257,10 @@ private long readLongLittleEndian(CodecDataInput cdi) {
         + ((long) (readBuffer[7] & 255) << 56));
   }
 
+  public boolean isSameCatalog(DataType other) {
```
birdstorm, line 260: how about rename to shouldNarrowDataTypeTo?

core/src/test/scala/org/apache/spark/sql/QueryTest.scala
windtalker: Because now there may be different data type for the same data, for example 1/0 and True/False, there is no easy way to sort the result so that the same data stay in the same position in `lhs/rhs`. And for most of the simple test(like `MultiDataTypeTests`), although there is no `order by` clause in the query, the result order is actually the data insertion order, we can compare it directly.

tikv-client/src/main/java/com/pingcap/tikv/types/DataType.java
windtalker: By `isSameCatalog`, I mean two different data type can be compared directly if them are in the same catalog, so I think `isSameCatalog` is better than `shouldNarrowDataTypeTo `

PR #1483, opened 2020-06-18
fix resolve lock bug with tiflash

marsishandsome:
resolve lock grpc request should be sent to TiKV, but currently it may be sent to TiFlash

send resolve lock grpc request to TiKV

### Check List 

Tests 

 - Unit test


core/src/test/scala/com/pingcap/tispark/tiflash/TiFlashSuite.scala
```diff
@@ -0,0 +1,127 @@
+package com.pingcap.tispark.tiflash
```
birdstorm, line 1: copyright

PR #1492, opened 2020-06-22
Support new row format encoding

birdstorm:


tikv-client/src/test/java/com/pingcap/tikv/codec/TableCodecV2Test.java
```diff
@@ -387,11 +393,49 @@ private void debug(Object[] a1, Object[] a2) {
...
+                }
+              })) {
+        assertTrue(equals(this.bytes, bytes));
+      } else {
+        // FIXME: temporary code just for reverse checking...
+        Row res = TableCodecV2.decodeRow(bytes, this.handle, this.tableInfo);
+        Object[] o = fromRow(res, this.tableInfo);
+        debug(this.value, o);
+        assertTrue(deepEquals(this.value, o));
+      }
```
birdstorm, line 433: Added this check because for float/double like `11.99`, its binary representation may not be consistent in IEEE 754 format due to precision loss.

birdstorm: ... so if float/double value exists in table info, we reverse check the result by first encode and then decode to see if the answer still matches.

birdstorm: After fixing float encoding, these lines are not needed.

tikv-client/src/main/java/com/pingcap/tikv/TiDBJDBCClient.java
```diff
@@ -67,6 +69,30 @@ public int getDelayCleanTableLock() throws IOException, SQLException {
...
+   *
+   * @return 1 if should not encode and write with new row format.(default) 2 if encode and write
+   *     with new row format.(default on v4.0.0 cluster)
+   * @throws SQLException if a database access error occurs
+   */
+  public int getRowFormatVersion() throws SQLException {
+    List<List<Object>> result = queryTiDBViaJDBC(TIDB_ROW_FORMAT_VERSION_SQL);
+    if (result.isEmpty()) {
+      // default set to 1
+      return 1;
```
windtalker, line 83: return `TIDB_ROW_FORMAT_VERSION_DEFAULT` instead of 1 ?

tikv-client/src/main/java/com/pingcap/tikv/TiDBJDBCClient.java
```diff
@@ -67,6 +69,30 @@ public int getDelayCleanTableLock() throws IOException, SQLException {
...
+    } else {
+      Object version = result.get(0).get(0);
+      if (version instanceof String) {
+        return Integer.parseInt((String) version);
+      } else if (version instanceof Number) {
+        return ((Number) version).intValue();
+      } else {
+        return 1;
```
windtalker, line 91: ditto

tikv-client/src/main/java/com/pingcap/tikv/codec/RowDecoderV2.java
```diff
@@ -0,0 +1,180 @@
...
+
+  private static Timestamp decodeTimestamp(byte[] val, DateTimeZone tz) {
+    ExtendedDateTime extendedDateTime =
+        DateTimeCodec.fromPackedLong(new CodecDataInputLittleEndian(val).readLong(), tz);
+    // Even though null is filtered out but data like 0000-00-00 exists
+    // according to MySQL JDBC behavior, it can chose the **ROUND** behavior converted to the
+    // nearest
+    // value which is 0001-01-01.
+    if (extendedDateTime == null) {
+      return DateTimeCodec.createExtendedDateTime(tz, 1, 1, 1, 0, 0, 0, 0).toTimeStamp();
```
windtalker, line 162: Do we have similar round behavior in default row decoder?

tikv-client/src/main/java/com/pingcap/tikv/TiDBJDBCClient.java
birdstorm: ok

tikv-client/src/main/java/com/pingcap/tikv/codec/RowDecoderV2.java
birdstorm: See https://github.com/pingcap/tispark/blob/316d552a0a742b3da93e69f41e115b50908dfbfb/tikv-client/src/main/java/com/pingcap/tikv/types/AbstractDateTimeType.java#L59-L70

Issue #1173, opened 2019-10-29
Support Spark-3.0


marsishandsome:
### Description
Adapt to Spark-3.0.

### Category
Feature

### Task List
* [x] Study new feature & API in Spark-3.0.
* [x] Code (https://github.com/pingcap/tispark/pull/1493).
* [ ] Pass regression tests.

GanttStart: 2020-07-01
GanttDue: 2020-08-24
GanttProgress: 70%

Tongzhenguo:
期待

Tongzhenguo:
快开发完了吧

marsishandsome:
预计月底能开发完

marsishandsome:
https://github.com/pingcap/tispark/pull/1606

PR #1493, opened 2020-06-24
support spark 3.0.0

marsishandsome:
close https://github.com/pingcap/tispark/issues/1173

support spark-3.0

#### scala version
spark-3.0.0 only support scala-2.12
spark-2.4.x support both scala-2.12 and scala-2.11
spark-2.3.x support only scala-2.11

if tispark wants to support spark-2.3 & spark-2.4 & spark-3.0, tispark should provide two versions:
- one for scala-2.11 (spark-2.3 & spark-2.4)
- another for scala-2.12 (spark-2.4 & spark-3.0)

#### SpecialSum (UT: DONE)
Literal.create(int value, long type), `int value` + `long type` do not match in spark-3.0

#### syntax `with as select` error  (DONE)
TPCHQuerySuite q15
```
 Table or view 'revenue0' not found in database 'tidb_tpch_test';
```

#### InsertIntoTable (DONE)
change to `InsertIntoStatement`

#### ExplainCommand (DONE)
add one more argument: `formatted: Boolean = false`

#### CacheTableCommand (DONE)
add one more argument: `options: Map[String, String]`

#### ParserInterface (DONE)
add one more function `override def parseMultipartIdentifier(sqlText: String): Seq[String]`

####  do not support `Decimal Type compare with null` (DONE)
- select  count(1)  from full_data_type_table  where tp_decimal >= null
- select  count(1)  from full_data_type_table  where tp_decimal != null
- ...

#### Database+Table => Catalog+Namespace+Table
- no `SetDatabaseCommand` & `ShowDatabasesCommand`, use `SetCatalogAndNamespace`
- `tableIdentifier: TableIdentifier` change to `tableIdentifier: Seq[String]`
- how to hack Catalog Metadata System?
- UT failed: support desc table
- UT failed: test large amount of tables
- UT failed: Test multiple Spark Session with different catalog
- UT failed: test support show columns

####  like with escape
TODO

#### builder order: wrapper before assembly
TODO

#### NPE when decoding datetime,date,timestamp
TODO

#### pyspark support
TODO

#### SparkR support
TODO


marsishandsome:
https://github.com/pingcap/tispark/pull/1606

PR #1568, opened 2020-08-20
BatchWrite: add parameter regionSplitUsingSize

marsishandsome:
Currently,  regions are split according to 960000 keys.
But if the table has a large number of columns, some huge regions will occur on Grafana.

The size of values should be used to split regions.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code


birdstorm (approved):
LGTM

PR #1569, opened 2020-08-20
BatchWrite: add backoffer for getRegion operation

marsishandsome:
BatchWrite: add backoffer for getRegion operation

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/region/RegionManager.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/region/RegionManager.java
@@ -65,7 +65,11 @@ public class RegionManager {
   }
 
   public TiRegion getRegionByKey(ByteString key) {
-    return cache.getRegionByKey(key);
+    return getRegionByKey(key, ConcreteBackOffer.newGetBackOff());
+  }
+
+  public TiRegion getRegionByKey(ByteString key, BackOffer backOffer) {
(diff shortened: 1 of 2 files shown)


tikv-client/src/main/java/com/pingcap/tikv/region/RegionManager.java
```diff
@@ -65,7 +65,11 @@ public RegionManager(
   }
 
   public TiRegion getRegionByKey(ByteString key) {
-    return cache.getRegionByKey(key);
+    return cache.getRegionByKey(key, ConcreteBackOffer.newGetBackOff());
```
birdstorm, line 68:
```suggestion
    return getRegionByKey(key, ConcreteBackOffer.newGetBackOff());
```

tikv-client/src/main/java/com/pingcap/tikv/region/RegionManager.java
```diff
@@ -123,7 +136,11 @@ public TiRegion getRegionById(long regionId) {
   }
 
   public Store getStoreById(long id) {
-    return cache.getStoreById(id);
+    return cache.getStoreById(id, ConcreteBackOffer.newGetBackOff());
```
birdstorm, line 139:
```suggestion
    return getStoreById(id, ConcreteBackOffer.newGetBackOff()); 
```

birdstorm (approved):
Others lgtm

PR #1572, opened 2020-08-20
BatchWrite: split region before prewrite

marsishandsome:
regions may be merged during prewrite because regions will be merged one hour after they are split

split region right before prewrite


birdstorm (approved):
LGTM

PR #1581, opened 2020-08-26
add test for fix: GenericInternalRow cannot be cast to ColumnarBatch

marsishandsome:
add test for https://github.com/pingcap/tispark/pull/1566


birdstorm (approved):
lgtm

PR #1582, opened 2020-08-26
optimize scatter region logic

marsishandsome:
the remaining regions will not scatter if an error occurs

continue to scatter the remaining regions if an error occurs


birdstorm (approved):
lgtm

PR #1583, opened 2020-08-26
prewrite primary key backoffer use prewriteBackOfferMS

marsishandsome:
currently prewrite key backoffer = 20s

prewrite primary key backoffer use parameter `prewriteBackOfferMS`


birdstorm (approved):
lgtm

PR #1584, opened 2020-08-26
set regionSplitUsingSize default value to true

marsishandsome:
regionSplitUsingSize=false

image

regionSplitUsingSize=true

image


birdstorm (approved):
LGTM

PR #1585, opened 2020-08-27
BatchWrite: use ShardRowIDBits to generate row id

marsishandsome:
currently `ShardRowIDBits` is ignored when generating row id

use `ShardRowIDBits` to generate row id

fix a bug:
- tidb starts row id with `1`
- but tispark starts with `0`

### Check List 

Tests 

 - Unit test
 - Integration test


tikv-client/src/main/java/com/pingcap/tikv/allocator/RowIDAllocator.java
```diff
@@ -47,6 +47,17 @@ private RowIDAllocator(long dbId, long step, TiConfiguration conf) {
     this.conf = conf;
   }
 
+  public static long getSharedRowId(long maxShardRowIDBits, long partitionIndex, long rowID) {
+    if (maxShardRowIDBits <= 0 || maxShardRowIDBits >= 16) {
+      return rowID;
+    }
+
+    // assert rowID < Math.pow(2, 64 - maxShardRowIDBits)
```
birdstorm, line 55: we can first call `shardRowBitsOverflow` for these checks.

birdstorm (approved):
LGTM

PR #1586, opened 2020-08-27
currently user provided auto increment value is not supported

marsishandsome:
`user-provided auto-increment value` is a little complicated.

disable it currently

### Check List 

Tests 

 - Unit test


birdstorm (approved):
LGTM

PR #1588, opened 2020-08-27
Fix partition table on tiflash

### LINKED ISSUE #1587: [BUG] Partition table on TiFlash returns empty result
**Describe the bug**

Partition table on TiFlash returns empty result.

**What did you do**
```
CREATE TABLE quarterly_report_status (
    report_id INT NOT NULL,
    report_status VARCHAR(20) NOT NULL,
    report_updated TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
)
PARTITION BY RANGE ( UNIX_TIMESTAMP(report_updated) ) (
    PARTITION p0 VALUES LESS THAN ( UNIX_TIMESTAMP('2020-07-01 00:00:00') ),
    PARTITION p1 VALUES LESS THAN ( UNIX_TIMESTAMP('2020-08-01 00:00:00') ),
    PARTITION p2 VALUES LESS THAN ( UNIX_TIMESTAMP('2020-09-01 00:00:00') ),
    PARTITION p3 VALUES LESS THAN (MAXVALUE)
);
insert into quarterly_report_status values (1,'a','2020-07-02 00:00:00'),(2,'b','2020-08-02 00:00:00'),(3,'c','2020-02-02 00:00:00');
alter table quarterly_report_status set tiflash replica 1;
select * from quarterly_report_status;
```

**What do you expect**

all data

**What happens instead**

empty result.

**Spark and TiSpark version info**
TiSpark v2.3

birdstorm:
Close #1587 

Partition table should use physical id rather than table id when building request.

Also, fix a performance issue relating to partition tables by only sending key ranges of one physical id at a time.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test

Related changes

### DIFF
--- a/core/src/main/scala/com/pingcap/tispark/TiDBRelation.scala
+++ b/core/src/main/scala/com/pingcap/tispark/TiDBRelation.scala
@@ -65,13 +65,13 @@ case class TiDBRelation(
 
   def logicalPlanToRDD(dagRequest: TiDAGRequest, output: Seq[Attribute]): List[TiRowRDD] = {
     import scala.collection.JavaConverters._
-    val ids = dagRequest.getIds.asScala
+    val ids = dagRequest.getPrunedPhysicalIds.asScala
     var tiRDDs = new ListBuffer[TiRowRDD]
     val tiConf = session.getConf
     tiConf.setPartitionPerSplit(TiUtil.getPartitionPerSplit(sqlContext))
     ids.foreach(id => {
       tiRDDs += new TiRowRDD(
-        dagRequest,
+        dagRequest.copyReqWithPhysicalId(id),
         id,
         TiUtil.getChunkBatchSize(sqlContext),
         tiConf,
@@ -85,7 +85,7 @@ case class TiDBRelation(
 
   def dagRequestToRegionTaskExec(dagRequest: TiDAGRequest, output: Seq[Attribute]): SparkPlan = {
     import scala.collection.JavaConverters._
-    val ids = dagRequest.getIds.asScala
+    val ids = dagRequest.getPrunedPhysicalIds.asScala
     var tiHandleRDDs = new ListBuffer[TiHandleRDD]()
     lazy val attributeRef = Seq(
       newAttributeReference("RegionId", LongType, nullable = false, Metadata.empty),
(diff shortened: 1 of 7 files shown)


tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
```diff
@@ -534,19 +556,17 @@ public TiTableInfo getTableInfo() {
 
   public TiDAGRequest setTableInfo(TiTableInfo tableInfo) {
     this.tableInfo = requireNonNull(tableInfo, "tableInfo is null");
+    setPhysicalId(tableInfo.getId());
     return this;
   }
 
-  public List<Long> getIds() {
```
birdstorm, line 540: This might have performance issue when a lot of partitions exist in table.

getIds() will be called for each partition's rdd.

tikv-client/src/main/java/com/pingcap/tikv/meta/TiDAGRequest.java
```diff
@@ -941,15 +965,14 @@ public String toString() {
...
     // Key ranges might be also useful
     if (!getRangesMaps().isEmpty()) {
-      sb.append(", KeyRange: ");
+      sb.append(", KeyRange: [");
       if (tableInfo.isPartitionEnabled()) {
         getRangesMaps()
-            .values()
             .forEach(
-                vList -> {
-                  for (int i = 9; i < vList.size(); i++) {
```
birdstorm, line 950: I'm sure this line was not reviewed carefully...

core/src/main/scala/com/pingcap/tispark/TiDBRelation.scala
```diff
@@ -65,13 +65,13 @@ case class TiDBRelation(
...
     import scala.collection.JavaConverters._
-    val ids = dagRequest.getIds.asScala
+    val ids = dagRequest.getPrunedPhysicalIds.asScala
     var tiRDDs = new ListBuffer[TiRowRDD]
     val tiConf = session.getConf
     tiConf.setPartitionPerSplit(TiUtil.getPartitionPerSplit(sqlContext))
     ids.foreach(id => {
       tiRDDs += new TiRowRDD(
-        dagRequest,
+        dagRequest.copyReqWithPhysicalId(id),
```
birdstorm, line 74: Request for each partition should only contain key ranges of the specific physical id.

The previous approach will contain all key ranges, that will be a N*N ranges in all.

marsishandsome (approved):

PR #1593, opened 2020-09-01
BatchWrite: split region according to insert data

marsishandsome:
currently, we split region according to not only `insert data` but also `delete data`

split region according to `insert data` only


birdstorm (approved):
lgtm

PR #1595, opened 2020-09-02
filter Tombstone store

marsishandsome:
tombstone store should be filtered when getting the minimal store version


windtalker (approved):
LGTM

PR #1597, opened 2020-09-03
Increase gRPC response default limit size to 2G

birdstorm:
Increase gRPC response default limit size to 2G. Because the region response may be larger than expected.

Modify doc's and config's default value.

### Check List 

Tests 

 - No code

Related changes


marsishandsome (approved):
LGTM

PR #1603, opened 2020-09-07
Fix multiple database scan test bug

birdstorm:
Possible bug when meta data of database number is more than 100 in one region. 

fix scanIterator.

### Check List 

Tests 

 - Unit test

Related changes


marsishandsome (approved):
LGTM

PR #1631, opened 2020-09-24
use set to store resolvedLocks

marsishandsome:
currently use list to store resolvedLocks

use set to store resolvedLocks


birdstorm (approved):
lgtm

PR #1644, opened 2020-10-13
Fix alter primary key for write

marsishandsome:
auto-increment value should not use `shard row bits`

### Check List 

Tests 

 - Unit test


birdstorm (approved):
LGTM

PR #1649, opened 2020-10-16
support invisible index

marsishandsome:
tidb-5.0 supports support invisible index
see https://github.com/pingcap/tidb/pull/16914

close https://github.com/pingcap/tispark/issues/1650

tispark should also support invisible index
- decode `invisible index` from json schema
- ignore `invisible index` in planner

### Check List 

Tests 

 - Unit test
 - Integration test


birdstorm (approved):
LGTM

PR #1709, opened 2020-10-30
Fix typo in DateTime calculation

birdstorm:
fix typo in datetime calculation, which may cause the result be incorrect when using datetime like `0000-00-00 00:00:00`.

### Check List 

Tests 

 - No code

Related changes


marsishandsome (approved):
LGTM

Issue #1730, opened 2020-11-03
[BUG] Too many keys specified; max 64 keys allowed


marsishandsome:
regression test failed with tidb master branch

https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/tispark_regression_test_daily/detail/tispark_regression_test_daily/1890/pipeline

```
[2020-11-02T20:48:51.206Z] org.apache.spark.sql.TiDBMapDatabaseSuite *** ABORTED ***

[2020-11-02T20:48:51.206Z]   Exception encountered when invoking run on a nested suite - Failed to initialize SQLContext:Too many keys specified; max 64 keys allowed, please check your TiDB cluster and Spark configuration (SharedSQLContext.scala:97)
```

marsishandsome:
should revert this PR https://github.com/pingcap/tispark/pull/1760 when this issue fix

marsishandsome:
https://github.com/pingcap/tidb/pull/20353

TODO: add a configuration to tidb

marsishandsome:
https://github.com/pingcap/tidb/pull/21192

PR #1760, opened 2020-11-18
fix test with tidb master branch, Too many keys specified; max 64 key…

marsishandsome:
tidb master branch now only support max 64 keys in a table

should revert this PR when https://github.com/pingcap/tispark/issues/1730 fix

this PR will not cherry-pick to release branch


birdstorm (approved):
lgtm

PR #1910, opened 2021-02-05
fix jackson-databind

marsishandsome:
A flaw was discovered in jackson-databind in versions before 2.9.10, 2.8.11.5 and 2.6.7.3, where it would permit polymorphic deserialization of a malicious object using commons-configuration 1 and 2 JNDI classes. An attacker could use this flaw to execute arbitrary code.

PATCHED VERSIONS ： 2.9.10.4


birdstorm (approved):
LGTM

PR #1919, opened 2021-03-01
disable index scan for clustered index table

marsishandsome:
1. fix table scan bug
2. disable index scan (because of bug)


marsishandsome:
@birdstorm PTAL

birdstorm (approved):
LGTM

PR #1923, opened 2021-03-03
Disable index scan on partition table

birdstorm:
Disable index scan on the partition table.

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/predicates/TiKVScanAnalyzer.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/predicates/TiKVScanAnalyzer.java
@@ -167,14 +167,18 @@ public class TiKVScanAnalyzer {
       }
     } else if (canUseTiKV && allowIndexScan) {
       minPlan.getFilters().forEach(dagRequest::addDowngradeFilter);
-      double minCost = minPlan.getCost();
-      for (TiIndexInfo index : table.getIndices()) {
-        if (supportIndexScan(index, table)) {
-          TiKVScanPlan plan =
-              buildIndexScan(columnList, conditions, index, table, tableStatistics, false);
-          if (plan.getCost() < minCost) {
-            minPlan = plan;
-            minCost = plan.getCost();
+      if (table.isPartitionEnabled()) {
+        // disable index scan
+      } else {
+        double minCost = minPlan.getCost();
+        for (TiIndexInfo index : table.getIndices()) {
+          if (supportIndexScan(index, table)) {
+            TiKVScanPlan plan =
+                buildIndexScan(columnList, conditions, index, table, tableStatistics, false);
+            if (plan.getCost() < minCost) {
+              minPlan = plan;
+              minCost = plan.getCost();
+            }
           }
         }
       }


shiyuhang0:
@birdstorm hi, do you remember why disable index scan on the partition table？This pr does not has any background

Some users want  index scan on the partition table. Is there any risk?

birdstorm:
> @birdstorm hi, do you remember why disable index scan on the partition table？This pr does not has any background
> 
> Some users want  index scan on the partition table. Is there any risk?

There was a bug for the index scan to run on partition tables that the scan will repeat on every partition on the partition table. It is due to the fact that the partition table was not fully functioning at the moment, and it cannot support all pruning methods supported by index scan.
When the partition table is ready to support pruning for the index scan, this PR can be reverted.

shiyuhang0:
I do not understand why the index scan need to repeat on every partition on the partition table.

for example:
```
CREATE TABLE employees (
    id INT NOT NULL,
    store_id INT NOT NULL,
		INDEX store_index (store_id)
)

PARTITION BY RANGE (store_id) (
    PARTITION p0 VALUES LESS THAN (6),
    PARTITION p1 VALUES LESS THAN (11),
    PARTITION p2 VALUES LESS THAN (16),
    PARTITION p3 VALUES LESS THAN (21)
);
```

If we execute `select * from employees where store_id<10`，p0 and p1 can be pruned. Then index scan can be performed in p0 and p1.

Even partition pruning is invalid, we just perform index scan in all partition tables.

so, why we need to disable it :(

birdstorm:
> @birdstorm  I do not understand why the index scan need to repeat on every partition on the partition table.
> 
> for example:
> ```
> CREATE TABLE employees (
>     id INT NOT NULL,
>     store_id INT NOT NULL,
> 		INDEX store_index (store_id)
> )
> 
> PARTITION BY RANGE (store_id) (
>     PARTITION p0 VALUES LESS THAN (6),
>     PARTITION p1 VALUES LESS THAN (11),
>     PARTITION p2 VALUES LESS THAN (16),
>     PARTITION p3 VALUES LESS THAN (21)
> );
> ```
> 
> If we execute `select * from employees where store_id<10`，p0 and p1 can be pruned. Then index scan can be performed in p0 and p1.
> 
> Even partition pruning is invalid, we just perform index scan in all partition tables.
> 
> so, why we need to disable it :(
> 

@shiyuhang0 we do not need to repeat it, it was a bug for the previous implementation because of the incorrect pruning method. That is exactly why we disabled the index scan on partition tables.

birdstorm:
> @birdstorm  I do not understand why the index scan need to repeat on every partition on the partition table.
> 
> for example:
> ```
> CREATE TABLE employees (
>     id INT NOT NULL,
>     store_id INT NOT NULL,
> 		INDEX store_index (store_id)
> )
> 
> PARTITION BY RANGE (store_id) (
>     PARTITION p0 VALUES LESS THAN (6),
>     PARTITION p1 VALUES LESS THAN (11),
>     PARTITION p2 VALUES LESS THAN (16),
>     PARTITION p3 VALUES LESS THAN (21)
> );
> ```
> 
> If we execute `select * from employees where store_id<10`，p0 and p1 can be pruned. Then index scan can be performed in p0 and p1.
> 
> Even partition pruning is invalid, we just perform index scan in all partition tables.
> 
> so, why we need to disable it :(
> 

@shiyuhang0 As for this particular SQL, the previous implementation will not be able to prune p0 and p1, so the index scan will be executed on every partition respectively.

shiyuhang0:
thanks

Issue #1939, opened 2021-03-11
[BUG] date type error on release branch


guliangliangatpingcap:
the following test failed on release-2.3 & release-2.4 branch， but success on master branch
```
test("test") {
    tidbStmt.execute("""
                       |drop table if exists `tispark_test`.`t`;
                       |CREATE TABLE `tispark_test`.`t` (
                       |  `col_date` date not null,
                       |  `col_int0` int(11) not null,
                       |  `col_int1` int(11) not null,
                       |  UNIQUE KEY (`col_int0`),
                       |  PRIMARY KEY (`col_date`)
                       |  );
                       |INSERT INTO `tispark_test`.`t` VALUES ('1025-10-26',11,111);
                       |""".stripMargin)
    val sql = "select * from t"
    spark.sql(s"explain $sql").show(200, false)
    spark.sql(s"$sql").show(200, false)
    runTest(sql, skipJDBC = true)
  }
```

```
Failed with
TiSpark:		[len: 1] = List(1025-10-20,11,111)
Spark With JDBC:[len: null] = null
TiDB:			[len: 1] = List(1025-10-26,11,111)
```

marsishandsome:
https://github.com/pingcap/tispark/pull/1961

PR #1941, opened 2021-03-11
disable date test in ClusteredIndexTest

guliangliangatpingcap:
disable date test in ClusteredIndexTest because of https://github.com/pingcap/tispark/issues/1939


birdstorm (approved):
LGTM

PR #1972, opened 2021-03-19
fix tidb_enable_clustered_index only support global variable

marsishandsome:

https://github.com/pingcap/tidb/pull/23270

tidb_enable_clustered_index only support global variable

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/core/src/test/scala/org/apache/spark/sql/IssueTestSuite.scala
+++ b/core/src/test/scala/org/apache/spark/sql/IssueTestSuite.scala
@@ -42,8 +42,8 @@ class IssueTestSuite extends BaseTiSparkTest {
       cancel("currently tidb instance does not support clustered index")
     }
     spark.sqlContext.setConf(TiConfigConst.USE_INDEX_SCAN_FIRST, "true")
+    enableClusteredIndex()
     tidbStmt.execute("""
-        |SET tidb_enable_clustered_index = 1;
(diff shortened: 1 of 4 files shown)


birdstorm (approved):
LGTM

core/src/test/scala/org/apache/spark/sql/IssueTestSuite.scala
```diff
@@ -43,7 +43,7 @@ class IssueTestSuite extends BaseTiSparkTest {
     }
     spark.sqlContext.setConf(TiConfigConst.USE_INDEX_SCAN_FIRST, "true")
     tidbStmt.execute("""
-        |SET tidb_enable_clustered_index = 1;
+        |SET GLOBAL tidb_enable_clustered_index = 1;
```
tangenta, line 46: It seems that the global variable cannot take effect immediately after you set it in one session.

> Similar to MySQL, changes to GLOBAL variables do not apply to either existing connections, or the local connection. Only new sessions reflect the changes to the value.

https://docs.pingcap.com/tidb/stable/sql-statement-set-variable#set-globalsession-variable

marsishandsome: got thx

PR #1981, opened 2021-03-24
Fix TiRegion race when invalidating regions in RegionManager

birdstorm:
This is a fix for release-2.1 only.

In the previous implementation of the RegionManager, when the regions are invalidated, it is possible that the region in RegionManager is either not correctly invalidated due to race, or not correctly re-fetched.

RegionManager has race when encountering the following processes:
1. `invalidateRegion()` may be inconsistent when the handler tries to invalidate the region in RegionCache, but the region was previously updated by another process (e.g., `updateLeader()`).
2. the region was invalidated in keyToRegionIdCache, but the corresponding range is not correct.
3. if the invalidated range was smaller than the original, it creates a dangerous key range, where all keys in this range do not have a corresponding region id.
4. when fetching the region from the cache using `getRegionByKey()`, we look first into the key range cache, where it points to the non-existing region.

Use CAS operations when trying to invalidate regions. Forbid the usage of the region id.

Future works:
For current usage, the cache invalidator has to find the region by its id because they are not on the same JVM. This operation should be carefully dealt with.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/.ci/integration_test.groovy
+++ b/.ci/integration_test.groovy
@@ -209,7 +209,6 @@ def call(ghprbActualCommit, ghprbCommentBody, ghprbPullId, ghprbPullTitle, ghprb
                         cp .ci/log4j-ci.properties core/src/test/resources/log4j.properties
                         bash core/scripts/version.sh
                         bash core/scripts/fetch-test-data.sh
-                        mv core/src/test core-test/src/
                         bash tikv-client/scripts/proto.sh
                         """
                         }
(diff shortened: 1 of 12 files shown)


marsishandsome (approved):
LGTM

JesusOfSurberbia:
LGTM

PR #1999, opened 2021-04-06
skip test bigint int ClusteredIndexTest because of issue 1997

marsishandsome:
skip test bigint in ClusteredIndexTest because of issue https://github.com/pingcap/tispark/issues/1997

revert this PR is issue 1997 is fixed


marsishandsome:
https://github.com/pingcap/tispark/pull/2002

Issue #2068, opened 2021-05-07
[BUG] The application does not fails when there are duplicate unique keys


fran123:
**Describe the bug**
I am using tispark 2.3.13 and I put my configuration with replace "false", and when I insert data with a unique key and it exists in the database the application fails, but when the unique key is only duplicated in the dataframe and not in the database it is only replaced and the application does not fail.

**What did you do**
1. Create table
 ```
CREATE TABLE `table_test` (
 `id` varchar(36) COLLATE utf8_general_ci NOT NULL,
 `name` varchar(36) COLLATE utf8_general_ci DEFAULT NULL,
 `school` varchar(36) COLLATE utf8_general_ci NOT NULL,
 PRIMARY KEY (`id`),
 UNIQUE KEY `test_unique_1` (`name`,`school`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci
```
2. Run my code 
```
val sparkConf = new SparkConf()
    .setIfMissing("spark.master", "local[*]")
    .setIfMissing("spark.sql.extensions", "org.apache.spark.sql.TiExtensions")
    .setIfMissing("spark.tispark.pd.addresses", "127.0.0.1:2379")
    .setIfMissing("spark.tispark.isolation_read_engines", "tiflash")
  val spark = SparkSession.builder.config(sparkConf).getOrCreate()
  import spark.implicits._
  val df = Seq(
    ("10", "n5", "n10"),
    ("11", "n5", "n10")
  ).toDF("id", "name", "school")
  val tidbOptions: Map[String, String] = Map(
    "tidb.addr" -> "127.0.0.1",
    "tidb.port" -> "4000",
    "tidb.user" -> "root",
    "tidb.password" -> "",
    "database" -> "db_test",
    "table"-> "table_test",
    "replace" -> "false"
  )
  df.write
    .format("tidb")
    .options(tidbOptions)
    .mode("append")
    .save()
```

**What do you expect**
The application should fail , because there is a duplicate unique key and I put the configuration with replace false 

**What happens instead**
Only insert one and the application doesn't fail

**Screenshots**
The data in spark
![image (2)](https://user-images.githubusercontent.com/3598916/117492521-50e24400-af37-11eb-8fe4-d45df3b9590a.png)

The result in Tidb
![image (3)](https://user-images.githubusercontent.com/3598916/117492613-740cf380-af37-11eb-9c1e-f10687c9050c.png)

[...]

PR #2069, opened 2021-05-10
BatchWrite: add argument deduplicate

marsishandsome:
close https://github.com/pingcap/tispark/issues/2068

tispark deduplicates data in DataFrame before writing to TiKV according to the primary key and unique key.
but some user may want a Error Exception if data duplicates.

add a new parameter `deduplicate`, default = true
- deduplicate=false: check if data duplicates, an error will be thrown
- deduplicate=true: do deduplicate before writing

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


birdstorm (approved):
LGTM

PR #2077, opened 2021-05-27
fix ci jenkins error: NotSerializableException

purelind:

fix ci jenkins error: java.io.NotSerializableException: java.util.regex.Matcher

jenkins pipeline needs to be able to serialize the state of global variables between nodes and sub-jobs in the build, so we set regex variable to null

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - set regex variable to null for jenkins could serialize the state of global variable

Side effects

Related changes


zhouqiang-cl (approved):
LGTM

PR #2089, opened 2021-07-23
Fix ResolveLock not called in Scan API

birdstorm:
ResolveLock was not called in Scan API, it would lead to 

* Sometimes the table/database info was not retrieved from TiKV, e.g., `table or view not found`
* Scan API may not get all KV pairs if some of them are locked by another transaction.

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/codec/MetaCodec.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/codec/MetaCodec.java
@@ -26,10 +26,13 @@ import com.pingcap.tikv.util.Pair;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.tikv.kvproto.Kvrpcpb;
 import org.tikv.kvproto.Kvrpcpb.KvPair;
 
 public class MetaCodec {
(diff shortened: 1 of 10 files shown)


tikv-client/src/main/java/com/pingcap/tikv/codec/MetaCodec.java
```diff
@@ -115,7 +118,7 @@ public static ByteString bytesGet(ByteString key, Snapshot snapshot) {
     List<Pair<ByteString, ByteString>> fields = new ArrayList<>();
     while (iterator.hasNext()) {
       Kvrpcpb.KvPair kv = iterator.next();
-      if (kv == null || kv.getKey() == null) {
+      if (kv == null || kv.getKey().isEmpty()) {
```
marsishandsome, line 121:
```suggestion
      if (kv == null || kv.getKey() == null || kv.getKey().isEmpty()) {
```

marsishandsome (commented):
this PR should port to client-java ?

tikv-client/src/main/java/com/pingcap/tikv/codec/MetaCodec.java
birdstorm: @marsishandsome key is non-null

marsishandsome (approved):
LGTM

PR #2096, opened 2021-08-13
Fix region not found

birdstorm:
When the leader switched to another node, the response may return RegionNotFound error.

We should deal with this situation by invalidating the region and re-fetch it from PD.


birdstorm:
closed because no changes are made.

PR #2101, opened 2021-09-03
fix region split: region split key should be a vaild key

marsishandsome:
fix region split: region split key should be a vaild key

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/KVClient.java
@@ -171,5 +171,4 @@ public class KVClient implements AutoCloseable {
       long version) {
     return new ConcreteScanIterator(conf, builder, startKey, endKey, version);
   }
-
 }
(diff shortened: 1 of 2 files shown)


birdstorm (approved):
LGTM

flowbehappy:
I think the `next()` function is quite dangerous because it will generate a none exists key, which both TiFlash and TiDB cannot recognize.

birdstorm:
> I think the `next()` function is quite dangerous because it will generate a none exists key, which both TiFlash and TiDB cannot recognize.

@flowbehappy  It is not. Generating a non-existing key should not be dangerous at all. On the contrary, we should never make an assumption that a region start/end key is decodable because the `SPLIT REGION` API does not check the key at all, and could be called by any other roles in the cluster. 

Also, there are quite a few scenarios where we need to use a non-existing key. e.g., when building a raw key range that is more than `A`, we should explicitly use `next()`.

PR #2106, opened 2021-09-08
Optimize region cache miss

birdstorm:
Optimize region cache miss by removing regionId cache.


shiyuhang0:
@birdstorm which error would this bug cause?
We met `Coprocessor task terminated due to exceeding the deadline` after we switch client-java to org.tikv.client-java which not contains this pr. Can cache miss cause too much retry and finally get the error?

birdstorm:
> @birdstorm which error would this bug cause?
> We met `Coprocessor task terminated due to exceeding the deadline` after we switch client-java to org.tikv.client-java which not contains this pr. Can cache miss cause too much retry and finally get the error?

@shiyuhang0 it is possible that cache miss result in too many retries. This PR optimizes the scenario where region cache expires frequently.

PR #2156, opened 2021-11-07
fix duplicate range in RowIDAllocator

humengyu2012:


tikv-client/src/main/java/com/pingcap/tikv/allocator/RowIDAllocator.java
```diff
@@ -200,8 +217,10 @@ private long updateHash(
       return 0L;
     }
 
-    set(dataKey, newVal);
-    updateMeta(key, oldVal, snapshot);
+    ArrayList<BytePairWrapper> pairs = new ArrayList<>();
```
sunxiaoguang, line 220: You can create ArrayList with initialCapacity as two explicitly. Otherwise it creates an array that's larger than necessary.

tikv-client/src/main/java/com/pingcap/tikv/allocator/RowIDAllocator.java
```diff
@@ -125,20 +130,30 @@ public long getEnd() {
...
 
-  // set key value pair to tikv via two phase committer protocol.
-  private void set(ByteString key, byte[] value) {
+  // set key value pairs to tikv via two phase committer protocol.
+  private void set(List<BytePairWrapper> pairs) {
+    if (pairs == null || pairs.size() == 0) {
+      return;
+    }
     TiSession session = TiSession.getInstance(conf);
     TwoPhaseCommitter twoPhaseCommitter =
```
sunxiaoguang, line 139: you can write it like this
      Iterator<BytePairWrapper> iterator = pairs.iterator();
      BytePairWrapper primaryPair = iterator.next();
And use iterator directly in the next to rewrite secondary keys.

sunxiaoguang (commented):
REST LGTM

sunxiaoguang (approved):

sunxiaoguang:
image

The old value is not read in the transaction.

birdstorm (approved):
LGTM

Issue #2176, opened 2021-12-17
[BUG] some codes does not adheres to a java coding standard according to checkstyle


shiyuhang0:
**Describe the bug**
some codes does not adheres to a java coding standard according to checkstyle

**What did you do**
use checkstyle to check the java codes with the config below
```
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN" "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">

    <property name="charset" value="UTF-8"/>

    <property name="severity" value="warning"/>

    <property name="fileExtensions" value="java, properties, xml"/>


    <module name="LineLength">
        <property name="severity" value="error"/>
        <property name="fileExtensions" value="java"/>
        <property name="max" value="200"/>
        <property name="ignorePattern" value="^implements.*|^extends.*|^package.*|^import.*|a href|href|http://|https://|ftp://"/>
    </module>
    
    <module name="Translation">
        <property name="severity" value="error"/>
    </module>

    <module name="TreeWalker">
        
        <module name="OuterTypeFilename"/>
        <module name="PackageName" />
        <module name="TypeName"/>
        <module name="MemberName"/>
        <module name="ParameterName"/>
        <module name="CatchParameterName"/>
        <module name="LocalVariableName"/>
        <module name="MethodName"/>
        <module name="ConstantName">
            <property name="applyToPrivate" value="false" />
        </module>
        <module name="StaticVariableName"/>
        
        <module name="AvoidStarImport">
            <property name="severity" value="error"/>
            <property name="allowStaticMemberImports" value="true" />
        </module>
        <module name="UnusedImports">
            <property name="severity" value="error"/>
        </module>
        <module name="RedundantImport">
            <property name="severity" value="error"/>
        </module>
        
        <module name="ModifierOrder">
            <property name="severity" value="error"/>
        </module>
        <module name="RedundantModifier"/>
        
        <module name="AvoidNestedBlocks">
            <property name="severity" value="error"/>
        </module>

        <module name="EmptyBlock"/>

        <module name="NeedBraces">
            <property name="severity" value="error"/>
        </module>

        <module name="LeftCurly">
            <property name="severity" value="error"/>
        </module>

        <module name="RightCurly">
            <property name="severity" value="error"/>
        </module>
        
        <module name="SimplifyBooleanExpression">
            <property name="severity" value="error"/>
        </module>

[...]

PR #2178, opened 2021-12-17
fix: fix the java code problems find by checkstyle

shiyuhang0:
https://github.com/pingcap/tispark/issues/2176

improve the code quantity
trigger when in mvn validate progress

### Check List 

Tests 

 - Manual test
 use "mvn clean package -Dmaven.test.skip=true" to confirm the correct of changes

Side effects

- build will be fail when checkstyle fail

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -1,0 +1,133 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN" "https://checkstyle.org/dtds/configuration_1_3.dtd">
+<module name="Checker">
+    <!-- 字符集编码 -->
+    <property name="charset" value="UTF-8"/>
+    <!-- 违规级别 -->
+    <property name="severity" value="warning"/>
+    <!-- 需要检查的文件扩展名 -->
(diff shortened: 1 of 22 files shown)


zhangyangyu:
IMHO we should first decide how to apply the checkstyle plugin. Such code churns are usually not good idea (easy to introduce bugs, destroy good git commit history). I'd like to leave them unchanged and gradually fix the style in the future.

shiyuhang0:
> IMHO we should first decide how to apply the checkstyle plugin. Such code churns are usually not good idea (easy to introduce bugs, destroy good git commit history). I'd like to leave them unchanged and gradually fix the style in the future.

your suggetion seems beetet. However i have no idea how to apply checkstyle to the Incremental code. Need futhuer thinking

zhangyangyu:
> > IMHO we should first decide how to apply the checkstyle plugin. Such code churns are usually not good idea (easy to introduce bugs, destroy good git commit history). I'd like to leave them unchanged and gradually fix the style in the future.
> 
> your suggetion seems beetet. However i have no idea how to apply checkstyle to the Incremental code. Need futhuer thinking

Maybe we could make the checker as an optional merge requirement first, as a suggestion. Don't make it complex.

PR #2167, opened 2021-12-10
MOD: update how to import TiSpark jars for version >= 2.0

xuanyu66:
### What problem does this PR solve? 
The description on how to import TiSpark jar was outdated, Since from version 2.0 Spark-wrapper has been introduced into the project.

### Check List 

Tests 

 - No code

Related changes

 - Need to update the documentation

PR #2191, opened 2021-12-29
Revert "MOD: update how to import TiSpark jars for version >= 2.0"

xuanyu66:
Reverts pingcap/tispark#2167
Should fix the bug that classloader can't find the class rather than mod user doc.


xuanyu66:
@sunxiaoguang  PTAL

sunxiaoguang (approved):
LGTM

Issue #2189, opened 2021-12-28
[BUG] db-random-test module's test fail


shiyuhang0:
**Describe the bug**

db-random-test module's test fail
**What did you do**
run test 'DataGeneratorSuite' and 'RandomTestSuite' in the db-random-test

**What do you expect**

succuess
**What happens instead**

fail

**Reason**
the except value does't equals actual value just as the picture show
![image](https://user-images.githubusercontent.com/52435083/147558213-b6c97fb6-f0e2-4f99-b9dc-cdbdaf734ee5.png)

**Solution**
I don't known is the except value wrong or the actual value wrong, i prefer fix the  except value because the actual value seems correct in the whole

shiyuhang0:
The picture shows the code diff, because i change the code according to the actual value

PR #2045, opened 2021-04-25
refactor database random test framework

marsishandsome:
### What problem does this PR solve? 
 database random test needs refactor

### What is changed and how it works?
refactor database random test framework
- move test framework code to a new maven module `db-random-test` (other projects like `TiBigData` also need this test framework)
- add a base test class `com.pingcap.tispark.test.RandomTest`
- refactor existing random test cases
- fix problem: do not support generate data (unique) for primary key >= 2 columns

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change
 - Has persistent data change

Side effects

 - Possible performance regression
 - Increased code complexity
 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to update the `tidb-ansible` repository
 - Need to be included in the release note

PR #2192, opened 2021-12-30
Fix:db-random-test

shiyuhang0:
https://github.com/pingcap/tispark/issues/2189

Change the code because i think is a small bug for the extra space. Besides,it's why the test 'DataGeneratorSuite' fail.
The bug is introduced with https://github.com/pingcap/tispark/pull/2060 in IndexColumn.scala: https://github.com/pingcap/tispark/pull/2060/files#diff-22c8b596e739ebc64046c1c8314c0785d5487b1525834e0c76a37c14de45e43bR116

change the test because the code changes without the test changes.
- move clusteredIndex to primary key in Schema.scala: https://github.com/pingcap/tispark/pull/2060/files#diff-ad25d5a6aa3dabb5f06c0ba44a3b17787213469281c9a474349935f11593eec2R98
- change some sqls in SchemaAndData.scala: https://github.com/pingcap/tispark/pull/2045/files#diff-262c5d9036ebd92a6208a9233324b90d57608dbe2c72fe1bf0ebf6af55bb770eR29
### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

### DIFF
--- a/db-random-test/src/test/scala/com/pingcap/tispark/test/RandomTestSuite.scala
+++ b/db-random-test/src/test/scala/com/pingcap/tispark/test/RandomTestSuite.scala
@@ -30,10 +30,12 @@ class RandomTestSuite extends FunSuite with RandomTest {
 
   private val database = "random_test"
 
-  test("random test") {
+  test("random test with isClusteredIndex") {
     val schemaAndDataList = genSchemaAndData(
       rowCount,
-      List(INT, INT).map(d => genDescription(d, NullableType.NumericNotNullable)),
+      List(
+        genDescription(INT, NullableType.NotNullablePK),
+        genDescription(INT, NullableType.NumericNotNullable)),
       database,
       isClusteredIndex = true)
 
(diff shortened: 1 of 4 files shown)


db-random-test/src/test/scala/com/pingcap/tispark/test/RandomTestSuite.scala
```diff
@@ -43,12 +43,14 @@ class RandomTestSuite extends FunSuite with RandomTest {
     assert(rowCount == schemaAndData.data.size)
 
     val answer = """CREATE DATABASE IF NOT EXISTS `random_test`;
-                   |DROP TABLE IF EXISTS `random_test`.`test_351626634_1517918040`;
+                   |USE `random_test`;
+                   |DROP TABLE IF EXISTS `test_351626634_1517918040`;
                    |CREATE TABLE `random_test`.`test_351626634_1517918040` (
                    |  `col_int0` int(11) not null,
-                   |  `col_int1` int(11) not null /*T![clustered_index] CLUSTERED */
+                   |  `col_int1` int(11) not null
```
sunxiaoguang, line 50: Can you try verifying this with different TiDB versions. Clustered index was only available since TiDB 5.0.

we can use select version() to find out what version we are running testing against and make it works for different versions.

birdstorm: +1

shiyuhang0: This test is an unit test and has no relation with TiDB version. whether use Clustered index is hard code like this:
image

The test errors because someone had moved Clustered index to primary key, but the test is not be changed.
image
image

sunxiaoguang: > This test is an unit test and has no relation with TiDB version. whether use Clustered index is hard code like this: image
> 
> The test errors because someone had moved Clustered index to primary key, but the test is not be changed. image image

I see. I thought they are integration tests.

shiyuhang0:
I have added a test to verify the situation without cluster index

sunxiaoguang (approved):

sunxiaoguang:
@birdstorm PTAL

birdstorm (approved):
LGTM

Issue #2242, opened 2022-02-23
DSV2 support


shiyuhang0:
## Problem:

TiSpark uses Datasource API V1 in writing API and catalyst, And spark 3.0 has introduced the Datasource API V2 which allows for a lot of great new possibilities. In my view, Datasource API V2 at least has following advantages for TiSpark
- support delete  
- support spark streaming

Of course, there are lots of other benefits for V2. I think V2 is most likely going to be the key way to interact with datasources in the future for Spark so i'd like to replace V1 API to V2 API

## Goals
- Replace Datasource API V1 to V2
- Obsolete non catalog plugin
- More understandable user doc

## Solutions
1. Obsolete non catalog plugin mode
2. Move the dependency of API v1 in Catalyst
3. Replace API v1 to v2
4. Rewrite original writing code. Because API v2 no longer depends on RDD and SparkSession, which are used in original write logic.

## Compatibility
1. The config of catalog is must now . It means you can't operate TiDB without tidb_catalog anymore
```
spark.sql.catalog.tidb_catalog org.apache.spark.sql.catalyst.catalog.TiCatalog
spark.sql.catalog.tidb_catalog.pd.addresses ${your-pd-address}
```
2. The following usage is no longer supported, because is a datasource v1 read/write path
```
sprak.sql("use spark_catalog")  
spark.sql("create table xxx using tidb")
spark.sql("select * from xxx") 
spark.sql("insert into xxx")  
```
## Design
- https://github.com/pingcap/tispark/pull/2257

## Implementation
- Obsolete non catalog plugin mode: https://github.com/pingcap/tispark/pull/2252
- Move the dependency of API v1 in Catalyst: https://github.com/pingcap/tispark/pull/2259
- Replace API v1 to v2: https://github.com/pingcap/tispark/pull/2264
- ~~Rewrite original writing code~~ : Don't do it now for two reasons
  - spark doesn't provide a suitable framework for processing global data
  - the original write logical conflicts with the optimization of the catalyst. such as data type convert

PR #2252, opened 2022-02-24
Obsolete non catalog plugin mode

shiyuhang0:
https://github.com/pingcap/tispark/issues/2242

- Delete the corresponding codes
  - TiParser
  - TiDDLRule
  - org.apache.spark.sql.execution.command
  - TiResolutionRule
  - TiCompositeSessionCatalog
  - TiConcreteSessionCatalog
  - TiDirectExternalCatalog
- Fix some codes
  - Some tests
  - TiAuthIntegrationSuite
- Update user doc
- Update CI

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - User can't visit TiDB without catalog plugin anymore.


shiyuhang0:
PTAL @zhangyangyu @xuanyu66 @birdstorm

core/src/main/scala/org/apache/spark/sql/TiContext.scala
```diff
@@ -48,10 +48,7 @@ class TiContext(val sparkSession: SparkSession) extends Serializable with Loggin
     } else Option.empty)
   final val tiSession: TiSession = TiSession.getInstance(tiConf)
   lazy val sqlContext: SQLContext = sparkSession.sqlContext
-  lazy val tiConcreteCatalog: TiSessionCatalog =
-    new TiConcreteSessionCatalog(this)(new TiDirectExternalCatalog(this))
   lazy val sessionCatalog: SessionCatalog = sqlContext.sessionState.catalog
```
xuanyu66, line 51: Should delete this?

core/src/main/scala/org/apache/spark/sql/extensions/rules.scala
```diff
@@ -26,7 +26,8 @@ class TiAuthRuleFactory(getOrCreateTiContext: SparkSession => TiContext)
     extends (SparkSession => Rule[LogicalPlan]) {
   private val logger = LoggerFactory.getLogger(getClass.getName)
   override def apply(sparkSession: SparkSession): Rule[LogicalPlan] = {
-    if (TiExtensions.authEnable(sparkSession)) {
+    // auth only work for catalog plugin mode
+    if (TiExtensions.catalogPluginMode(sparkSession) && TiExtensions.authEnable(sparkSession)) {
```
xuanyu66, line 30: We should check catalogPluginMode before all factories, and throw Exception if catalogPluginMode is off.
SInce nonecatalogPluginMode is not supported anymore, we don't need to maintain else condition in all factories

core/src/test/scala/org/apache/spark/sql/BaseTiSparkTest.scala
```diff
@@ -99,24 +98,15 @@ class BaseTiSparkTest extends QueryTest with SharedSQLContext {
         }
       }
     } else {
```
xuanyu66, line 100: we don't need to test without catalog

docs/authorization_userguide.md
```diff
@@ -48,18 +48,19 @@ spark.sqlContext.setConf("spark.sql.tidb.password", your_tidb_server_password)
...
+| USE DATABASE  | ✅ | ✅ |   
+| SHOW DATABASES| ✅  | ✅ |   
+| SHOW TABLES   | ✅  | ✅ |
+| DESCRIBE TABLE| ✅  | ✅ |
+| SHOW COLUMNS  | ❌ | ❌ |
+| DESCRIBE COLUMN| ❌ | ❌ |
 
 ## limitations
 
+- Only support catalog plugin mode
```
xuanyu66, line 63: It's not Auth limitation.

xuanyu66 (changes requested):

core/src/main/scala/org/apache/spark/sql/extensions/rules.scala
xuanyu66, line 30: From my point of view, we should rename catalogPluginMode to validateCatalog, Since we don't support without catalogMode.

core/src/main/scala/org/apache/spark/sql/TiContext.scala
shiyuhang0: i think it should be deleted. it's useless after we only support catalog plugin mode.

core/src/main/scala/org/apache/spark/sql/extensions/rules.scala
shiyuhang0: i will try to do it

shiyuhang0: good idea

core/src/main/scala/org/apache/spark/sql/TiExtensions.scala
```diff
@@ -34,14 +38,20 @@ class TiExtensions extends (SparkSessionExtensions => Unit) {
...
-      tiContextMap.get(sparkSession) match {
-        case Some(tiContext) => tiContext
-        case None =>
-          // TODO: make Meta and RegionManager independent to sparkSession
-          val tiContext = new TiContext(sparkSession)
-          tiContextMap.put(sparkSession, tiContext)
-          tiContext
+    // check catalog plugin mode
+    if (!TiExtensions.validateCatalog(sparkSession)) {
+      logger.error("TiSpark is not work in catalog plugin mode!")
```
xuanyu66, line 43: The description will cause misunderstanding, be more clear.
Don't use is, use model verbs such as "should, must".
And also add some description on how to resolve.

core/src/main/scala/org/apache/spark/sql/TiExtensions.scala
```diff
@@ -34,14 +38,20 @@ class TiExtensions extends (SparkSessionExtensions => Unit) {
...
-    synchronized {
-      tiContextMap.get(sparkSession) match {
-        case Some(tiContext) => tiContext
-        case None =>
-          // TODO: make Meta and RegionManager independent to sparkSession
-          val tiContext = new TiContext(sparkSession)
-          tiContextMap.put(sparkSession, tiContext)
-          tiContext
+    // check catalog plugin mode
+    if (!TiExtensions.validateCatalog(sparkSession)) {
```
xuanyu66, line 42: In my opinion, catalog validation should be added to apply(), since getOrCreateTiContext invoke would happen after Rule construction.

shiyuhang0: It may not suitable if add catalog validation to the existing apply(), so, i did that before. Maybe we can add a inject method before others.

core/src/main/scala/org/apache/spark/sql/TiExtensions.scala
```diff
@@ -65,8 +65,8 @@ object TiExtensions {
       .find(pair => TiCatalog.className.equals(pair._2)) match {
       case None =>
         logger.error(
-          "TiSpark is not work in catalog plugin mode! Please add catalog config in spark conf")
-        throw new TiInternalException("TiSpark is not work in catalog plugin mode")
+          "TiSpark must work in catalog plugin mode! Please add catalog config in spark conf")
```
xuanyu66, line 68: Perhaps `catalog plugin mode` is a deprecated description.
TiSpark must work with TiCatalog. Please add TiCatalog in spark conf.

core/src/main/scala/org/apache/spark/sql/TiExtensions.scala
```diff
@@ -65,8 +65,8 @@ object TiExtensions {
...
+        throw new TiInternalException("TiSpark must work in catalog plugin mode")
```
xuanyu66, line 69: The same as log, since the exception message is the first message that the user notices.

xuanyu66 (changes requested):

core/src/main/scala/org/apache/spark/sql/TiContext.scala
xuanyu66: It hasn't been deleted yet

xuanyu66 (changes requested):

core/src/main/scala/org/apache/spark/sql/TiExtensions.scala
shiyuhang0: i forget change log

xuanyu66 (approved):

sunxiaoguang (approved):

PR #2253, opened 2022-02-24
fix bit pushdown

shiyuhang0:
TiKV support bit push down in https://github.com/tikv/tikv/pull/11968. But it causes some tests of TiSpark fail. The following are the fail test
- Fix Bit Type default value bug
- Test Convert from java.lang.Short to BIT
- Test Convert from java.lang.Integer to BIT
- Test Convert from java.lang.Long to BIT

This pr fixes those tests. By the way, `Fix Bit Type default value bug` fails because of the TiKV bug. So, ignore it before TiKV fixes it

### DIFF
--- a/core/src/test/scala/org/apache/spark/sql/IssueTestSuite.scala
+++ b/core/src/test/scala/org/apache/spark/sql/IssueTestSuite.scala
@@ -247,7 +247,7 @@ class IssueTestSuite extends BaseTiSparkTest {
   }
 
   // https://github.com/pingcap/tispark/issues/1147
-  test("Fix Bit Type default value bug") {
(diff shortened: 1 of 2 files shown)


shiyuhang0:
TiKV revert bit push down in https://github.com/tikv/tikv/pull/12030.
So, we needn't push down bit too.
But the bug exists actually, so just fix the bug.

shiyuhang0:
PTAL @xuanyu66 @zhangyangyu @birdstorm

xuanyu66 (approved):

zhangyangyu (approved):

sunxiaoguang (approved):

PR #2264, opened 2022-03-11
replace v1 to v2

shiyuhang0:
https://github.com/pingcap/tispark/issues/2242

- add TiDBWriterBuilder, TiDBBatchWrite, TiDBDataWriterFactory, TiDBDataWrite
- remove the logical in TiDBRelation
- remove some  v1 interface
- add some tests

### Check List 

Tests 

 - Unit test
 - Integration test

### New feature
`dataset.writeto` or `dataframe.writeto` is available now, but it is experimental. Some of its behavior may not be the same as `dataframe.write`

example
```
val tidbOptions: Map[String, String] = Map(
  "tidb.addr" -> "127.0.0.1",
  "tidb.password" -> "",
  "tidb.port" -> "4000",
  "tidb.user" -> "root",
  "spark.tispark.pd.addresses" -> "127.0.0.1:2379"
)
val df = spark.sql("select * from tidb_catalog.$dbname.$tablename")
df.writeTo("tidb_catalog.$dbname.$tablename").options(tidbOptions).append()

```


shiyuhang0:
@xuanyu66 @birdstorm PTAL

birdstorm (approved):
LGTM

core/src/main/scala/com/pingcap/tispark/v2/TiDBTable.scala
```diff
@@ -103,6 +107,25 @@ case class TiDBTable(
   def logicalPlanToRDD(dagRequest: TiDAGRequest, output: Seq[Attribute]): List[TiRowRDD] = {
     getLogicalPlanToRDD(dagRequest, output, session, sqlContext, tableRef)
   }
+
+  override def newWriteBuilder(info: LogicalWriteInfo): WriteBuilder = {
+    var scalaMap = info.options().asScala.toMap
+    // insert sql without option
+    if (scalaMap.isEmpty) {
+      throw new TiBatchWriteException(
+        "option is neccessary. maybe you are using insert sql, it's not supported yet")
```
xuanyu66, line 116: The initial letter should be uppercase.
Please check the whole PR.
`maybe you are using insert sql, it's not supported yet`, I can't get it.

core/src/main/scala/com/pingcap/tispark/v2/TiDBTableProvider.scala
```diff
@@ -62,32 +62,25 @@ class TiDBTableProvider
...
+    require(options.get("table") != null, "Option 'table' is required.")
+    Identifier.of(Array(options.get("database")), options.get("table"))
+  }
+
+  def extractCatalog(options: CaseInsensitiveStringMap): String = {
+    "tidb_catalog"
+  }
+
+  // df.write still go v1 path now
+  // because v2 path will go through catalyst, which may block some thing like datatype convert.
```
xuanyu66, line 76: something

core/src/main/scala/org/apache/spark/sql/TiContext.scala
```diff
@@ -71,61 +71,6 @@ class TiContext(val sparkSession: SparkSession) extends Serializable with Loggin
   val autoLoad: Boolean =
     conf.getBoolean(TiConfigConst.ENABLE_AUTO_LOAD_STATISTICS, defaultValue = true)
 
-  // tidbMapTable does not do any check any meta information
```
xuanyu66, line 74: image
Can you double-check it

core/src/main/scala/org/apache/spark/sql/TiExtensions.scala
```diff
@@ -72,8 +76,8 @@ object TiExtensions {
   }
 
   /**
-   * use TiAuthorizationRule to judge if TiExtensions is enable.
-   * it needs to be changed if TiAuthorizationRule is deleted
+   * Use TiAuthorizationRule to judge if TiExtensions is enable.
+   * It needs to be changed if TiAuthorizationRule is deleted
```
xuanyu66, line 80: when TiAuthorizationRule is not a must-have

xuanyu66 (changes requested):

core/src/main/scala/com/pingcap/tispark/v2/TiDBTable.scala
shiyuhang0: Insert SQL will go into this method, but it will fail because of the missing of option

core/src/main/scala/org/apache/spark/sql/TiContext.scala
shiyuhang0: The doc is outdated, only TiSpark >=2.0 and <2.3 need tidbMapDatabase. I will delete this Note for we don't support TiSpark before 2.3

shiyuhang0:
@xuanyu66 I have fixed the problems, PTAL again

xuanyu66 (approved):

sunxiaoguang (approved):

Issue #2268, opened 2022-03-17
[BUG] Select SQL return wrong result in cluster index with varchar pk


shiyuhang0:
**Describe the bug**
select SQL return wrong result

**What did you do**
1. config `spark.tispark.show_rowid=true`
2. JDBC : `create table $dbtable(i varchar(64), s int,PRIMARY KEY (i)/*T![clustered_index] CLUSTERED */,unique key (s))`
3. JDBC: `insert into $dbtableWithPrefix values('0', 0),('1',1)`
4. spark.sql("select * from $dbtable").show

result
<img width="259" alt="image" src="https://user-images.githubusercontent.com/52435083/158833463-05e1ac72-ec54-49f9-853f-419b52ff85f4.png">

PR #2270, opened 2022-03-18
Fix copyTableWithRowId

shiyuhang0:
https://github.com/pingcap/tispark/issues/2268

copyTableWithRowId need to exclude isCommonHandle. isCommonHandle is cluster index without  _tidb_rowid


shiyuhang0:
@xuanyu66 @birdstorm PTAL

xuanyu66 (approved):

tikv-client/src/main/java/com/pingcap/tikv/meta/TiTableInfo.java
```diff
@@ -354,7 +354,7 @@ private TiColumnInfo copyColumn(TiColumnInfo col) {
   }
 
   public TiTableInfo copyTableWithRowId() {
-    if (!isPkHandle()) {
+    if (!isPkHandle() && !isCommonHandle) {
```
zhexuany, line 357: better to add some comments saying the change is made due to behavior shift between non-cluster index and cluster index.

shiyuhang0: This is a bug. The change makes the correct behavior of show _tidb_rowid. Do I need to add comments in this case?

shiyuhang0: I have added comments to clear it

zhexuany: > This is a bug. The change makes the correct behavior of show _tidb_rowid. Do I need to add comments in this case?

It is probably because back to non-cluster index days, we do not need to consider that. ;)

zhexuany (approved):

sunxiaoguang (approved):

PR #1987, opened 2021-03-30
Fix region manager race

birdstorm:
Signed-off-by: birdstorm <samuelwyf@hotmail.com>

### What problem does this PR solve? 
Fix region manager race, previous PR has been pushed to release-2.1 #1981 

### What is changed and how it works?
In accordance to #1981 

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has interface methods change

Side effects

 - Possible performance regression

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to be included in the release note

PR #1768, opened 2020-11-20
do not retry for some key error

windtalker:
### What problem does this PR solve? 
fix #1766 

### What is changed and how it works?

If the key error is not recoverable, do not retry.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change
 - Has persistent data change

Side effects

 - Possible performance regression
 - Increased code complexity
 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to update the `tidb-ansible` repository
 - Need to be included in the release note

PR #2279, opened 2022-03-29
Fix retry bug

shiyuhang0:
we should not retry for some errors in the prewrite primary key.

This has been done in https://github.com/pingcap/tispark/pull/1768. But It was broken in https://github.com/pingcap/tispark/pull/1987

So, just recover what has been done in https://github.com/pingcap/tispark/pull/1768.


shiyuhang0:
@birdstorm PTAL if this change will affect https://github.com/pingcap/tispark/pull/1987

xuanyu66 (approved):

PR #2287, opened 2022-04-08
support spark 3.2

shiyuhang0:
support spark 3.2

### Solution
1. Add spark-3.2 profile
2. Add Spark-3.2 module in spark-wrapper and add Code in spark-3.2
3. Fix the code in spark-core to be compatible with spark 3.0, 3.1 and 3.2
   -  obsolete `V1WriteBuilder` that has changed to `WriteBuilder` in spark 3.2
   - `SpecialSum` and `ColumnarRegionTaskExec` need add method `withNewChildrenInternal` that is required to override  in spark 3.2
   - delete `seesion` parameter in `ColumnarRegionTaskExec` because spark 3.2 add a parameter with same name, and the original `seesion` parameter is useless
   - change the pattern matching in `Sum` and `Average` because they add one more parameter： matching on type rather than on case class
   - compile datavalue:LocalDate to 'datavalue' because LocalDate.toString has already formate it to "yyyy-mm-dd"
4. Update netty-all's version to 4.1.75-Final and exclude it in other dependencies. or you will get `java.lang.NoSuchMethodError: io.netty.buffer.PooledByteBufAllocator.<init>(ZIIIIIIZ)V` in spark 3.2
5. Update assembly to package spark-3.2 in spark wrapper
6. Fix Some tests

## Test
1. IT 
  - /run-all-tests profile=spark-3.2.1
  - /run-all-tests profile=spark-3.1.1
  - /run-all-tests (spark 3.0.2)
  - /run-all-tests profile=spark 3.2.1 tidb = v5.3.0 tikv = v5.3.0 pd = v5.3.0
2. Use Test
  - Package TiSpark with 3.0.2 and work with spark 3.0.2
  - Package TiSpark with 3.1.2 and work with spark 3.1.3
  - Package TiSpark with 3.2.1 and work with spark 3.2.1


shiyuhang0:
/run-all-tests profile=spark-3.2.1 tidb = v5.3.0 tikv = v5.3.0 pd = v5.3.0

shiyuhang0:
PTAL @birdstorm @xuanyu66

core/src/main/scala/com/pingcap/tispark/v2/TiDBTable.scala
```diff
@@ -298,10 +296,7 @@ object TiDBTable {
             DateTimeUtils.getZoneId(SQLConf.get.sessionLocalTimeZone))
           s"'${timestampFormatter.format(timestampValue)}'"
         case dateValue: Date => "'" + dateValue + "'"
-        case dateValue: LocalDate =>
-          val dateFormatter = DateFormatter(
-            DateTimeUtils.getZoneId(SQLConf.get.sessionLocalTimeZone))
-          s"'${dateFormatter.format(dateValue)}'"
+        case dateValue: LocalDate => "'" + dateValue + "'"
```
xuanyu66, line 299: Could you please explain in PR description?

core/src/main/scala/com/pingcap/tispark/v2/sink/TiDBWriteBuilder.scala
```diff
@@ -16,27 +16,12 @@
...
-import com.pingcap.tispark.write.{TiDBOptions, TiDBWriter}
-import org.apache.spark.sql.{DataFrame, SQLContext, SaveMode}
-import org.apache.spark.sql.connector.write.{LogicalWriteInfo, V1WriteBuilder}
-import org.apache.spark.sql.sources.InsertableRelation
+import com.pingcap.tispark.write.TiDBOptions
+import org.apache.spark.sql.SQLContext
+import org.apache.spark.sql.connector.write.{LogicalWriteInfo, WriteBuilder}
 
-case class TiDBWriterBuilder(
+case class TiDBWriteBuilder(
```
xuanyu66, line 23: Why should we change the class name?

core/src/test/scala/com/pingcap/tispark/auth/TiAuthIntegrationSuite.scala
```diff
@@ -164,8 +164,10 @@ class TiAuthIntegrationSuite extends SharedSQLContext {
   test(f"Show tables should not contain invisible table") {
     noException should be thrownBy spark.sql(s"use $databaseWithPrefix")
 
+    // spark 3.2 add isTemporary col when `show tables`, we need to exclude it.
     val tables = spark
       .sql(s"show tables")
+      .drop("isTemporary")
```
xuanyu66, line 170: Why should we exclude it, since this doesn't influence the test's result

xuanyu66 (changes requested):

core/src/main/scala/com/pingcap/tispark/v2/TiDBTable.scala
```diff
@@ -129,7 +129,6 @@ case class TiDBTable(
   override def capabilities(): util.Set[TableCapability] = {
     val capabilities = new util.HashSet[TableCapability]
     capabilities.add(TableCapability.BATCH_READ)
-    capabilities.add(TableCapability.V1_BATCH_WRITE)
```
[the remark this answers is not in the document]
shiyuhang0: Can't use `df.writeto` which is not exposed to the user yet.
obsolete it because its' behavior is different from `df.write`

core/src/test/scala/com/pingcap/tispark/auth/TiAuthIntegrationSuite.scala
shiyuhang0: Sadly, it will influence the test's result. 
tables will contain (f"[$databaseWithPrefix,$table,flase]") rather than [$databaseWithPrefix,$table]
`false`  corresponds to the new column `isTemporary` in spark 3.2

xuanyu66: Got it, so should we mention it in README?

shiyuhang0: it's the spark feature, I think we need not mention it for spark change.

xuanyu66 (approved):

birdstorm (approved):
LGTM

Issue #2290, opened 2022-04-10
[BUG] TiExpressionException: Cannot encode row key with non-long type


zzzzming95:
**Describe the bug**

create table with `BITINT UNSIGNED` primary key, when run sample SQL with range predicate, i will meet `TiExpressionException: Cannot encode row key with non-long type`. 

**What did you do**

first , create sample table in mysql client.
```
CREATE TABLE `test` (
  `id` BIGINT(20) UNSIGNED  NOT NULL AUTO_INCREMENT,
  `name` varchar(255) NOT NULL,
  `age` varchar(255) DEFAULT NULL,
  PRIMARY KEY (`id`) 
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
```

than , run sample sql with range predicate, than will meet the exception. like 

> ./bin/spark-sql --jars ~/tispark-assembly-2.4.3-scala_2.11.jar

> SELECT * FROM test.test WHERE id >= 0 LIMIT 10;

**What happens instead**

full stack:
```
com.pingcap.tikv.exception.TiExpressionException: Cannot encode row key with non-long type
    at com.pingcap.tikv.key.RowKey.toRowKey(RowKey.java:64)
    at com.pingcap.tikv.predicates.TiKVScanAnalyzer.buildTableScanKeyRangePerId(TiKVScanAnalyzer.java:321)
    at com.pingcap.tikv.predicates.TiKVScanAnalyzer.lambda$buildTableScanKeyRangeWithIds$0(TiKVScanAnalyzer.java:349)
    at java.lang.Iterable.forEach(Iterable.java:75)
    at com.pingcap.tikv.predicates.TiKVScanAnalyzer.buildTableScanKeyRangeWithIds(TiKVScanAnalyzer.java:347)
    at com.pingcap.tikv.predicates.TiKVScanAnalyzer.buildTableScanKeyRange(TiKVScanAnalyzer.java:376)
    at com.pingcap.tikv.predicates.TiKVScanAnalyzer.buildIndexScan(TiKVScanAnalyzer.java:273)
    at com.pingcap.tikv.predicates.TiKVScanAnalyzer.buildTableScan(TiKVScanAnalyzer.java:228)
    at com.pingcap.tikv.predicates.TiKVScanAnalyzer.buildTiDAGReq(TiKVScanAnalyzer.java:162)
[...]

PR #2291, opened 2022-04-10
fix TiExpressionException in TiKVScanAnalyzer

zzzzming95:

see here : https://github.com/pingcap/tispark/issues/2290

catch exception and run 

### Check List 

Tests 

```
CREATE TABLE `test` (
  `id` BIGINT(20) UNSIGNED  NOT NULL AUTO_INCREMENT,
  `name` varchar(255) NOT NULL,
  `age` varchar(255) DEFAULT NULL,
  PRIMARY KEY (`id`) 
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
```
> ./bin/spark-sql --jars ~/tispark-assembly-2.4.3-scala_2.11.jar

> SELECT * FROM test.test WHERE id >= 0 LIMIT 10;

no exception.

Side effects


zzzzming95:
cc @sunxiaoguang @xuanyu66 @shiyuhang0 

thanks

shiyuhang0:
@zzzzming95 thanks for your contribution! please add the test for your bug fix.

shiyuhang0:
@birdstorm cc is there a better way to fix the bug.
judge by exception is not a good way, maybe we can judge the type somewhere?

tikv-client/src/main/java/com/pingcap/tikv/predicates/TiKVScanAnalyzer.java
```diff
@@ -318,8 +319,12 @@ TiKVScanPlan buildIndexScan(
         // -INF
         startKey = RowKey.createMin(id);
       } else {
-        // Comparison with null should be filtered since it yields unknown always
-        startKey = RowKey.toRowKey(id, r.lowerEndpoint());
+        try {
+          // Comparison with null should be filtered since it yields unknown always
+          startKey = RowKey.toRowKey(id, r.lowerEndpoint());
+        }catch (TiExpressionException e){
+          startKey = RowKey.createMin(id);
```
shiyuhang0, line 326: just taking the min or max value will lead to an error answer.
here are the problems I found: insert a row with id = 9223372036854775808
1. `where id>0` can't get this row
2. `where id<0` get this row
3. `where id<=0` throw exception `Handle overflow for Long MAX` (maybe another bug)

birdstorm: This is due to another bug because unsigned long PK in TiDB actually has two separate ranges when it is stored.
Unsigned long PK created by TiDB is treated as int64 rather than uint64.
Thus, the PK data is stored in order from `[0, INT64MAX]`, and then from `[INT64_MIN, 0)`.
e.g., 
`id <= 0` should be interpreted as range `[0, 0], [INT64_MIN, 0)`

zzzzming95: @shiyuhang0  sorry for the late response...

as @birdstorm  describe , `RowKey` with Long.Min and Long.Max range can fetch all rowset . I'll add some UTs later. 

How to resolve other derived errors is subject to further discussion.

xuanyu66: https://github.com/pingcap/tidb/blob/master/distsql/request_builder.go#L447-L510
You can check how TIDB handles this situation.

shiyuhang0: I found this question involves not only this bug, I will work on it later.

zzzzming95:
@shiyuhang0 

UT has been added, Please review again. Thanks ~

xuanyu66 (approved):

xuanyu66:
@zzzzming95 plz check your format

xuanyu66:
@shiyuhang0 Can you confirm that tests are complete?

zzzzming95:
> @zzzzming95 plz check your format

Done, please rebuild CI.

tikv-client/src/test/java/com/pingcap/tikv/predicates/TiKVScanAnalyzerTest.java
```diff
@@ -408,4 +429,40 @@ private int[] getIndexLens() {
...
+    Coprocessor.KeyRange keyRange = keyRanges.get(table.getId()).get(0);
+
+    assertEquals(
+        ByteString.copyFrom(
+            new byte[] {116, -128, 0, 0, 0, 0, 0, 0, 6, 95, 114, 0, 0, 0, 0, 0, 0, 0, 1}),
+        keyRange.getStart());
+    assertEquals(
+        ByteString.copyFrom(
+            new byte[] {116, -128, 0, 0, 0, 0, 0, 0, 6, 95, 115, 0, 0, 0, 0, 0, 0, 0, 0}),
+        keyRange.getEnd());
```
shiyuhang0, line 466: I have a question about your test. Your predicate is 10>c1>0, but the range you get is not (0,10).
could you explain the test more clearly?

zzzzming95: @shiyuhang0 

This is mainly to fix `TiExpressionException: Cannot encode row key with non-long type` when create table with BITINT UNSIGNED primary key.

So the actual generation of startkey and endkey is here:

https://github.com/pingcap/tispark/blob/ba30d8c74cdd36edff26bdfb2a5d98c52f7491a4/tikv-client/src/main/java/com/pingcap/tikv/predicates/TiKVScanAnalyzer.java#L326

https://github.com/pingcap/tispark/blob/ba30d8c74cdd36edff26bdfb2a5d98c52f7491a4/tikv-client/src/main/java/com/pingcap/tikv/predicates/TiKVScanAnalyzer.java#L340

As discussed above, this fixes the issue
https://github.com/pingcap/tispark/pull/2291#discussion_r847199850

shiyuhang0: I am sorry that I don't pay much attention to this pr because I've been so busy the last two weeks. I am free now.
As for your solution:
It can work because TiSpark doesn't push down the filter with BIGINT UNSIGNED primary key, So Spark will do the filter instead. It will lead to the wrong answer if Tispark pushdown this filters to TiKV later.
I am going to check if it should be pushdown, if it can't be pushdown your solution can be accepted. otherwise, you should return the current range (start_key & end_key) to fetch the correct data from tikv

shiyuhang0:
Hi, thank you for your contribution, but I am sorry to say that your solution is not suitable. To fix the problems, there are two steps
1. fix the bug that unsigned long is not be pushdown 
2. fix the wrong range returned in  `buildTableScanKeyRangePerId ` (what you are doing in the pr)

The first step: I have found the bug, change `((BigDecimal) value).min(UNSIGNED_LONG_MAX).signum() > 0` to `((BigDecimal) value).subtract(UNSIGNED_LONG_MAX).signum() > 0` in Constant class
https://github.com/pingcap/tispark/blob/3f7382e7cf1062aedf98238015a251fd142df95b/tikv-client/src/main/java/com/pingcap/tikv/expression/Constant.java#L142 

The second step: your fix will return `(Long.MIN_VALUE,Long.MAX_VALUE)` for any filter, it is wrong when filter is pushdown to TiKV. you need to return the correct range. 

Ask us if you meet any problems!

xuanyu66:
closed due to https://github.com/pingcap/tispark/pull/2348 resolve this issue

PR #2309, opened 2022-04-24
MOD: move config out of dir workflows

xuanyu66:

Github would treat config.yml as a workflow and try to run it, which causes a notification `Some jobs were not successful`


shiyuhang0 (approved):

sunxiaoguang:
LGTM

Issue #2313, opened 2022-04-28
[BUG] some aggregations are not push down


shiyuhang0:
## sum does not push down

1. create a table and insert some rows
```
CREATE TABLE test  (
  `id` int(11) NOT NULL,
  PRIMARY KEY (`id`)
) 
```
2. execute `spark.sql("select sum(id) from tidb_catalog.dbname.test").explain`
```
*(2) HashAggregate(keys=[], functions=[sum(id#146L)])
+- Exchange SinglePartition, true, [id=#41]
   +- *(1) HashAggregate(keys=[], functions=[partial_sum(id#146L)])
      +- *(1) ColumnarToRow
         +- TiKV CoprocessorRDD{[table: t1] TableScan, Columns: id@LONG, KeyRange: [([t\200\000\000\000\000\000\000o_r\000\000\000\000\000\000\000\000], [t\200\000\000\000\000\000\000o_s\000\000\000\000\000\000\000\000])], startTs: 432836764553707521}
```
we can find sum is not push down to CoprocessorRDD

## count does not push down
because of this pr https://github.com/pingcap/tispark/pull/1840

## avg does not push down
avg is rewritten to sum and count, so we need to push down sum and count first.

## Version
TiSpark: 2.4 2.5
TiKV：The bug appeared in 2019.9(the time of this pr https://github.com/tikv/tikv/pull/5429) or before

PR #2314, opened 2022-04-28
fix aggregation not push down

shiyuhang0:
https://github.com/pingcap/tispark/issues/2313

### What is changed and how does it work?
1. Fix pushdown logical
    - fix the wrong pattern match in `PromotedSum.unapply` and `SumNotNullable.unapply`, so that sum and avg can be pushed down
    - fix `SumNotNullable.apply` to return the correct `SpecialSum`
2. Fix Request TiKV
    - push down the correct type for sum (for example: sum(bigint) needs decimal type). TiKV will check dataType after this pr https://github.com/tikv/tikv/pull/5429. 
(Reference [this TiDB code](https://github.com/pingcap/tidb/blob/2ef71689913081f99cae1ee1ce51a0e4a7341e13/expression/aggregation/base_func.go#L184)）
3. Fix return data type
    - fix `sumRewrite ` in `TiAggregationImpl` to cast decimal back to the long when using specialsum
4. fix `TiDAGRequest` so that explain will show the right info

## Results
- sum can pushdown
- count can't pushdown 
- avg can pushdown after count can pushdown


core/src/main/scala/org/apache/spark/sql/catalyst/planner/TiAggregationImpl.scala
```diff
@@ -84,15 +84,11 @@ object TiAggregationImpl {
...
         val sumRewrite = sumsRewriteMap.map {
           case (ref, sum) =>
-            val resultDataType = sum.resultAttribute.dataType
-            val castedSum = resultDataType match {
-              case LongType => Cast(sum.resultAttribute, DecimalType.BigIntDecimal)
-              case FloatType => Cast(sum.resultAttribute, DoubleType)
-              case DoubleType => Cast(sum.resultAttribute, DoubleType)
-              case d: DecimalType => Cast(sum.resultAttribute, d)
-            }
+            val castedSum = Cast(sum.resultAttribute, ref.dataType)
```
xuanyu66, line 91: Will it throw exception when value is overflow

core/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/SpecialSum.scala
```diff
@@ -31,37 +31,34 @@ import org.apache.spark.sql.types._
...
 
 object SumNotNullable {
   def apply(child: Expression): SpecialSum = {
-    val retType = child.dataType match {
-      case DecimalType.Fixed(precision, scale) =>
-        DecimalType.bounded(precision + 10, scale)
-      case _: IntegralType => LongType
-      case _ => DoubleType
-    }
+    // Use LongType because the push down count always return IntegerType.BIGINT
```
xuanyu66, line 53: always convert? since we convert it to bigint

core/src/main/scala/org/apache/spark/sql/catalyst/planner/TiAggregationImpl.scala
```diff
@@ -84,15 +84,11 @@ object TiAggregationImpl {
...
         }.toMap
 
+        // ResultDataType of sum only can be double or decimal
+        // We need to cast decimal back to long for the promotedSum, or spark will regard decimal as long cause the wrong answer
         val sumRewrite = sumsRewriteMap.map {
           case (ref, sum) =>
-            val resultDataType = sum.resultAttribute.dataType
-            val castedSum = resultDataType match {
-              case LongType => Cast(sum.resultAttribute, DecimalType.BigIntDecimal)
-              case FloatType => Cast(sum.resultAttribute, DoubleType)
```
xuanyu66, line 92: Should we keep the float case

shiyuhang0: the float type will not be pushed down. but I think your suggestion is better in case float  can pushdown in some day

core/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/SpecialSum.scala
shiyuhang0: I will change the confused name

core/src/main/scala/org/apache/spark/sql/catalyst/planner/TiAggregationImpl.scala
shiyuhang0: we can't control the behavior when overflow, it depends on the logic of spark cast operator.

shiyuhang0: And you can find the error results also happens in Spark JDBC rather than exception when overflow

xuanyu66 (approved):

PR #2316, opened 2022-05-05
add telemetry | collect usage information

Daemonxiao:
Add telemetry function.
When the telemetry collection feature is enabled, usage information will be shared, including(but not limited to):
* A randomly generated telemetry ID.
* OS and hardware information, such as OS name, OS version, CPU, memory, and disk.
* A part of TiSpark configuration.

### What is changed and how does it works?
The telemetry default is on.
And it can be closed in configuration, such as `spark.tispark.telemetry.enable = flase` .

Telemetry will work at injecting Spark rule in `com.pingcap.tispark.telemetry.TelemetryRule`. And each Spark application only sends a telemetry message once even if sending fails.

The collection entry will be sent by Restful API to the telemetry server.
Here is a simply telemetry-report example.
```scala
// build Telemetry
val telemetry = new Telemetry

// make TeleMsg
val teleMsg = new TeleMsg(sparkSession)

// report msg to server by HTTP post
telemetry.report(teleMsg)
```
TeleMsg entry would be transformed to JSON, such as
```JSON
{
  "track_id" : "trkid_7178a3d8-da26-41ca-a7a3-00e165877bda",
  "time" : 1652689510,
  "hardware" : {
    "os" : "Ubuntu",
    "disks" : [ {
      "name" : "/dev/nvme0n1",
      "size" : 512110190592
    } ],
    "version" : "20.04.4 LTS (Focal Fossa) build 5.13.0-41-generic",
    "cpu" : {
      "model" : "11th Gen Intel(R) Core(TM) i7-1160G7 @ 1.20GHz",
      "logicalCores" : 8,
      "physicalCores" : 4
    },
    "memory" : "Available: 1.7 GiB/15.4 GiB"
  },
  "instance" : {
    "tispark_version" : "2.5.0-SNAPSHOT",
    "tidb_version" : "v5.4.0",
    "spark_version" : "3.0.2"
  },
  "configuration" : {
    "spark.tispark.plan.use_index_scan_first" : false,
    "spark.tispark.index.scan_concurrency" : 5,
    "spark.tispark.type.unsupported_mysql_types" : "",
    "spark.tispark.plan.allow_index_read" : true,
    "spark.tispark.request.command.priority" : "LOW",
    "spark.tispark.request.isolation.level" : "SI",
    "spark.tispark.show_rowid" : false,
    "spark.tispark.plan.allow_agg_pushdown" : false,
    "spark.tispark.isolation_read_engines" : "tikv",
    "spark.sql.auth.enable" : false,
    "spark.tispark.coprocessor.chunk_batch_size" : 1024,
    "spark.tispark.index.scan_batch_size" : 20000,
    "spark.tispark.coprocess.codec_format" : "chblock",
    "spark.tispark.coprocess.streaming" : false
  }
}
```
Suggested review order：
1. TiExtensions.scala
2. TelemetryRule.scala
3. Telemetry.scala
4. TeleMsg.scala
5. all that remains

Tests 

 - Unit test

Side effects

 - Increased code complexity

Related changes


xuanyu66:
Plz list the exact information which would be collected in the description.

core/src/main/scala/com/pingcap/tispark/telemetry/TelemetryRule.scala
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * A telemetry rule injected into Spark.
+ * Telemetry message will be sent at here.
+ *
+ * @param sparkSession
+ */
+case class TelemetryRule(sparkSession: SparkSession) extends (LogicalPlan => Unit) {
+  private val logger = LoggerFactory.getLogger(getClass.getName)
+
+  if (TiExtensions.telemetryEnable(sparkSession)) {
```
shiyuhang0, line 32: telemetry should not affect SQL process. Maybe we can
1. catch the exception
2. use another thread

core/src/main/scala/com/pingcap/tispark/telemetry/TelemetryRule.scala
```diff
@@ -0,0 +1,42 @@
...
+    val telemetry = new Telemetry
+    val teleMsg = new TeleMsg(sparkSession)
+    if (teleMsg.shouldSendMsg) {
+      telemetry.report(teleMsg)
+      logger.info("Telemetry done")
+    }
+  }
+
+  override def apply(plan: LogicalPlan): Unit = plan
+}
```
shiyuhang0, line 42: why put telemetry logic in the constructor rather than apply method.

core/src/main/scala/org/apache/spark/sql/TiExtensions.scala
```diff
@@ -33,6 +34,7 @@ class TiExtensions extends (SparkSessionExtensions => Unit) {
 
     e.injectResolutionRule(new TiAuthRuleFactory(getOrCreateTiContext))
     e.injectPlannerStrategy(new TiStrategyFactory(getOrCreateTiContext))
+    e.injectCheckRule(TelemetryRule)
```
shiyuhang0, line 37: if you inject checkRule. The telemetry will do at the end of the analyzer. which means you can't execute telemetry once the analyzer fail such as `table or view not found` error.
if that's not your intention, maybe you should inject other stages

core/src/main/scala/org/apache/spark/sql/TiExtensions.scala
```diff
@@ -110,4 +112,16 @@ object TiExtensions {
...
+  /**
+   * check that telemetry is on
+   * default on
+   *
+   * @param sparkSession
+   * @return
+   */
+  def telemetryEnable(sparkSession: SparkSession): Boolean = {
+    sparkSession.sparkContext.conf
+      .get("spark.tispark.telemetry.enable", "true")
```
shiyuhang0, line 124: I think it is better to define a const whose value is `spark.tispark.telemetry.enable` rather than use it directly

core/src/main/scala/com/pingcap/tispark/telemetry/TelemetryRule.scala
Daemonxiao: Thanks. I want only telemetry sending once in one spark context. If it is in the apply(), telemetry will send in every SQL logic plan.

core/src/main/scala/com/pingcap/tispark/telemetry/TelemetryRule.scala
Daemonxiao: @shiyuhang0 Thanks for your review. I will add try-catch here. Telemetry takes a very short time, so maybe it no needs to restart the thread.

core/src/main/scala/org/apache/spark/sql/TiExtensions.scala
Daemonxiao: Telemetry should not send messages in every SQL analysis. When TelemetryRule is built-in spark context, the message will be sent in TelemetryRule's construct function.

core/src/main/scala/com/pingcap/tispark/telemetry/TelemetryRule.scala
```diff
@@ -0,0 +1,53 @@
...
+ */
+case class TelemetryRule(sparkSession: SparkSession) extends (LogicalPlan => Unit) {
+  private val logger = LoggerFactory.getLogger(getClass.getName)
+
+  val task: Runnable = () => {
+    try {
+      val telemetry = new Telemetry
+      val teleMsg = new TeleMsg(sparkSession)
+      if (teleMsg.shouldSendMsg) {
+        telemetry.report(teleMsg)
```
xuanyu66, line 38: There is a potential that many tasks would be able to send messages. You need concurrency utility such as Lock to avoid data race

core/src/main/scala/com/pingcap/tispark/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,57 @@
...
+    val httpClient = new HttpClientUtil
+
+    val mapper = new ObjectMapper()
+      .registerModule(DefaultScalaModule)
+      .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+    val msgString = mapper.writeValueAsString(msg)
+    logger.info("Telemetry report: " + msgString)
+
+    try {
+      httpClient.postJSON(url, msg)
```
xuanyu66, line 48: Do we need to add some retry in case of network fluctuations, and how huch is the connectTimeout and readTimeout

xuanyu66 (changes requested):

core/src/main/scala/com/pingcap/tispark/telemetry/Telemetry.scala
Daemonxiao: Default connectTimeout is 1 sec and readTimeout is 5 sec.

core/src/main/scala/com/pingcap/tispark/telemetry/TelemetryRule.scala
Daemonxiao: teleMsg is a local variable for each thread. Currently, telemetry would only be sent once when the rule is injected. So I deleted `shouldSendMsg` function to avoid confusion.

core/src/main/scala/org/apache/spark/sql/TiExtensions.scala
Daemonxiao: yep. I have used a const `TELEMETRY_ENABEL` to replace it

xuanyu66 (approved):

core/src/main/scala/com/pingcap/tispark/telemetry/TelemetryRule.scala
```diff
@@ -0,0 +1,56 @@
...
+  private val logger = LoggerFactory.getLogger(getClass.getName)
+
+  val task: Runnable = () => {
+    try {
+      val telemetry = new Telemetry
+      val teleMsg = new TeleMsg(sparkSession)
+      telemetry.report(teleMsg)
+    } catch {
+      case e: Throwable =>
+        logger.info("Failed to send telemetry message. " + e.getMessage)
```
shiyuhang0, line 40: warn will be better?

core/src/main/scala/com/pingcap/tispark/telemetry/TelemetryRule.scala
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * A telemetry rule injected into Spark.
+ * Telemetry message will be sent at here.
+ *
+ * @param sparkSession
+ */
+case class TelemetryRule(sparkSession: SparkSession) extends (LogicalPlan => Unit) {
+  private val logger = LoggerFactory.getLogger(getClass.getName)
+
+  val task: Runnable = () => {
```
shiyuhang0, line 33: use executor poll may be better

core/src/main/scala/com/pingcap/tispark/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,56 @@
...
+  private val logger = LoggerFactory.getLogger(getClass.getName)
+  private var url = "https://telemetry.pingcap.com/api/v1/tispark/report"
+
+  /**
+   * Send telemetry message.
+   *
+   * @param msg the msg sent to telemetry server
+   */
+  def report(msg: TeleMsg): Unit = {
+    val httpClient = new HttpClientUtil
```
shiyuhang0, line 37: will httpclient be closed automatically? If not, we should close it in finally

shiyuhang0:
ReadMe has been changed , fix the conflict. you can add the doc in wiki

core/src/main/scala/com/pingcap/tispark/telemetry/Telemetry.scala
Daemonxiao: yes, it would. Only httpserver needs to close actively.

core/src/main/scala/com/pingcap/tispark/telemetry/TelemetryRule.scala
Daemonxiao: In fact, only one telemetry would be reported in a spark application, So a single thread is enough.

shiyuhang0 (approved):

core/src/main/scala/com/pingcap/tispark/utils/HttpClientUtil.scala
```diff
@@ -0,0 +1,243 @@
...
+   * @param headers  HTTP header
+   * @return HTTP response string
+   */
+  def getHttps(
+      url: String,
+      conf: TiConfiguration,
+      headers: Map[String, String] = null): HttpResponse[String] = {
+    try {
+      val header: Map[String, String] =
+        if (headers != null) headers else Map("Content-Type" -> "application/json")
```
shiyuhang0, line 107: why application/json type?

core/src/main/scala/com/pingcap/tispark/utils/HttpClientUtil.scala
```diff
@@ -0,0 +1,243 @@
...
+      resp
+    } catch {
+      case e: Throwable =>
+        throw e
+    }
+  }
+
+  private def checkResp(url: String, resp: HttpResponse[String]): Unit = {
+    if (!resp.isSuccess) {
+      logger.info(
```
shiyuhang0, line 170: info may not sutiable

core/src/main/scala/com/pingcap/tispark/telemetry/TiSparkTeleInfo.scala
```diff
@@ -0,0 +1,157 @@
...
+        resp = httpClient.get(url)
+      }
+
+      val mapper = new ObjectMapper
+      val entry = mapper.readValue(resp.body, classTag[T].runtimeClass)
+
+      Option(entry.asInstanceOf[T])
+    } catch {
+      case e: Throwable =>
+        logger.info("Failed to get PD version " + e.getMessage)
```
shiyuhang0, line 110: it is better to judge if the resp is successful, and get the error message. 
or exception may be thrown in other places without root cause

core/src/main/scala/com/pingcap/tispark/telemetry/TeleMsg.scala
```diff
@@ -55,7 +56,8 @@ class TeleMsg(sparkSession: SparkSession) {
       putKeyValue(TRACK_ID, uuid, conf, tiSession)
       uuid
     } catch {
-      case _: Throwable =>
+      case e: Throwable =>
+        logger.info("Failed to generated telemetry track ID", e.getMessage)
```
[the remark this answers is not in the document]
Daemonxiao: Thanks @shiyuhang0 . I have changed it to warn.

core/src/main/scala/com/pingcap/tispark/utils/HttpClientUtil.scala
```diff
@@ -77,10 +88,156 @@ class HttpClientUtil {
...
+      val resp = Http(url)
+        .headers(header)
+        .option(HttpOptions.sslSocketFactory(sslContext.getSocketFactory))
+        .asString
+      checkResp(url, resp)
+
+      resp
+    } catch {
+      case e: Throwable =>
+        throw e
```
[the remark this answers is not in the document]
Daemonxiao: I add some logs here to help locate questions.

core/src/main/scala/com/pingcap/tispark/utils/HttpClientUtil.scala
Daemonxiao: I have changed it to warn.

core/src/main/scala/com/pingcap/tispark/telemetry/TiSparkTeleInfo.scala
Daemonxiao: yep, I do a pre-check, judging whether resp is successful or not, and add a warn log here when resp is failede.

core/src/main/scala/com/pingcap/tispark/utils/HttpClientUtil.scala
Daemonxiao: `application/json` type tells the server that the client comes with JSON content. However it is a GET request here, we don't need to add this head here. The `application/json` head would not affect in GET here.

core/src/main/scala/com/pingcap/tispark/telemetry/TiSparkTeleInfo.scala
shiyuhang0: If the check is not passed, it's better to stop the process.

zhangyangyu (approved):

xuanyu66 (approved):

core/src/test/scala/com/pingcap/tispark/tls/TiSparkTLSSuite.scala
```diff
@@ -88,4 +89,12 @@ class TiSparkTLSSuite extends BaseTiSparkTest {
     assert(2 == df.collect().head.get(0))
     assert("TiKV".equals(df.collect().head.get(1)))
   }
+
+  test("test get TiDB version with HTTPS") {
+    if (!TLSEnable) {
+      cancel
+    }
+    val tiSparkTeleInfo = TiSparkTeleInfo.getTiSparkTeleInfo()
```
xuanyu66, line 97: Could we test both jks and pem

Daemonxiao: yes, we have already test both jks and pem in Github ActionTLS test. [here](https://github.com/pingcap/tispark/runs/6665044299?check_suite_focus=true) show `test TLS` (pem cert) and `test JKS` (jks test).

xuanyu66: Does action include scalaj. If not, we should test it here or add to action.

core/src/main/scala/com/pingcap/tispark/utils/HttpClientUtil.scala
Daemonxiao: I delete default head ("Content-Type" -> "application/json") in GET.

Issue #2328, opened 2022-05-10
[BUG] Limit not push down when df.show


shiyuhang0:
** Version **
TiSpark: 2.5/master
Spark: any version
TiDB: any version

**Describe the bug**
spark.sql("select * from tidb_catalog.db.table limit 1").explain          limit will pushdown
spark.sql("select * from tidb_catalog.db.table limit 1").collect          limit will pushdown
spark.sql("select * from tidb_catalog.db.table limit 1").show            limit will not pushdown
spark.sql("select * from tidb_catalog.db.table limit 1").foreach(xxx) limit will not pushdown

Most of the actions will lead to this problem. (except collect and explain)

PR #2329, opened 2022-05-10
fix limit pushdown

shiyuhang0:
https://github.com/pingcap/tispark/issues/2328
### The cause of the problems
Dataframe with different actions may have different plans. for example 
```
plan of spark.sql("select * from tidb_catalog.db.table limit 1").show()
ReturnAnswer
+- GlobalLimit 2
   +- LocalLimit 2
      +- Project [cast(id#144L as string) AS id#147]
         +- GlobalLimit 1
            +- LocalLimit 1
               +- RelationV2[id#144L] test.t2

plan of spark.sql("select * from tidb_catalog.db.table limit 1").collect()
ReturnAnswer
+- GlobalLimit 1
   +- LocalLimit 1
      +- RelationV2[id#144L] test.t2
```
But TiSpark's push down of limit only matches some actions, such as `collect`. while limit + `show` is not be matched

### How to solve
we just need to match `logical.Limit(IntegerLiteral(limit), child)` in `TiStrategy`


shiyuhang0:
@birdstorm @xuanyu66 PTAL

birdstorm (approved):
LGTM

PR #2349, opened 2022-05-20
fix TLS and add JKS supporting

Daemonxiao:
1. Fix TLS: TLS test failure #2354 and TLS can't work when building TiSpark in jar #2356.
2. Add JKS support.
3. Fix TLS Github action #2350

#### Fix TLS
* Dependency conflicting makes TLS test failure.
To solve it, I have added a netty-handler dependency in tikv-client
```
        <dependency>
            <groupId>io.netty</groupId>
            <artifactId>netty-handler</artifactId>
            <version>4.1.48.Final</version>
        </dependency>
```
* ALPN lacking makes TLS can't work after shading io.netty.
It can be solved by affording some native dynamic libraries. Such as libnetty_tcnative_linux_x86_64.so.
But when shading io.netty, we should change lib names and add prefix, such as libshade_netty_tcnative_linux_x86.so.

#### Add JKS support.
Now you can use the JKS key store to build a TLS connection.
To enable JKS, you have to open it in configuration. such as,
```
spark.tispark.tikv.tls_enable=true
spark.tispark.tikv.jks_enable=true
spark.tispark.tikv.jks_key_path=/config/cert/client-keystore
spark.tispark.tikv.jks_key_password=123456
spark.tispark.tikv.jks_trust_path=/config/cert/server-cert-store
spark.tispark.tikv.jks_trust_password=12345678
```

When JKS and PEM cert are set simultaneously, JKS would have a higher priority.
That means TLS builder will use JKS cert first. 
So, do not set `spark.tispark.tikv.jks_enable=true` when you just want to use a common PEM cert.

#### Correct TLS test command 

Change Github action TLS test command: 

```
mvn test -am -pl core -Dtest=moo -DwildcardSuites=com.pingcap.tispark.tls.JDBCTLSSuite,com.pingcap.tispark.tls.TiKVClientTLSSuite,com.pingcap.tispark.tls.TiSparkTLSSuite -DfailIfNoTests=false
```

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

Related changes


xuanyu66:
Can you open issues describing what problem is TLS and TLS Github action

core/src/test/resources/tidb_config.properties.TLS.template
```diff
@@ -2,14 +2,19 @@ spark.tispark.pd.addresses=https://pd:2379
...
+
+#spark.tispark.tikv.jks_enable=true
+#spark.tispark.tikv.jks_key_path=/config/cert/client-keystore
+#spark.tispark.tikv.jks_key_password=123456
+#spark.tispark.tikv.jks_trust_path=/config/cert/server-cert-store
+#spark.tispark.tikv.jks_trust_password=12345678
 
 spark.tispark.jdbc.tls_enable=true
-spark.tispark.jdbc.server_cert_store = /config/cert/server-cert-store
+spark.tispark.jdbc.server_cert_store = file:/config/cert/server-cert-store
```
zhangyangyu, line 17: why this? does users also need to set `file` prefix?

tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
```diff
@@ -68,14 +68,26 @@ public class TiSession implements AutoCloseable {
 
   private TiSession(TiConfiguration conf) {
     this.conf = conf;
-    this.channelFactory =
-        conf.isTlsEnable()
-            ? new ChannelFactory(
+    if (conf.isTlsEnable()) {
```
zhangyangyu, line 71: what if tls is not enabled and jks is enabled?

core/src/test/scala/org/apache/spark/sql/test/SharedSQLContext.scala
```diff
@@ -348,7 +348,7 @@ trait SharedSQLContext
 
     var SSLPara = "false"
     if (conf.contains("jdbc.tls_enable") && conf.get("jdbc.tls_enable").equals("true")) {
-      SSLPara = "true&verifyServerCertificate=false&requireSSL=true"
+      SSLPara = "true&verifyServerCertificate=false&requireSSL=true&enabledTLSProtocols=TLSv1.2"
```
zhangyangyu, line 351: no 1.3?

core/src/test/resources/tidb_config.properties.TLS.template
Daemonxiao: The JDBC driver needs a URL to import the cert.
Yes, I will change the TLS configuration note in the wiki later.

tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
Daemonxiao: TLS enables is a prerequisite.
If tls is not enabled and jks is enabled, TLS will not be used.
To be consistent with the tikv-client TLS enable logic, I do this.

core/src/test/scala/org/apache/spark/sql/test/SharedSQLContext.scala
Daemonxiao: Thank you for your reminder. I will add it. 
And. It is a test, that does not affect really using.

.github/workflows/TLS-test.yml
```diff
@@ -20,12 +20,13 @@ jobs:
         uses: actions/setup-java@v2
         with:
           java-version: '8'
-          distribution: 'adopt'
+          distribution: 'zulu'
```
xuanyu66, line 23: Have you tried all major distributions?

core/src/test/resources/tidb_config.properties.TLS.template
xuanyu66: We can check the file and automatically add `file` prefix

xuanyu66 (changes requested):

.github/workflows/TLS-test.yml
Daemonxiao: well, I have tried OpenJDK, Oracle JDK, and Zulu.

shiyuhang0:
Doc need be  updated too.

.github/workflows/TLS-test.yml
```diff
@@ -34,5 +39,11 @@ jobs:
       - name: build
         run: mvn clean package -Dmaven.test.skip=true -B
 
-      - name: test
-        run: mvn test -Dtest=moo -DwildcardSuites=JDBCTLSSuite,TiKVClientTLSSuite,TiSparkTLSSuite -DfailIfNoTests=false
\ No newline at end of file
+      - name: test TLS
+        run: mvn test -am -pl core -Dtest=moo -DwildcardSuites=com.pingcap.tispark.tls.JDBCTLSSuite,com.pingcap.tispark.tls.TiKVClientTLSSuite,com.pingcap.tispark.tls.TiSparkTLSSuite -DfailIfNoTests=false
```
zhangyangyu, line 43: maybe `failIfNoTests=true` is better?

Daemonxiao: I'm afraid not.
Because there is no running test in other models, such as tikv-client, you will get the error "_No tests were executed!  (Set -DfailIfNoTests=false to ignore this error._" when setting `failIfNoTests=true`

Daemonxiao:
@shiyuhang0 You are right. I have tested Spark-SQL with TiSpark-jar in spark 3.0, 3.1, and 3.2.

core/src/test/resources/tidb_config.properties.TLS.template
Daemonxiao: @xuanyu66 Thanks. I have checked in the code. Now we can use both `file:/confgiXXX` and `/configXXX`.

xuanyu66 (approved):

shiyuhang0 (approved):

PR #2353, opened 2022-05-24
fix implicit narrowing conversion in compound assignment

ajian2002:
### What has changed and how does it work?

Change the data type of the field to prevent overflow
### Check List 

Tests 

   - unit test
   - Integration Testing
   - no code

code changes
   - Exported variable/field changes

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/codec/CodecDataInput.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/codec/CodecDataInput.java
@@ -281,19 +281,19 @@ public class CodecDataInput implements DataInput {
    */
   private static class UnSyncByteArrayInputStream extends InputStream {
     protected byte[] buf;
-    protected int pos;
+    protected long pos;
     protected int mark = 0;
     protected int count;
 
     UnSyncByteArrayInputStream(byte[] buf) {
       this.buf = buf;
-      this.pos = 0;
+      this.pos = 0L;
       this.count = buf.length;
     }
 
     @Override
     public int read() {
-      return (pos < count) ? (buf[pos++] & 0xff) : -1;
+      return (pos < count) ? (buf[Math.toIntExact(pos++)] & 0xff) : -1;
     }
 
     @Override
@@ -308,14 +308,14 @@ public class CodecDataInput implements DataInput {
         return -1;
       }
 
-      int avail = count - pos;
+      int avail = Math.toIntExact(count - pos);
       if (len > avail) {
         len = avail;
       }
       if (len <= 0) {
         return 0;
       }
-      System.arraycopy(buf, pos, b, off, len);
+      System.arraycopy(buf, Math.toIntExact(pos), b, off, len);
       pos += len;
       return len;
     }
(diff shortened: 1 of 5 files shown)


tikv-client/src/main/java/com/pingcap/tikv/codec/CodecDataInput.java
```diff
@@ -281,19 +281,19 @@ public byte[] toByteArray() {
    */
   private static class UnSyncByteArrayInputStream extends InputStream {
     protected byte[] buf;
-    protected int pos;
+    protected long pos;
```
xuanyu66, line 284: why change to long

ajian2002: > why change to long

yes ,we can still use int, let me change it

xuanyu66:
@shiyuhang0 PTAL

tikv-client/src/main/java/com/pingcap/tikv/codec/CodecDataInput.java
```diff
@@ -327,7 +327,7 @@ public long skip(long n) {
         k = n < 0 ? 0 : n;
       }
 
-      pos += k;
+      pos = Math.toIntExact(pos + k);
```
shiyuhang0, line 330: pos + k will not overflow for it always < count which is int type

shiyuhang0:
if you can reproduce any bugs because of the overflow, add them to the issue and add the tests.
if you find the potential risks of overflow by some tools, you can explain what tools are used in the pr.

tikv-client/src/main/java/com/pingcap/tikv/codec/CodecDataInput.java
ajian2002: it can only be changed to type strong transfer

shiyuhang0: you can set k to int type, and `k = n < 0 ? 0 : (int)n`

zhangyangyu (approved):

tikv-client/src/main/java/com/pingcap/tikv/codec/CodecDataInput.java
```diff
@@ -327,7 +327,7 @@ public long skip(long n) {
         k = n < 0 ? 0 : n;
       }
 
-      pos += k;
+      pos = (int) (pos + k);
```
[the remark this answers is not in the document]
ajian2002: https://github.com/pingcap/tispark/pull/2353#discussion_r880450872

shiyuhang0 (approved):

tikv-client/src/main/java/com/pingcap/tikv/codec/CodecDataInput.java
```diff
@@ -322,9 +322,9 @@ public int read(byte[] b, int off, int len) {
 
     @Override
     public long skip(long n) {
-      long k = count - pos;
+      int k = count - pos;
```
xuanyu66, line 325: why change to int

xuanyu66 (approved):

PR #2362, opened 2022-05-26
use lombok to clean TiConfiguratuin code

ajian2002:
use lombok to clean TiConfiguratuin code
### Check List 

Tests 
 - Integration test

Code changes

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/TiConfiguration.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/TiConfiguration.java
@@ -27,10 +27,16 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.experimental.Accessors;
 import org.joda.time.DateTimeZone;
 import org.tikv.kvproto.Kvrpcpb.CommandPri;
 import org.tikv.kvproto.Kvrpcpb.IsolationLevel;
 
+@Getter
+@Setter
+@Accessors(chain = true)
 public class TiConfiguration implements Serializable {
   private static final DateTimeZone DEF_TIMEZONE = Converter.getLocalTimezone();
   private static final int DEF_TIMEOUT = 10;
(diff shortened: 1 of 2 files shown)


tikv-client/src/main/java/com/pingcap/tikv/TiConfiguration.java
```diff
@@ -83,7 +89,7 @@ public class TiConfiguration implements Serializable {
   private IsolationLevel isolationLevel = DEF_ISOLATION_LEVEL;
   private int maxRequestKeyRangeSize = MAX_REQUEST_KEY_RANGE_SIZE;
   private boolean showRowId = DEF_SHOW_ROWID;
-  private String dbPrefix = DEF_DB_PREFIX;
+  private String DBPrefix = DEF_DB_PREFIX;
```
xuanyu66, line 92: why do this

xuanyu66 (changes requested):

xuanyu66 (approved):

tikv-client/src/main/java/com/pingcap/tikv/TiConfiguration.java
```diff
@@ -167,278 +151,10 @@ public int getScanBatchSize() {
...
-    return this;
-  }
-
-  public int getMaxFrameSize() {
-    return maxFrameSize;
-  }
-
-  public TiConfiguration setMaxFrameSize(int maxFrameSize) {
-    this.maxFrameSize = maxFrameSize;
-    return this;
```
shiyuhang0, line 194: does lombok will return TiConfiguration in setMaxFrameSize? if not, we need to see if this change will affect other places.
plz check other code with the same problems in this pr

xuanyu66: https://projectlombok.org/features/experimental/Accessors
@Accessor(chain = true) will do this

ajian2002: @Accessors(chain = true) can change setter method return value ,more info you can see [lombok/Accessors](https://projectlombok.org/features/experimental/Accessors)

shiyuhang0 (approved):

PR #2363, opened 2022-05-27
change name url to uri

ajian2002:
The methods of the PDUtils class are different from the methods of the PDUtils class in client-java

change name addrToUrl to addrToUri 

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes


ajian2002:
cc  @xuanyu66  @shiyuhang0 
thanks

xuanyu66:
You need to figure out why there is a difference and whether we need write some unit tests

PR #2365, opened 2022-05-29
Add TLS reload

Daemonxiao:
- Support TiKV client TLS reloading: A backport from [#597](https://github.com/tikv/client-java/pull/597) [#589](https://github.com/tikv/client-java/pull/589)

#### TiKV client TLS reloading
Auto reloading certificates. There are two configurations to control this work.

* `spark.tispark.tikv.tls_reload_interval` : The interval time of checking if there is any reloading certificates. If your certificate needs to be updated frequently, you can actively set this parameter. The default is `10s` (10 seconds).
*  `spark.tispark.tikv.conn_recycle_time`: The interval time of cleaning the expired connection with TiKV. Only take effect when enabling cert reloading. The default is `60s` (60 seconds).

```
spark.tispark.tikv.tls_reload_interval = 10s
spark.tispark.tikv.conn_recycle_time = 60s
```

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Related changes


core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
```diff
@@ -240,6 +248,16 @@ object TiUtil {
           conf.setKeyCertChainFile(sqlConf.getConfString(TiConfigConst.TIKV_KEY_CERT_CHAIN))
           conf.setKeyFile(sqlConf.getConfString(TiConfigConst.TIKV_KEY_FILE))
         }
+
```
shiyuhang0, line 251: can  we use sparkConfToTiConfWithoutPD to create TiConf in TiCatalog? so that we needn't modify two places

shiyuhang0: TiContext is inited before TiCatalog, and TiSession will be cached  after the init of  TiContext. Thus, TiCatalog just need pd address to get TiSession which is cached.
As for the TLS config, I think it has been added into the configuration when TiSession is created in TiContext. injectTLSParam is redundant.

core/src/main/scala/com/pingcap/tispark/write/TiDBWriter.scala
```diff
@@ -32,27 +34,27 @@ object TiDBWriter {
...
-            throw new TiBatchWriteException(
-              s"table `${options.database}`.`${options.table}` does not exists!")
-            // TiDBUtils.createTable(conn, df, options, tiContext)
-            // TiDBUtils.saveTable(tiContext, df, Some(df.schema), options)
+        var tableExists: Boolean = false
+        val tiSession: TiSession = tiContext.tiSession
+        val catalog: Catalog = tiSession.getCatalog
+        val table: TiTableInfo = catalog.getTable(options.database, options.table)
+        if (table != null)
+          tableExists = true
```
shiyuhang0, line 42:
```suggestion
          tableExists = true
```
val tableExists = (table!=null)

core/src/main/scala/com/pingcap/tispark/write/TiDBWriter.scala
```diff
@@ -32,27 +34,27 @@ object TiDBWriter {
...
-              case _ =>
-                throw new TiBatchWriteException(
-                  s"SaveMode: $saveMode is not supported. TiSpark only support SaveMode.Append.")
-            }
-          } else {
-            throw new TiBatchWriteException(
-              s"table `${options.database}`.`${options.table}` does not exists!")
-            // TiDBUtils.createTable(conn, df, options, tiContext)
-            // TiDBUtils.saveTable(tiContext, df, Some(df.schema), options)
+        var tableExists: Boolean = false
```
xuanyu66, line 37:
```suggestion
       val catalog = tiContext.tiSession.getCatalog
        val tableExists = catalog.getTable(options.database, options.table) == null ? false : true
```

xuanyu66: make it cleaner

tikv-client/src/main/java/com/pingcap/tikv/jdbc/TiDBJDBCClient.java
```diff
@@ -157,8 +173,17 @@ public void updateTableStatistics(long startTS, long tableId, long delta, long c
     }
   }
 
+  public void updateTableStatisticsWithRetry(long startTS, long tableId, long delta, long count)
+      throws SQLException {
+    executeWithUpdateConnRetry(
+        () -> {
+          updateTableStatistics(startTS, tableId, delta, count);
+          return null;
```
shiyuhang0, line 181: why return null

tikv-client/src/main/java/com/pingcap/tikv/jdbc/AbstractJDBCClient.java
```diff
@@ -0,0 +1,104 @@
...
+    return retryPolicy.executeWithRetry(
+        () -> {
+          if (connection == null || !connection.isValid(1)) {
+            this.updateConnection();
+          }
+          return method.call();
+        });
+  }
+
+  public <RespT> RespT executeWithRetry(
```
shiyuhang0, line 85: executeWithUpdateConnRetry may better?

tikv-client/src/main/java/com/pingcap/tikv/PDClient.java
```diff
@@ -347,22 +347,33 @@ LeaderWrapper getLeaderWrapper() {
...
-      return resp;
+    }
+  }
+
+  private GetMembersResponse getMembers(URI uri) {
+    BackOffer backOffer = ConcreteBackOffer.newCustomBackOff(BackOffer.PD_INFO_BACKOFF);
+    try {
+      return doGetMembers(backOffer, uri);
     } catch (Exception e) {
-      logger.warn("failed to get member from pd server.", e);
```
xuanyu66, line 363: Why swallow the exception

tikv-client/src/main/java/com/pingcap/tikv/policy/JDBCRetryPolicy.java
```diff
@@ -0,0 +1,60 @@
...
+        result = proc.call();
+      } catch (CommunicationsException e) {
+        boolean retry = handler.handleRequestError(backOffer, e);
+        if (retry) {
+          continue;
+        }
+      } catch (SQLException e) {
+        throw e;
+      } catch (Exception e) {
+        throw new RuntimeException(e);
```
shiyuhang0, line 54: why change Exception to RuntimeException

tikv-client/src/main/java/com/pingcap/tikv/jdbc/AbstractJDBCClient.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  public AbstractJDBCClient(Connection connection, String jdbcUrl) {
+    this.connection = connection;
+    this.jdbcUrl = jdbcUrl;
+  }
+
+  @SneakyThrows
+  public synchronized void updateConnection() {
+    if (connection != null && connection.isValid(1)) return;
```
shiyuhang0, line 49: magic number, define a static final int to replace 1

xuanyu66 (changes requested):

tikv-client/src/main/java/com/pingcap/tikv/jdbc/AbstractJDBCClient.java
```diff
@@ -0,0 +1,104 @@
...
+import com.pingcap.tikv.util.BackOffer;
+import com.pingcap.tikv.util.ConcreteBackOffer;
+import java.sql.Connection;
+import java.sql.Driver;
+import java.sql.SQLException;
+import java.util.Properties;
+import java.util.concurrent.Callable;
+import lombok.SneakyThrows;
+
+public class AbstractJDBCClient {
```
xuanyu66, line 31: It's weird to name a un abstract class start with AbstarcXXX

tikv-client/src/main/java/com/pingcap/tikv/policy/JDBCRetryPolicy.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  public <RespT> RespT executeWithRetry(Callable<RespT> proc) throws SQLException {
+    RespT result = null;
+    while (true) {
+      try {
+        result = proc.call();
+      } catch (CommunicationsException e) {
+        boolean retry = handler.handleRequestError(backOffer, e);
+        if (retry) {
```
xuanyu66, line 48: move it to While(retry)

xuanyu66 (changes requested):

shiyuhang0 (commented):
1

tikv-client/src/main/java/com/pingcap/tikv/jdbc/AbstractJDBCClient.java
Daemonxiao: I will change it to an abstract class.

tikv-client/src/main/java/com/pingcap/tikv/PDClient.java
Daemonxiao: Shall we print an exception stack here?
Now, tikv-client is `return null` here and null will be checked at the call.

core/src/main/scala/com/pingcap/tispark/write/TiDBWriter.scala
Daemonxiao: OK. Thanks for your advice.

tikv-client/src/main/java/com/pingcap/tikv/jdbc/TiDBJDBCClient.java
Daemonxiao: Maybe there is a better way to solve it, but I have no idea now.
"why return null" here is due to `void` can not be used for generics, so `return null` is used instead of Void.

tikv-client/src/main/java/com/pingcap/tikv/policy/JDBCRetryPolicy.java
Daemonxiao: If we do that, the result may be null, and while(true) is not useful.

tikv-client/src/main/java/com/pingcap/tikv/policy/JDBCRetryPolicy.java
Daemonxiao: proc.call() should not throw other exceptions except SQLException. And When other exceptions happen, we should throw them directly as runtime exceptions without any handle at the call, such as try-catch.

tikv-client/src/main/java/com/pingcap/tikv/policy/JDBCRetryPolicy.java
xuanyu66: change while(true) to While(retry) and remove code here

tikv-client/src/main/java/com/pingcap/tikv/jdbc/AbstractJDBCClient.java
Daemonxiao: Sorry, It's my fault. I have deleted updateConnection() in  executeWithRetry().

core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
Daemonxiao: @shiyuhang0 Great. You are right. I have deleted the duplicate, redundant, unuseful function injectTLSParam().

core/src/main/scala/com/pingcap/tispark/write/TiDBWriter.scala
```diff
@@ -32,28 +32,26 @@ object TiDBWriter {
...
-                  s"SaveMode: $saveMode is not supported. TiSpark only support SaveMode.Append.")
-            }
-          } else {
-            throw new TiBatchWriteException(
-              s"table `${options.database}`.`${options.table}` does not exists!")
-            // TiDBUtils.createTable(conn, df, options, tiContext)
-            // TiDBUtils.saveTable(tiContext, df, Some(df.schema), options)
+        val catalog = tiContext.tiSession.getCatalog
+        val tableExists =
+          if (catalog.getTable(options.database, options.table) == null) false else true
```
shiyuhang0, line 37: `val tableExists =  catalog.getTable(options.database, options.table) != null ` will be simper

zhangyangyu (approved):

core/src/main/scala/com/pingcap/tispark/write/TiDBWriter.scala
Daemonxiao: Thanks, I have changed it to `val catalog = tiContext.tiSession.getCatalog.getTable(options.database, options.table) != null`.

tikv-client/src/main/java/com/pingcap/tikv/PDClient.java
xuanyu66: Yes

shiyuhang0 (approved):

xuanyu66 (approved):

PR #2370, opened 2022-06-08
new naming rule for tispark's jar

shiyuhang0:
This pr introduces a new naming rule for TiSpsark jar.
`tispark-assembly` will change to `tispark-assembly-${spark_version}_${scala_version}`
`spark-wrapper-spark-3.x` will change to `spark-wrapper-spark-3.x_${scala_version}`

- `${spark_version}`: use the first two numbers of spark version, for example, 3.0 for spark version 3.0.x
- `${scala_version}`: use the first two numbers of scala version, for example, 2.12 for scala version of 2.12.x

in this way we can
- Solve most of the compatibility problems.
- Help user select version easier

### DIFF
--- a/spark-wrapper/spark-3.0/pom.xml
+++ b/spark-wrapper/spark-3.0/pom.xml
@@ -9,7 +9,7 @@
         <relativePath>../../pom.xml</relativePath>
     </parent>
 
-    <artifactId>spark-wrapper-spark-3.0</artifactId>
+    <artifactId>spark-wrapper-spark-3.0_${spark.version.release}</artifactId>
     <packaging>jar</packaging>
     <name>TiSpark Project Spark Wrapper Spark-3.0</name>
     <url>http://github.copm/pingcap/tispark</url>
--- a/spark-wrapper/spark-3.0/pom.xml
+++ b/spark-wrapper/spark-3.0/pom.xml
@@ -9,7 +9,7 @@
         <relativePath>../../pom.xml</relativePath>
     </parent>
 
-    <artifactId>spark-wrapper-spark-3.0_${spark.version.release}</artifactId>
+    <artifactId>spark-wrapper-spark-3.0_${scala.version.release}</artifactId>
     <packaging>jar</packaging>
     <name>TiSpark Project Spark Wrapper Spark-3.0</name>
     <url>http://github.copm/pingcap/tispark</url>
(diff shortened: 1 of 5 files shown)


spark-wrapper/spark-3.0/pom.xml
```diff
@@ -9,7 +9,7 @@
         <relativePath>../../pom.xml</relativePath>
     </parent>
 
-    <artifactId>spark-wrapper-spark-3.0</artifactId>
+    <artifactId>spark-wrapper-spark-3.0_${spark.version.release}</artifactId>
```
Daemonxiao, line 12: Is here ${scala.version}?
Because you said "spark-wrapper-spark-3.x will change to spark-wrapper-spark-3.x_${scala_version}".

xuanyu66: +1

spark-wrapper/spark-3.1/pom.xml
```diff
@@ -9,7 +9,7 @@
         <relativePath>../../pom.xml</relativePath>
     </parent>
 
-    <artifactId>spark-wrapper-spark-3.1</artifactId>
+    <artifactId>spark-wrapper-spark-3.1_${spark.version.release}</artifactId>
```
xuanyu66, line 12: +1

spark-wrapper/spark-3.2/pom.xml
```diff
@@ -9,7 +9,7 @@
         <relativePath>../../pom.xml</relativePath>
     </parent>
 
-    <artifactId>spark-wrapper-spark-3.2</artifactId>
+    <artifactId>spark-wrapper-spark-3.2_${spark.version.release}</artifactId>
```
xuanyu66, line 12: +1

xuanyu66 (changes requested):

shiyuhang0:
@birdstorm PTAL

spark-wrapper/spark-3.0/pom.xml
shiyuhang0: typo,sorry

Daemonxiao (approved):

birdstorm (approved):
LGTM

xuanyu66:
should combine spark.version.release, spark.version.compile, spark.version.test into spark.version?
Since there is no point when the three versions are different.

shiyuhang0:
> should combine spark.version.release, spark.version.compile, spark.version.test into spark.version? Since there is no point when the three versions are different.

I think we can combine spark.version.compile and spark.version.test.
But spark.version.release should not, because spark.version.release just takes the first two numbers of the spark version.

PR #2373, opened 2022-06-12
[TEST]Remove tikv-client

ajian2002:
what did I do?
1. I hold the upstream pdclient instance in the PDClient's constructor
2. In order to hold the upstream pdclient instance, I do some conversions between TiConfiguration and ChannelFactory.
3. I replaced the getTimestampe method implemented by PDClient with upstream pdclient. But I can't solve the compatibility problem of BackOffer. I simply found that the existing implementation logic of getTimestampe according to the call stack does not require the BackOffer parameter, but a ConcreteBackOffer.newCopNextMaxBackOff() in com/pingcap/tikv/policy/RetryPolicy. in java.


Daemonxiao:
I am afraid you could not "hold the upstream pdclient instance in the PDClient's constructor". Because there are different `io.netty.xxx` versions between tispark-tikv-client and upstream-tikv-client, the dependence conflict will happen when you import upstream-tikv-client directly in pom.
To remove the whole tispark-tikv-client, maybe you should remove them together and delete the duplicate dependence in pom.

PR #2374, opened 2022-06-12
Change the  fixed tidb catalog name to arbitrary name with the "spark.sql.catalog." prefix

wfxxh:
change fiedx tidb_catalog_name to dynamic name prefix with "spark.sql.catalog."

org.apache.spark.sql.catalyst.catalog.TiCatalog
org.apache.spark.sql.catalyst.parser.TiParser
org.apache.spark.sql.TiExtensions

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/TiExtensions.scala
+++ b/core/src/main/scala/org/apache/spark/sql/TiExtensions.scala
@@ -68,18 +68,22 @@ object TiExtensions {
    * Catalog for tidb is necessary now.
    * @param sparkSession
    */
-  def validateCatalog(sparkSession: SparkSession): String = {
-    sparkSession.sparkContext.conf
-      .getAllWithPrefix("spark.sql.catalog.")
-      .toSeq
-      .find(pair => TiCatalog.className.equals(pair._2)) match {
+  def validateCatalog(sparkSession: SparkSession): Unit = {
+    getCatalogConf(sparkSession) match {
       case None =>
         logger.error("TiSpark must work with TiCatalog. Please add TiCatalog in spark conf.")
         throw new TiInternalException(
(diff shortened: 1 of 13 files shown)


xuanyu66:
Can you explain the propose of this PR

wfxxh:
image
the catalog name can be any not be the fixed "tidb_catalog"

xuanyu66:
@wfxxh Can you fix the failed tests and update the base branch

xuanyu66:
> image the catalog name can be any not be the fixed "tidb_catalog"

@wfxxh Why should we change the catalog name？ Can you explain the purpose

wfxxh:
@xuanyu66  @shiyuhang0  when i use iceberg ,the catalog name can be any string, so I think tispark  catalog name shoule not be  the fixed string

core/src/main/scala/org/apache/spark/sql/TiExtensions.scala
```diff
@@ -68,7 +68,7 @@ object TiExtensions {
    * Catalog for tidb is necessary now.
    * @param sparkSession
    */
-  def validateCatalog(sparkSession: SparkSession): Unit = {
+  def validateCatalog(sparkSession: SparkSession): String = {
```
shiyuhang0, line 71: the name should not be the `validateCatalog` for it can also return the catalog name

shiyuhang0:
add test for your change

xuanyu66:
@wfxxh use `mvn com.coveo:fmt-maven-plugin:format` and `mvn mvn-scalafmt_2.12:format -Dscalafmt.skip=false -B`
and also you should update the branch

xuanyu66:
@wfxxh Why close this PR?

wfxxh:
@xuanyu66 Sorry, I think this PR is not useful for every one.

PR #2399, opened 2022-06-21
[TEST]Use client-java/PDClient to replace  tikv-client/PdClient .(ReadOnlyPDClient interface)

ajian2002:
- After the `tikv-client` module introduces the `client-java` dependency, the current `ReadonlyPDClient` interface inherits the upstream interface, and the current `PDClient` holds the upstream `PDClient` instance.

- Use the `shade` plugin of `maven` to `shade` the upstream dependencies of `io.netty`, `com.google`, etc., thus solving the version conflict problem of some libraries.

- [protobuf](https://github.com/pingcap/tispark/pull/2399/files#diff-42d1c8d6beec4539985d8a1587b43da2cef4ef079cb7cfbc8945e9b7c3f40cf4R207:~:text=16,scripts/proto.sh), removed references to `kvproto` and `raft_rs ` to use the upstream implementation.

- [`TiRegion`, `TiStoreType`, `TITimestamp` use upstream, `Metapb.Store` is replaced by `TiStore`

- abstract class AbstractRegionStoreClient
    extends AbstractGRPCClient<
        TikvGrpc.TikvFutureStub, TikvGrpc.TikvBlockingStub, TikvGrpc.TikvStub> inherits FutureStub and is not used yet.

- Use upstream `regionCache`, `RegionManager` uses upstream `regionCache` to implement behavior.

- BackOffer extends org.tikv.common.util.BackOffer
- ConcreteBackOffer implements BackOffer

- use upstream `FutureObserver`
-
- region.getContext is deprecated, and the calling method needs to be changed.

- The coding method is currently quite confusing, so I will test it first and then run it.
  
- The parameters passed when new TiStroe are not clear, let's run the test first.

- Added support to [healthCheck](https://github.com/pingcap/tispark/pull/2399/files#diff-1e1afcfbb6ac1fb40147265f456f43daae1e925b5d4bd40fb47814d508dbcedb:~:text=private%20static%20class%20HealBase%20extends%20%HealthCheck%20extends%20%HealthCheck%207B) Support

TODO:

- [de9a0f](https://github.com/pingcap/tispark/pull/2399/commits/de9a0f81c0d70aa0e1ef8f9d49af33b8c99187b9) Replace Stub with upstream FutureStub in AbstractGRPCClient or directly use upstream Abstract, etc.
- Replace exception with upstream exception.


core/src/main/scala/com/pingcap/tispark/utils/TiUtil.scala
```diff
@@ -29,6 +29,8 @@ import org.apache.spark.sql.types.{MetadataBuilder, StructField, StructType}
 import org.apache.spark.sql.{DataFrame, Row, SQLContext, SparkSession}
 import org.apache.spark.{SparkConf, sql}
 import org.slf4j.LoggerFactory
+import org.tikv.common.meta
+import org.tikv.common.meta.TiTimestamp
```
shiyuhang0, line 33: you have imported meta, why import meta.TiTimestamp

tikv-client/src/main/java/com/pingcap/tikv/util/ConcreteBackOffer.java
```diff
@@ -153,4 +182,36 @@ public void doBackOffWithMaxSleep(
...
+  public SlowLog getSlowLog() {
+    return upstreamConcreteBackOffer.getSlowLog();
+  }
+
+  @Override
+  public Long getClusterId() {
+    // https://github.com/tikv/client-java/pull/558
+    // TODO: to implement
+    return 0L;
+  }
```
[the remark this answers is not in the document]
ajian2002: Returning 0 will have no effect. Returning null will destroy unit tests and integration tests. This is an interface newly added upstream, and the current project has not used it yet.

tikv-client/src/main/java/com/pingcap/tikv/util/BackOffer.java
```diff
@@ -18,48 +18,32 @@
...
   int GC_DELETE_RANGE_MAX_BACKOFF = 100 * seconds;
-  int RAWKV_MAX_BACKOFF = 40 * seconds;
-  int SPLIT_REGION_BACKOFF = 20 * seconds;
   int BATCH_COMMIT_BACKOFF = 10 * seconds;
-  int PD_INFO_BACKOFF = 5 * seconds;
   int ROW_ID_ALLOCATOR_BACKOFF = 40 * seconds;
 
+  //  TODO:different from org.tikv.common.util.BackOffer default value
+  int RAWKV_MAX_BACKOFF = 40 * seconds;
+  int SPLIT_REGION_BACKOFF = 20 * seconds;
```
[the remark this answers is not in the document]
ajian2002: Currently different from the upstream default. I didn't modify it in order to keep the behavior unchanged. Does it mean that the upstream SPLIT_REGION_BACKOFF can be used?

tikv-client/src/main/java/com/pingcap/tikv/util/BackOffer.java
```diff
@@ -18,48 +18,32 @@
...
   int GC_RESOLVE_LOCK_MAX_BACKOFF = 100 * seconds;
   int GC_DELETE_RANGE_MAX_BACKOFF = 100 * seconds;
-  int RAWKV_MAX_BACKOFF = 40 * seconds;
-  int SPLIT_REGION_BACKOFF = 20 * seconds;
   int BATCH_COMMIT_BACKOFF = 10 * seconds;
-  int PD_INFO_BACKOFF = 5 * seconds;
   int ROW_ID_ALLOCATOR_BACKOFF = 40 * seconds;
 
+  //  TODO:different from org.tikv.common.util.BackOffer default value
+  int RAWKV_MAX_BACKOFF = 40 * seconds;
```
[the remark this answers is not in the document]
ajian2002: ok,listen to you

tikv-client/src/main/java/com/pingcap/tikv/util/ConcreteBackOffer.java
```diff
@@ -35,46 +37,72 @@ public class ConcreteBackOffer implements BackOffer {
...
+
   public static ConcreteBackOffer newCustomBackOff(int maxSleep) {
-    return new ConcreteBackOffer(maxSleep);
+    return newCustomBackOff(maxSleep, 0);
   }
 
   public static ConcreteBackOffer newScannerNextMaxBackOff() {
-    return new ConcreteBackOffer(SCANNER_NEXT_MAX_BACKOFF);
+    return new ConcreteBackOffer(BackOffer.SCANNER_NEXT_MAX_BACKOFF, 0);
```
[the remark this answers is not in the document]
ajian2002: I will restore it.

tikv-client/src/main/java/com/pingcap/tikv/PDClient.java
```diff
@@ -95,34 +97,93 @@ public class PDClient extends AbstractGRPCClient<PDBlockingStub, PDStub>
...
+    this.upstreamPDClient = createUpstreamPDClient(keyCodec);
+  }
+
+  private org.tikv.common.ReadOnlyPDClient createUpstreamPDClient(RequestKeyCodec keyCodec) {
+    org.tikv.common.TiConfiguration tikvConf = convertTiConfiguration(conf);
+    org.tikv.common.util.ChannelFactory channelFactory = convertChannelFactory(conf, tikvConf);
+    return org.tikv.common.PDClient.create(tikvConf, keyCodec, channelFactory);
+  }
+
+  private org.tikv.common.TiConfiguration convertTiConfiguration(
```
xuanyu66, line 116: Should be careful about this convertation.
Can you give a doc about diff and mapping of two configurations

tikv-client/src/main/java/com/pingcap/tikv/region/RegionManager.java
```diff
@@ -92,7 +92,11 @@ public Pair<TiRegion, Store> getRegionStorePairByKey(ByteString key) {
   }
 
   public Pair<TiRegion, Store> getRegionStorePairByKey(ByteString key, TiStoreType storeType) {
-    return getRegionStorePairByKey(key, storeType, ConcreteBackOffer.newGetBackOff());
+    return getRegionStorePairByKey(key, storeType, ConcreteBackOffer.newGetBackOff(getClusterId()));
+  }
+
+  public long getClusterId() {
+    return getPDClient() != null ? getPDClient().getClusterId() : 0L;
```
xuanyu66, line 99: Under which situation getPDClient will be null?
Should it return 0L or just throw exception when pdclient==null?

tikv-client/src/main/java/com/pingcap/tikv/PDClient.java
```diff
@@ -95,34 +97,93 @@ public class PDClient extends AbstractGRPCClient<PDBlockingStub, PDStub>
...
+              conf.getMaxFrameSize(),
+              tikvConf.getKeepaliveTime(),
+              tikvConf.getKeepaliveTimeout(),
+              tikvConf.getIdleTimeout());
+    }
+    return tikvFactory;
+  }
+
+  public Long getClusterId() {
+    return header.getClusterId();
```
xuanyu66, line 171: Why we getClusterId from header

xuanyu66 (changes requested):

tikv-client/src/main/java/com/pingcap/tikv/region/RegionManager.java
ajian2002: 虽然构造函数中需要提供Pdclient,但是外部可以传入Null值作为PDClient.

tikv-client/src/main/java/com/pingcap/tikv/PDClient.java
ajian2002: Implements the ReadOnlyPDClient interface, which has the same behavior as upstream

tikv-client/src/main/java/com/pingcap/tikv/ReadOnlyPDClient.java
```diff
@@ -16,52 +16,19 @@
...
+public interface ReadOnlyPDClient extends org.tikv.common.ReadOnlyPDClient {
 
-  /**
-   * Get Region from PD by key specified
-   *
-   * @param key key in bytes for locating a region
-   * @return the region whose startKey and endKey range covers the given key
-   */
-  TiRegion getRegionByKey(BackOffer backOffer, ByteString key);
+  Future<Pair<Metapb.Region, Metapb.Peer>> getRegionByKeyAsync(BackOffer backOffer, ByteString key);
```
xuanyu66, line 29: It's better to keep the comment

tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
```diff
@@ -419,7 +428,7 @@ private List<TiRegion> splitRegion(List<ByteString> splitKeys, BackOffer backOff
             "split key equal to region start key or end key. Region splitting is not needed.");
       } else {
         logger.info("start to split region id={}, split size={}", region.getId(), splits.size());
-        List<TiRegion> newRegions;
+        List<Metapb.Region> newRegions;
```
xuanyu66, line 431: Why use TiStore and Metapb.Region rather than Metapb.Store and TiRegion

tikv-client/src/main/java/com/pingcap/tikv/region/RegionManager.java
```diff
@@ -183,121 +261,20 @@ public void invalidateRegion(TiRegion region) {
...
-        if (store.getState().equals(StoreState.Tombstone)) {
-          return null;
-        }
-        storeCache.put(id, store);
-        return store;
+        stores.add(getStoreById(peer.getStoreId(), backOffer));
+        peers.add(peer);
       } catch (Exception e) {
-        throw new GrpcException(e);
+        logger.warn("Store {} not found: {}", peer.getStoreId(), e.toString());
```
xuanyu66, line 272: Why change the behavior?

tikv-client/src/main/java/com/pingcap/tikv/util/ConverterUpstream.java
```diff
@@ -0,0 +1,96 @@
...
+ */
+
+package com.pingcap.tikv.util;
+
+import java.util.Objects;
+import java.util.stream.Collectors;
+import org.tikv.common.TiConfiguration.ApiVersion;
+import org.tikv.common.apiversion.RequestKeyCodec;
+
+public class ConverterUpstream {
```
xuanyu66, line 24: add some code comments to describe the use of this class

tikv-client/src/test/java/com/pingcap/tikv/RegionManagerTest.java
```diff
@@ -95,16 +121,36 @@ public void getStoreByKey() {
...
+    pdServer.addGetStoreResp(
+        GrpcUtils.makeGetStoreResponse(
+            pdServer.getClusterId(),
+            GrpcUtils.makeStore(
+                ids[i.getAndIncrement()],
+                testAddress,
+                StoreState.Up,
+                GrpcUtils.makeStoreLabel("k1", "v1"),
+                GrpcUtils.makeStoreLabel("k2", "v2"))));
+    //    pdServer.addGetStoreResp(
```
xuanyu66, line 144: clean these code

xuanyu66 (changes requested):

tikv-client/src/main/java/com/pingcap/tikv/region/RegionManager.java
ajian2002: Our original logic does not need pdclient, nor does it need to obtain clusterId, and use the default behavior directly. So a null pdclient is passed in it. Now after using the upstream it needs to display the clusterId from pdclient, so to keep the behavior consistent, I pass the default value of 0.

tikv-client/src/main/java/com/pingcap/tikv/PDClient.java
ajian2002: Maybe a conversion tool class code can more clearly show how we convert some upstream and downstream configurations. as I am
[tikv-client/src/main/java/com/pingcap/tikv/util/ConverterUpstream.java](https://github.com/pingcap/tispark/pull/2399/files#diff-127d716c9af974c920da0c8817c800a1b7029895c7fd0514eae3ac454d12c3b5:~:text=tikv%2Dclient/src/main/java/com/pingcap/tikv/util/ConverterUpstream.java)
  as in the file

tikv-client/src/main/java/com/pingcap/tikv/ReadOnlyPDClient.java
ajian2002: Is this necessary? When not more people are using this, I prefer to align with upstream.

tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
ajian2002: Only here the private method will use MetaPB.Region. The reason is that the construction method of TiRegion is not as simple as before (the old TIRegion can be used directly), I copied the upstream to write the code and the interface of RegionStoreClient is very different upstream and downstream, I will merge the upstream and downstream RegionStoreClient in another pull request.

tikv-client/src/main/java/com/pingcap/tikv/region/RegionManager.java
ajian2002: This is because getStoreById needs to throw exceptions before, but now our getStoreById will still throw exceptions, and createRegion needs to ignore exceptions generated by calling getStoreById.

PR #2433, opened 2022-06-24
fix region may be missed

### LINKED ISSUE #2368: Table not found when TiDB has more than 10,000 tables in a DB
## Enhancement
When TiDB has more than 10,000 tables in one Database, TiSpark may throw exceptions like `Table not found`.

This problem may be because of the large schema, if so, we need to clarify how large the schema will lead to the problem. 

### Verison
TiDB 5.4
Spark 3.1.3
TiSpark 2.5

TrafalgarRicardoLu:
close #2368 
When TiDB has more than 10,000 tables in one Database, TiSpark may throw Table not found  exceptions.

The problem reported by the issue is caused by not timely update of region.

When there have stale data in region cache, the iterator will use end key in stale region to scan regions, and it may miss some new regions. And if we try to read the table in new region, table not found  exceptions will be thrown.

In fact, RegionStoreClient has already gotten the new region, but doesn't send it to ConcreteScanIterator. So we add the code to update region.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ConcreteScanIterator.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ConcreteScanIterator.java
@@ -66,6 +66,10 @@ public class ConcreteScanIterator extends ScanIterator {
           try {
             int scanSize = Math.min(limit, conf.getScanBatchSize());
             currentCache = client.scan(backOffer, startKey, scanSize, version);
+            // If we get region before scan, we will use region from cache which
+            // may have wrong end key. This may miss some regions that split from old region.
+            // Client will get the newest region during scan. So we need to
+            // update region after scan.
             region = client.getRegion();
           } catch (final TiKVException e) {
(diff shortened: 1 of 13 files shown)


tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/ConcreteScanIterator.java
```diff
@@ -66,6 +66,7 @@ TiRegion loadCurrentRegionToCache() throws GrpcException {
           try {
             int scanSize = Math.min(limit, conf.getScanBatchSize());
             currentCache = client.scan(backOffer, startKey, scanSize, version);
+            region = client.getRegion();
```
xuanyu66, line 69: Plz add some comment on the side effect of scan and why we update here

zhangyangyu (approved):

Daemonxiao (approved):

xuanyu66 (approved):

shiyuhang0:
LGTM

PR #2436, opened 2022-06-30
Support host mapping in TiSpark

Daemonxiao:
TiKV client supports custom HostMapping implementation in this PR: https://github.com/tikv/client-java/pull/171.

TiSpark implements the HostMapping interface and set a host-to-host mapping rule.
You can set it in the configuration file, such as:
```
spark.tispark.tikv.host_mapping = host1:host1_mapping;host2:host2_mapping
```
The host1 will be replaced by host1_mapping when TiSpark establishes or get a connection with TiKV.

### When to use this configuration?
When your TiDB cluster is running on the intranet, while Spark is in another intranet, you need to set public IPs to the TiDB cluster in TiSpark for connecting Spark with TiDB. Because of the PD-leader-switch(to find a new leader PD and get started with an intranet IP of PD) in TiSpark, we always get an intranet IP of PD which can not be used to comminute from Spark to TiDB. Then, we need a host mapping to make Spark always use a public IP to connect PD.

Custom host mapping rule, and use rule in ChannelFactory.
```
// Custom host mapping
public class UriHostMapping implements HostMapping {

  private static final Logger logger = LoggerFactory.getLogger(UriHostMapping.class.getName());
  private ConcurrentMap<String, String> hostMapping = null;

  // Parsing the host mapping configuration
  // Set a map that stores the mapping rule, such as {host1: host1_mapping; host2: host2_mapping}
  public UriHostMapping(String hostMappingString) {
      ...
  }

  // If the URI has a mapping rule, return the mapping rule. If not return itself.
  @Override
  public URI getMappedURI(URI uri) {
  }
}

// In ChannelFactory use rule
private ManagedChannel addrToChannel(SslContextBuilder sslContextBuilder, String addressStr, HostMapping mapping){
    ...
    // Get host mapping
    mapped = mapping.getMappedURI(address);
    
    // use mapped to connect to TiKV
     ... 
}
```

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Related changes


core/src/main/java/com/pingcap/tikv/hostmap/UriHostMapping.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class UriHostMapping implements HostMapping {
+
+  private static final Logger logger = LoggerFactory.getLogger(UriHostMapping.class.getName());
+  private ConcurrentMap<String, String> hostMapping = null;
```
xuanyu66, line 32:
```suggestion
  private final ConcurrentMap<String, String> hostMapping;
```

core/src/main/scala/com/pingcap/tispark/telemetry/TiSparkTeleInfo.scala
```diff
@@ -93,11 +94,22 @@ object TiSparkTeleInfo {
...
+      if (conf.getHostMapping != null) {
+        pd_uri = conf.getHostMapping.getMappedURI(pd_uri)
+      }
+
+      if (conf.isTlsEnable) {
+        val url = pd_uri.toString
         resp = httpClient.getHttpsWithTiConfiguration(url, conf)
       } else {
-        val url = "http://" + pd_address.get + urlPattern
+        val url = pd_uri.toString
```
xuanyu66, line 112: Can we remove the val url here and above, just use pd_uri.toString.
Since url is only used once, and defining url doesn't increase the readability.

core/src/test/scala/com/pingcap/tispark/hostmap/HostMappingSuite.scala
```diff
@@ -0,0 +1,72 @@
...
+    val tiConf: TiConfiguration = new TiConfiguration()
+    TiUtil.sparkConfToTiConfWithoutPD(conf, tiConf)
+    assert(tiConf.getHostMapping.asInstanceOf[UriHostMapping].getHostMapping.get("host1") == "1")
+    assert(tiConf.getHostMapping.asInstanceOf[UriHostMapping].getHostMapping.size() == 3)
+  }
+
+  test("Host map configuration test with wrong host map") {
+    the[IllegalArgumentException] thrownBy {
+      val conf: SparkConf = new SparkConf(false)
+      conf.set(TIKV_HOST_MAPPING, "")
```
xuanyu66, line 43: Empty is reasonably considered to be valid

tikv-client/src/main/java/com/pingcap/tikv/PDClient.java
```diff
@@ -545,19 +557,42 @@ protected PDStub getAsyncStub() {
 
   private void initCluster() {
     GetMembersResponse resp = null;
-    List<URI> pdAddrs = getConf().getPdAddrs();
+    List<URI> pdAddrs = new ArrayList<>(getConf().getPdAddrs());
+    // shuffle PD addresses so that clients call getMembers from different PD
```
xuanyu66, line 561: What's the reason for different PD?

tikv-client/src/main/java/com/pingcap/tikv/TiConfiguration.java
```diff
@@ -116,6 +117,9 @@ public class TiConfiguration implements Serializable {
   private long connRecycleTime = getTimeAsSeconds(DEF_TIKV_CONN_RECYCLE_TIME);
   private long certReloadInterval = getTimeAsSeconds(DEF_TIKV_TLS_RELOAD_INTERVAL);
 
+  private String networkMappingName = "";
```
xuanyu66, line 120: Why add two private variables here and never access to them.

xuanyu66 (changes requested):

core/src/main/java/com/pingcap/tikv/hostmap/UriHostMapping.java
```diff
@@ -0,0 +1,76 @@
...
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Arrays;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class UriHostMapping implements HostMapping {
```
xuanyu66, line 29: Why this class doesn't locate in tikv-client package

shiyuhang0:
it seems we don't support host mapping with Datasource option?

core/src/main/java/com/pingcap/tikv/hostmap/UriHostMapping.java
Daemonxiao: @xuanyu66 Thanks for your review. I will change it.

core/src/main/scala/com/pingcap/tispark/telemetry/TiSparkTeleInfo.scala
Daemonxiao: Yes, It is significant to make the code more readable. I will delete the unuseful variable.

core/src/test/scala/com/pingcap/tispark/hostmap/HostMappingSuite.scala
Daemonxiao: OK, I will change the logic here.

tikv-client/src/main/java/com/pingcap/tikv/PDClient.java
Daemonxiao: `getConf().getPdAddrs()` is a list that came from configurations created by users. `pdAddrs` is a new list used to be shuffled. Because we always use the first address in the list to get the PD leader in init or retry. Shuffling is helpful to make loading balance.

tikv-client/src/main/java/com/pingcap/tikv/TiConfiguration.java
Daemonxiao: This is a special configuration for some users. Maybe we can ignore it here.

core/src/main/java/com/pingcap/tikv/hostmap/UriHostMapping.java
Daemonxiao: The location of UriHostMapping is not a matter. It depends on our view of the `tikv-client` package. The UriHostMapping is a customization of HostMapping in TiSpark, and I try to put it in the `core` model instead of the 'tikv-client' model.

core/pom.xml
```diff
@@ -14,6 +14,10 @@
     <name>TiSpark Project Core Internal</name>
     <url>http://github.copm/pingcap/tispark</url>
 
+    <properties>
+        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.compiler.source>1.8</maven.compiler.source>
+    </properties>
```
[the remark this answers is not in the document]
Daemonxiao: That isn't very clear to me. If I don't set the JDK version, It will use the 1.6 language level, which will make us unable to use lambda, steam, and so on in the `core` module.

core/src/main/java/com/pingcap/tikv/hostmap/UriHostMapping.java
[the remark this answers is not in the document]
Daemonxiao: @shiyuhang0 Thanks for your review. I will change it.

tikv-client/src/main/java/com/pingcap/tikv/PDClient.java
```diff
@@ -545,19 +557,42 @@ protected PDStub getAsyncStub() {
...
     long clusterId = resp.getHeader().getClusterId();
     header = RequestHeader.newBuilder().setClusterId(clusterId).build();
     tsoReq = TsoRequest.newBuilder().setHeader(header).setCount(1).build();
-    this.pdAddrs = pdAddrs;
-    this.etcdClient = Client.builder().endpoints(pdAddrs).build();
+    this.pdAddrs =
+        resp.getMembersList()
+            .stream()
+            .map(mem -> addrToUrl(mem.getClientUrls(0)))
+            .collect(Collectors.toList());
```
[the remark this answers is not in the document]
Daemonxiao: It is used to flush the PD address list from PD server.

docs/configuration.md
```diff
@@ -40,6 +40,7 @@
...
 | `spark.tispark.tikv.tls_reload_interval`        | `10s`            | The interval time of checking if there is any reloading certificates. The default is `10s` (10 seconds).                                                                                                                                                                                                              |
 | `spark.tispark.tikv.conn_recycle_time`          | `60s`            | The interval time of cleaning the expired connection with TiKV. Only take effect when enabling cert reloading. The default is `60s` (60 seconds).                                                                                                                                                                     |
+| `spark.tispark.tikv.host_mapping`               | ``               | This is route map used to configure public IP and intranet IP mapping.The format is `{Intranet IP}:{Public IP}`, e.g. `127.0.0.1:192.168.100.100`.                                                                                                                                                                    |
```
zhangyangyu, line 43: As for public ip, I recommend using ips like 8.8.8.8. And the example should also show that we use `;` as separator, otherwise, uses can't get it.

core/src/main/java/com/pingcap/tikv/hostmap/UriHostMapping.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  @Override
+  public URI getMappedURI(URI uri) {
+    if (hostMapping != null && hostMapping.containsKey(uri.getHost())) {
+      try {
+        return new URI(
+            uri.getScheme(),
+            uri.getUserInfo(),
+            hostMapping.get(uri.getHost()),
```
zhangyangyu, line 68: Does the `host_mapping` must contain all the ip pairs? I mean is it possbile `hostMapping.get(uri.getHost())` returns `NULL`?

docs/configuration.md
Daemonxiao: @zhangyangyu Thanks for your review. I've changed it.

core/src/main/java/com/pingcap/tikv/hostmap/UriHostMapping.java
Daemonxiao: No, `host_mapping` can be empty. If an IP address can not find a map value in `host_mapping`, then, return itself.

zhangyangyu: Oops, miss the above `hostMapping.containsKey(uri.getHost())`.

core/src/main/java/com/pingcap/tikv/hostmap/UriHostMapping.java
```diff
@@ -0,0 +1,80 @@
...
+                  s -> {
+                    String[] hostAndPort = s.split(":");
+                    return new ConcurrentHashMap.SimpleEntry<>(hostAndPort[0], hostAndPort[1]);
+                  })
+              .collect(
+                  Collectors.toConcurrentMap(
+                      ConcurrentHashMap.SimpleEntry::getKey,
+                      ConcurrentHashMap.SimpleEntry::getValue));
+    } catch (Exception e) {
+      logger.error("Invalid host mapping string: {}", hostMappingString, e);
```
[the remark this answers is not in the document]
xuanyu66: For log4j, an error with  a formatter only print message while an error without a formatter will print both message and stack

docs/configuration.md
```diff
@@ -3,43 +3,44 @@
...
+| `spark.tispark.tikv.tls_reload_interval`        | `10s`            | The interval time of checking if there is any reloading certificates. The default is `10s` (10 seconds).                                                                                                                                                                                                                           |
+| `spark.tispark.tikv.conn_recycle_time`          | `60s`            | The interval time of cleaning the expired connection with TiKV. Only take effect when enabling cert reloading. The default is `60s` (60 seconds).                                                                                                                                                                                  |
+| `spark.tispark.host_mapping`                    | ``               | This is route map used to configure public IP and intranet IP mapping. When the TiDB cluster is running on the intranet, you can map a set of intranet IPs to public IPs for a outside Spark cluster to access. The format is `{Intranet IP}:{Public IP};{Intranet IP}:{Public IP}`, e.g. `127.0.0.1:8.8.8.8;192.168.0.1:8.8.8.8`. |
```
zhangyangyu, line 43:
```suggestion
| `spark.tispark.host_mapping`                    | ``               | This is route map used to configure public IP and intranet IP mapping. When the TiDB cluster is running on the intranet, you can map a set of intranet IPs to public IPs for an outside Spark cluster to access. The format is `{Intranet IP1}:{Public IP1};{Intranet IP2}:{Public IP2}`, e.g. `192.168.0.2:8.8.8.8;192.168.0.3:9.9.9.9`. |
```

xuanyu66 (approved):

zhangyangyu (approved):

shiyuhang0 (approved):

Issue #1142, opened 2019-10-10
count(*) is not being pushdown


zhexuany:
**Describe the bug**

For a simple query `select count(*) from t`, tispark fails to push down to storage engine. 

**What did you do**
count(*) case:

```
scala> spark.sql("select count(*) from test.t").explain
== Physical Plan ==
*(2) HashAggregate(keys=[], functions=[count(1)])
+- Exchange SinglePartition
   +- *(1) HashAggregate(keys=[], functions=[partial_count(1)])
      +- TiKV CoprocessorRDD{[table: t] TableScan, Columns: [i], KeyRange: ([t\200\000\000\000\000\000\0002_r\000\000\000\000\000\000\000\000], [t\200\000\000\000\000\000\0002_s\000\000\000\000\000\000\000\000]), startTs: 411745195321720833}
```

count(i) case (the difference is * and a specific column)
```
scala> spark.sql("select count(i) from test.t").explain
== Physical Plan ==
*(2) HashAggregate(keys=[], functions=[specialsum(count(i#140L)#143L, LongType, 0)])
+- Exchange SinglePartition
   +- *(1) HashAggregate(keys=[], functions=[partial_specialsum(count(i#140L)#143L, LongType, 0)])
      +- TiKV CoprocessorRDD{[table: t] TableScan, Columns: [i], KeyRange: ([t\200\000\000\000\000\000\0002_r\000\000\000\000\000\000\000\000], [t\200\000\000\000\000\000\0002_s\000\000\000\000\000\000\000\000]), Aggregates: Count([i]), startTs: 411745336001298433}
```

crabo:
Hi there, any progress?

PR #2445, opened 2022-07-07
Fix count/avg push down

shiyuhang0:
close https://github.com/pingcap/tispark/issues/2313
 https://github.com/pingcap/tispark/pull/1840
close https://github.com/pingcap/tispark/issues/1142

this bug is simply fixed by push down Constant(1) when count(1) or count(*)
 - push down count by add pattern match for count, but
    -  count(set) will not be pushed down because tikv can't accpect set type.
    - count(col1,col2,...) will not be pushed down because TiKV can't accpect (TIDB not support it)
 - push down Constant(1) when count(1) or count(*)
 - rewrite test and add test for count/avg

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/catalyst/expressions/TiExprUtils.scala
+++ b/core/src/main/scala/org/apache/spark/sql/catalyst/expressions/TiExprUtils.scala
@@ -98,6 +98,7 @@ object TiExprUtils {
           }
 
           dagRequest.addRequiredColumn(firstColRef)
+          // we need to push down Constant(1) for count(1) or count(*)
           Constant.create(1, null)
         } else {
           args.flatMap(BasicExpression.convertToTiExpr).head
--- a/core/src/main/scala/org/apache/spark/sql/catalyst/expressions/TiExprUtils.scala
+++ b/core/src/main/scala/org/apache/spark/sql/catalyst/expressions/TiExprUtils.scala
@@ -96,8 +96,9 @@ object TiExprUtils {
               dagRequest.getFields.head
             }
           }
-
-          dagRequest.addRequiredColumn(firstColRef)
+          if (!dagRequest.getFields.isEmpty) {
+            dagRequest.addRequiredColumn(firstColRef)
+          }
           // we need to push down Constant(1) for count(1) or count(*)
           Constant.create(1, null)
         } else {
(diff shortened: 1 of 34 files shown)


core/src/main/scala/org/apache/spark/sql/catalyst/expressions/TiExprUtils.scala
```diff
@@ -91,7 +98,7 @@ object TiExprUtils {
           }
 
           dagRequest.addRequiredColumn(firstColRef)
-          firstColRef
+          Constant.create(1, null)
```
xuanyu66, line 101: Add some code comments

xuanyu66 (changes requested):

Daemonxiao:
Could you please add some content about how to solve this bug in "How it works"?

shiyuhang0:
> Could you please add some content about how to solve this bug in "How it works"?

done

xuanyu66 (approved):

core/src/test/scala/org/apache/spark/sql/pushdown/AvePushDownSuite.scala
```diff
@@ -0,0 +1,75 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.pushdown
+
+/**
+ * AVG only push down when count and sum can push down
+ * year type is not tested, because spark jdbc not support
+ */
+class AvePushDownSuite extends BasePushDownSuite() {
```
zhangyangyu, line 23: rename to avg plz

zhangyangyu (approved):

core/src/test/scala/org/apache/spark/sql/pushdown/AvgPushDownSuite.scala
```diff
@@ -0,0 +1,75 @@
...
+    "select avg(tp_double) from ",
+    "select avg(tp_int) from ",
+    "select avg(tp_mediumint) from ",
+    "select avg(tp_real) from ",
+    "select avg(tp_smallint) from ",
+    "select avg(tp_tinyint) from ")
+
+  private val notPush = Seq[String]("select avg(tp_float) from ")
+
+  test("Test - Ave push down pk") {
```
zhangyangyu, line 38:
```suggestion
  test("Test - Avg push down pk") {
```

PR #2447, opened 2022-07-08
Update license-check.yml

kezhenxu94:
We have a dedicate GitHub actions for header check so this should be updated, and we don't recommend using `@main` so there is no unexpected behaviors when we have changes in our codebase.

See https://github.com/apache/skywalking-eyes/pull/123


shiyuhang0 (approved):

shiyuhang0:
should closed by https://github.com/pingcap/tispark/pull/2446

kezhenxu94:
Closing in favor of

PR #2455, opened 2022-07-12
fix: Incorrect unique index key when table is not intHandle & Duplicate values for unique indexes

### LINKED ISSUE #2391: [BUG] incorrect uniqueIndex key when table is not intHandle
It's a mirror bug, you can see https://github.com/tidb-incubator/TiBigData/issues/214.
TiBigData and TiSpark have the same code in logic of generating uniqueIndex key.

### LINKED ISSUE #2452: [BUG] Duplicate values for unique indexes
**Describe the bug**

When we insert two values with duplicate unique indexes but non-duplicate primary keys, both data are stored in TiDB.
**What did you do**

```scala
    tidbStmt.execute("drop table if exists t")
    tidbStmt.execute(
      """
        |CREATE TABLE `t` (
        |  `id` bigint(20),
        |  `name` varchar(20) primary key clustered,
        |  `age` int(11) null default null,
        |   unique index(id)
        |) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin
      """.stripMargin)
    val schema = StructType(List(
      StructField("id", IntegerType, nullable = true),
      StructField("name", StringType, nullable = false),
      StructField("age", IntegerType, nullable = true),
    ))
    val tidbOptions: Map[String, String] = Map(
      "tidb.addr" -> "127.0.0.1",
      "tidb.password" -> "",
      "tidb.port" -> "4000",
      "tidb.user" -> "root",
      "replace" -> "true",
    )
    val rdd1 = sc.parallelize(Seq(
      Row(1, "1", 0),
    ))
    val row1 = sqlContext.createDataFrame(rdd1, schema)
    row1.write.format("tidb").
      option("database", "tispark_test").
      option("table", "t").
      options(tidbOptions).
      mode("append").
      save()
    val rdd2 = sc.parallelize(Seq(
      Row(1, "2", 0),
    ))
    val row2 = sqlContext.createDataFrame(rdd2, schema)
    row2.write.format("tidb").
      option("database", "tispark_test").
      option("table", "t").
      options(tidbOptions).
      mode("append").
      save()
```
**What do you expect**

image
I expect only one row in tidb.
**What happens instead**

image
There two rows in tidb.
** Version **
TiSpark: master

qidi1:

* close #2452
* close #2391
* Incorrect unique index key when the table is not intHandle
  
  * Why does this problem occur

    In TiDB, if there are null values in two unique indexes, then these two unique indexes are not conflicting.
    In order to achieve the purpose of no conflict, TiDB encodes the rows with null in the unique index by adding the clustered index or RowID to the key of the unique index.
    But in TiSpark, when the clustered index is not Integer type, the clustered index will not be added after the encoded key. This leads to a conflict in TiSpark when inserting two rows with null values in the unique index when the clustered index is not of Integer type
  
  * How do we solve this problem
  
    The original logic only adds handle to Rowley when any col in the unique index is if the handle is of type int. It is now added in all cases.
    The original code:
  
      ```java
          if (handle.isInt()) {
            // set appendHandle=true when any col in the unique index is null.
             ....
          }
         ...
          }
          if (appendHandle) {
            Key key = TypedKey.toTypedKey(handle, IntegerType.BIGINT);
            keys[keys.length - 1] = key;
          }
      ```

      The changed code:

      ```java
          // set appendHandle=true when any col in the unique index is null
           ....
          if (appendHandle) {
            Key key = TypedKey.toRawKey(handle.encodedAsKey());
            keys[keys.length - 1] = key;
          }
      ```
  
* Duplicate values for unique indexes

  * Why does this problem occur

    This problem is caused by the fact that When we insert data, we can't decode the Handle out of a row with a unique index conflict.
    For example:

    First, let's assume that the unique index data in TIKV is {1: “1”} and the cluster key data is {"1":1,“1“,0}.
    Now we use TiSpark to insert a row of data {1,“2”,0} to TiDB.

    1. determine whether there is a conflicting row, request the row with unique index data of 1 from TiKV and return the value as {1, "1"}, request the row with cluster index key of "2" from TiKV and return null.

    2. resolve the cluster index key value from the conflicting row {1: "1"} and delete the row corresponding to the primary key. However, due to the error of decode function, we can't resolve the correct cluster index key, so we can't delete the row corresponding to the cluster index key,.

    3. We insert the unique index data {1:"2"} and the primary key data {"2":1,“2”,0} into TiKV.

    At this time, the unique index data in TiKV is {1:"2"} and the primary key data is {"1":1,“1”,0} and {"2":1,"2",0}. Two unique indexes of the same data appear in the database.

  * How we solve this problem
    
    The index value layout is like that.

    ```text
       // Value layout:
       //		+-- IndexValueVersion0  (with restore data, or common handle, or index is global)
       //		|
       //		|  Layout: TailLen | Options      | Padding      | [IntHandle] | [UntouchedFlag]
       //		|  Length:   1     | len(options) | len(padding) |    8        |     1
       //		|
       //		|  TailLen:       len(padding) + len(IntHandle) + len(UntouchedFlag)
       //		|  Options:       Encode some value for new features, such as common handle, new collations or global index.
       //		|                 See below for more information.
       //		|  Padding:       Ensure length of value always >= 10. (or >= 11 if UntouchedFlag exists.)
       //		|  IntHandle:     Only exists when table use int handles and index is unique.
       //		|  UntouchedFlag: Only exists when index is untouched.
       //		|
       //		+-- Old Encoding (without restore data, integer handle, local)
       //		|
       //		|  Layout: [Handle] | [UntouchedFlag]
       //		|  Length:   8      |     1
       //		|
       //		|  Handle:        Only exists in unique index.
       //		|  UntouchedFlag: Only exists when index is untouched.
       //		|
       //		|  If neither Handle nor UntouchedFlag exists, value will be one single byte '0' (i.e. []byte{'0'}).
       //		|  Length of value <= 9, use to distinguish from the new encoding.
       // 		|
       //		+-- IndexValueForClusteredIndexVersion1
       //		|
       //		|  Layout: TailLen |    VersionFlag  |    Version     ｜ Options      |   [UntouchedFlag]
       //		|  Length:   1     |        1        |      1         |  len(options) |         1
       //		|
       //		|  TailLen:       len(UntouchedFlag)
       //		|  Options:       Encode some value for new features, such as common handle, new collations or global index.
       //		|                 See below for more information.
       //		|  UntouchedFlag: Only exists when index is untouched.
       //		|
       //		|  Layout of Options:
       //		|
       //		|     Segment:             Common Handle                 |     Global Index      |   New Collation
       // 		|     Layout:  CHandle flag | CHandle Len | CHandle      | PidFlag | PartitionID |    restoreData
       //		|     Length:     1         | 2           | len(CHandle) |    1    |    8        |   len(restoreData)
       //		|
    ```
    
    When the clustered index is not int type, the original decodeHandle cannot decode the value of unqiue index correctly, so we added a new decode method `decodeIndexValueForClusteredIndexVersion1`.The new decode code is same as [TiDB](https://github.com/pingcap/tidb/blob/master/tablecodec/tablecodec.go#L937:~:text=value%29%2C%20nil-,%7D,-//%20decodeIntHandleInIndexValue%20uses%20to)
    
    The logic for decoding is shown in the flowchart
    
    ```mermaid
     graph TD
          A[decodeHandleInUniqueIndexValue] --> B{isCommHandle}
          B -->|Yes| C{getIndexVersion}
          C-->|1|D[decode<br/>commonHandle version1]
            C-->|0|F[decode commmonHandle <br/>version0 new encoding]
          B --No-->E{encodeData <br/>length < 9}
          E --Yes--> G[decode version0<br/> old encoding]
          E --No--> H[decode version0<br/> new encoding]
    ```

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects


tikv-client/src/test/java/com/pingcap/tikv/key/IndexKeyTest.java
```diff
@@ -48,6 +58,87 @@ public void createTest() {
     }
   }
 
+  @Test
+  public void encodeIndexDataValuesTest() {
+    TiColumnInfo col1 = new TiColumnInfo(1, "a", 0, IntegerType.INT, false);
+    TiColumnInfo col2 = new TiColumnInfo(1, "a", 1, StringType.VARCHAR, true);
```
shiyuhang0, line 64: use different id in new TiColumnInfo()

tikv-client/src/test/java/com/pingcap/tikv/key/IndexKeyTest.java
```diff
@@ -48,6 +58,87 @@ public void createTest() {
...
+          },
+          new Key[] {TypedKey.toTypedKey("2", StringType.VARCHAR)},
+        };
+    indexColumns2.add(index2);
+    for (int i = 0; i < testRows2.length; i++) {
+      Row row = ObjectRowImpl.create(testRows2[i]);
+      Handle handle = new IntHandle((Integer) row.get(0, IntegerType.BIGINT));
+      IndexKey.EncodeIndexDataResult result =
+          IndexKey.encodeIndexDataValues(row, indexColumns2, handle, true, tableInfo);
+      assertArrayEquals(expectations2[i], result.keys);
```
shiyuhang0, line 138: we need not alter tableInfo's pkHandle to true?

shiyuhang0:
does delete logical need be changed too, for it will decodeHandle When not pkhandle

shiyuhang0:
check if the upstream tikv-client-java need to be fixed

core/src/main/scala/com/pingcap/tispark/write/TiBatchWriteTable.scala
```diff
@@ -215,103 +215,73 @@ class TiBatchWriteTable(
...
+          .count()) {
+        throw new TiBatchWriteException("duplicate unique key or primary key")
+      }
+      distinctWrappedRowRdd
+    } else {
+      val rowIDAllocator = getRowIDAllocator(count)
+      tiRowRdd.zipWithIndex.map { row =>
+        val index = row._2 + 1
+        val rowId = rowIDAllocator.getShardRowId(index)
+        WrappedRow(row._1, new IntHandle(rowId))
```
shiyuhang0, line 233: I think we need to do deduplicate when uniqueIndices.nonEmpty here since you just judge `if (isCommonHandle || tiTableInfo.isPkHandle)`

core/src/main/scala/com/pingcap/tispark/write/TiBatchWriteTable.scala
```diff
@@ -482,94 +455,72 @@ class TiBatchWriteTable(
...
+              val oldRow = TableCodec.decodeRow(oldValue, oldHandle, tiTableInfo)
+              rowBuf += WrappedRow(oldRow, oldHandle)
             }
           }
         }
 
         def processNextBatch(): Unit = {
           rowBuf = mutable.ListBuffer.empty[WrappedRow]
-
+          val mayConflictRowKeys = mutable.Set.empty[Array[Byte]]
```
shiyuhang0, line 489: why need mayConflictRowKeys, and not just use rowBuf?

qidi1: > check if the upstream tikv-client-java need to be fixed

OK,I will check

qidi1:
> check if the upstream tikv-client-java need to be fixed

OK, I will check.

core/src/main/scala/com/pingcap/tispark/write/TiBatchWriteTable.scala
qidi1: > why need mayConflictRowKeys, and not just use rowBuf?

There are two reasons here.
1. the primary keys of inserted rows are not necessarily conflicting, we only know which columns are really conflicting after we send all the primary keys that may conflict to TiKV and get the returned results from TiKV.
2. the primary keys of the columns with conflicting unique indexes may be the same as the primary keys in the inserted columns, which are stored using Set variables for de-duplication.

tikv-client/src/test/java/com/pingcap/tikv/key/IndexKeyTest.java
qidi1: > we need not alter tableInfo's pkHandle to true?

we need. I changed

tikv-client/src/test/java/com/pingcap/tikv/key/IndexKeyTest.java
qidi1: > use different id in new TiColumnInfo()

changed

core/src/main/scala/com/pingcap/tispark/write/TiBatchWriteTable.scala
qidi1: > I think we need to do deduplicate when uniqueIndices.nonEmpty here since you just judge `if (isCommonHandle || tiTableInfo.isPkHandle)`

changed

qidi1:
> does delete logical need be changed too, for it will decodeHandle When not pkhandle

I has checked delete code, It will not has the problem.

core/src/main/scala/com/pingcap/tispark/write/TiBatchWriteTable.scala
```diff
@@ -215,103 +215,73 @@ class TiBatchWriteTable(
     // check value not null
     checkValueNotNull(tiRowRdd)
 
+    val wrappedRowRdd: RDD[WrappedRow] = if (isCommonHandle || tiTableInfo.isPkHandle) {
```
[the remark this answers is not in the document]
qidi1: add some annotations,ptal

tikv-client/src/main/java/com/pingcap/tikv/key/IndexKey.java
```diff
@@ -42,54 +43,51 @@ private IndexKey(long tableId, long indexId, Key[] dataKeys) {
...
-  public static EncodeIndexDataResult encodeIndexDataValues(
-      Row row,
-      List<TiIndexColumn> indexColumns,
-      Handle handle,
-      boolean appendHandleIfContainsNull,
-      TiTableInfo tableInfo) {
+  public static EncodeIndexDataResult genIndexKey(
+      long physicalID, Row row, TiIndexInfo indexInfo, Handle handle, TiTableInfo tableInfo) {
     // when appendHandleIfContainsNull is true, append handle column if any of the index column is
     // NULL
```
shiyuhang0, line 62: your logical has changed, the code annotation needs be changed too

tikv-client/src/main/java/com/pingcap/tikv/codec/TableCodec.java
```diff
@@ -85,4 +119,57 @@ private static void encodeCommonHandle(CodecDataOutput cdo, Handle handle) {
...
+  //    }
+  //    if (value.length > 0 && value[0] == PartitionIDFlag) {
+  //      segments.partitionID = Arrays.copyOfRange(value, 1, 9);
+  //      value = Arrays.copyOfRange(value, 9, value.length);
+  //    }
+  //    if (value.length > 0 && value[0] == RestoreDataFlag) {
+  //      segments.restoredValues = value;
+  //    }
+  //    return segments;
+  //  }
```
shiyuhang0, line 148: delete it if it is useless

tikv-client/src/main/java/com/pingcap/tikv/key/IndexKey.java
```diff
@@ -42,54 +43,51 @@ private IndexKey(long tableId, long indexId, Key[] dataKeys) {
...
-    boolean appendHandle = false;
```
shiyuhang0, line 66: why do you want to change appendHandle to distinct?
and if you changed the name, you need to change the place where you use it. such as https://github.com/pingcap/tispark/blob/6b6ac15859515145c16cfefedb5d2ac5e7ca8f01/core/src/main/scala/com/pingcap/tispark/utils/WriteUtil.scala#L238

core/src/main/scala/com/pingcap/tispark/utils/WriteUtil.scala
```diff
@@ -225,37 +225,30 @@ object WriteUtil {
...
+    val encodeResult =
+      IndexKey.genIndexKey(locatePhysicalTable(row, tiTableInfo), row, index, handle, tiTableInfo)
     val value = if (remove) {
       new Array[Byte](0)
     } else {
-      if (encodeResult.appendHandle) {
-        val value = new Array[Byte](1)
-        value(0) = '0'
-        value
+      if (!handle.isInt) {
```
shiyuhang0, line 234: can you add some explanation for your changed logic.
for example: when appendHandle is true and handle.isInt is false, the logic is diffierent.

tikv-client/src/main/java/com/pingcap/tikv/codec/TableCodec.java
```diff
@@ -66,15 +83,32 @@ public static Handle decodeHandle(byte[] value, boolean isCommonHandle) {
...
+      TiIndexInfo index, Handle handle, boolean distinct) {
     CodecDataOutput cdo = new CodecDataOutput();
     cdo.writeByte(0);
     cdo.writeByte(IndexVersionFlag);
     cdo.writeByte(1);
 
     assert (index.isUnique());
-    encodeCommonHandle(cdo, handle);
-
+    if (distinct) {
```
shiyuhang0, line 109: why judge it

core/src/main/scala/com/pingcap/tispark/utils/WriteUtil.scala
qidi1: I add some explain here.

tikv-client/src/main/java/com/pingcap/tikv/codec/TableCodec.java
qidi1: because only the distinct index key needs handle values in the index value.

qidi1: > why judge it

because only the distinct index key needs handle values in the index value.

tikv-client/src/main/java/com/pingcap/tikv/key/IndexKey.java
qidi1: > why do you want to change appendHandle to distinct? and if you changed the name, you need to change the place where you use it. such as
> 
> https://github.com/pingcap/tispark/blob/6b6ac15859515145c16cfefedb5d2ac5e7ca8f01/core/src/main/scala/com/pingcap/tispark/utils/WriteUtil.scala#L238

I renamed here

tikv-client/src/main/java/com/pingcap/tikv/codec/TableCodec.java
qidi1: > delete it if it is useless

Done

tikv-client/src/main/java/com/pingcap/tikv/key/IndexKey.java
qidi1: > your logical has changed, the code annotation needs be changed too

changed the annotation

core/src/main/scala/com/pingcap/tispark/utils/WriteUtil.scala
```diff
@@ -228,98 +228,42 @@ object WriteUtil {
...
+      remove)
   }
 
   /**
    * construct unique index and non-unique index and value to be inserted into TiKV
    * NOTE:
-   * pk is not handle case is equivalent to unique index.
-   * for non-unique index, handle will be encoded as part of index key. In contrast, unique
-   * index encoded handle to value.
+   *      pk is not handle case is equivalent to unique index.
```
xuanyu66, line 247: Plz change the comment related to code change

core/src/main/scala/com/pingcap/tispark/write/TiBatchWriteTable.scala
```diff
@@ -475,94 +455,80 @@ class TiBatchWriteTable(
...
             batch: List[WrappedRow],
-            index: TiIndexInfo): (util.List[Array[Byte]], util.List[TiRow]) = {
+            index: TiIndexInfo): util.List[Array[Byte]] = {
           val keyList = new util.ArrayList[Array[Byte]]()
-          val rowList = new util.ArrayList[TiRow]()
           batch.foreach { wrappedRow =>
             val encodeResult = buildUniqueIndexKey(wrappedRow.row, wrappedRow.handle, index)
-            if (!encodeResult._2) {
+            if (encodeResult._2) {
               // only add the key if handle is not appended, since if handle is appened,
```
xuanyu66, line 465: Change the comment

tikv-client/src/main/java/com/pingcap/tikv/codec/TableCodec.java
```diff
@@ -65,16 +83,49 @@ public static Handle decodeHandle(byte[] value, boolean isCommonHandle) {
...
+    }
+    int handleLen = ((int) value[2]) << 8 + value[3];
+    byte[] encode = Arrays.copyOfRange(value, 4, handleLen + 4);
+    return new CommonHandle(encode);
+  }
+
+  public static byte[] genIndexValue(Handle handle, boolean distinct) {
+    if (!handle.isInt()) {
+      // TODO
+      //  We need to implement the encoding of the index value version 0 when handle is not int.
```
xuanyu66, line 108: Why todo

xuanyu66 (approved):

core/src/main/scala/com/pingcap/tispark/write/TiBatchWriteTable.scala
qidi1: > Change the comment

changed

tikv-client/src/main/java/com/pingcap/tikv/codec/TableCodec.java
qidi1: > Why todo

support version0 encode now.

core/src/main/scala/com/pingcap/tispark/utils/WriteUtil.scala
qidi1: > Plz change the comment related to code change

The logic of this part code is not changed. The comment is no need to change.

xuanyu66 (approved):

shiyuhang0 (approved):

Issue #2462, opened 2022-07-20
[Proposal] Tispark support insert using SQL


TrafalgarRicardoLu:
## Problem:

TiSpark only support write data using datasource api. Users can‘t write data by insert SQL statement.

## Goals

TiSpark supports write data using insert SQL statement with following goals:

- Support with Spark 3.x
- Use friendly with spark SQL

### API

```
spark.sql("insert into tidb_catalog.db.table values(x, x,...)(y, y...)..")
```
## Solutions

The model of Spark datasource V2 is not suitable for TiSpark currently. So we are going to use V1 write in V2 interface.

### Implement
- Add new capability to TiDBTable
- Implement V1 Write in V2 interface
- Fix type convert
- Adapt to different spark versions

## Design

https://github.com/pingcap/tispark/pull/2464
## Implementation

https://github.com/pingcap/tispark/pull/2471/
## Test

https://github.com/pingcap/tispark/pull/2471/
## Doc

https://github.com/pingcap/tispark/pull/2471/

PR #2471, opened 2022-07-26
[Code] Tispark support insert using SQL

TrafalgarRicardoLu:
https://github.com/pingcap/tispark/issues/2462

- Add new capability to TiDBTable
- Remove write options
- Implement V2 interface using V1 code
- Compatible with different Spark versions
- close  https://github.com/pingcap/tispark/issues/2462
### Check List 

Tests 

 - Unit test
 - Integration test

Code changes


docs/design/2022-07-20-insert-support.md
```diff
@@ -0,0 +1,117 @@
+# Support Insert SQL 
+
+- Author(s): TrafalgarRicardoLu
```
xuanyu66, line 3: duplicated design doc？

core/src/main/scala/com/pingcap/tispark/write/TiDBOptions.scala
```diff
@@ -135,7 +135,7 @@ class TiDBOptions(@transient val parameters: CaseInsensitiveMap[String]) extends
   // Optional parameters for read
   // ------------------------------------------------------------
   val tidbRowId: Boolean = getOrDefault(TIDB_ROWID, "false").toBoolean
-
+  val tidbRowFormatVersion: Int = getOrDefault(TiDB_ROW_FORMAT_VERSION, "2").toInt
```
xuanyu66, line 138: It's better to write comment to explain why use 2 as default

xuanyu66 (approved):

xuanyu66 (changes requested):
Plz add use doc for reviewing

docs/insert_sql_userguide.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Type Conversion for Write
+
+The following types of SparkSQL Data are currently not supported for writing to TiDB:
+
+- ArrayType
+- MapType
+- StructType
+
+Spark SQL and TIDB have different conversion rules. We follow the rule of TIDB when using DataSource API,
```
xuanyu66, line 86: Plz give an example of how to cast such as String2TimeStamp and make spark link

core/src/main/scala/com/pingcap/tispark/v2/TiDBTable.scala
```diff
@@ -128,21 +130,31 @@ case class TiDBTable(
...
   override def newWriteBuilder(info: LogicalWriteInfo): WriteBuilder = {
     var scalaMap = info.options().asScala.toMap
-    // TODO https://github.com/pingcap/tispark/issues/2269 we need to move TiDB dependencies which will block insert SQL.
-    // if we don't support it before release, insert SQL should throw exception in catalyst
-    if (scalaMap.isEmpty) {
-      throw new TiBatchWriteException("tidbOption is neccessary.")
+    // For update statics
+    val enableUpdateStatics =
+      sqlContext.getConf(TiDBOptions.TIDB_ENABLE_UPDATE_TABLE_STATISTICS, "false").toBoolean
+    if (enableUpdateStatics) {
```
shiyuhang0, line 136: Just judge if the tidb config exists. if so, add them to scalaMap.
Judge by enableUpdateStatics is not a good idea. what if we need tidb config in another situation？

core/src/main/scala/com/pingcap/tispark/write/TiDBOptions.scala
```diff
@@ -282,6 +286,7 @@ object TiDBOptions {
   val TIDB_ENABLE_UPDATE_TABLE_STATISTICS: String = newOption("enableUpdateTableStatistics")
   val TIDB_DEDUPLICATE: String = newOption("deduplicate")
   val TIDB_ROWID: String = newOption("tidbRowId")
+  val TiDB_ROW_FORMAT_VERSION: String = newOption("rowFormatVersion")
```
shiyuhang0, line 289: how can insert sql pass this config?

core/src/test/scala/com/pingcap/tispark/insert/InsertSuite.scala
```diff
@@ -0,0 +1,119 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.pingcap.tispark.insert
+
+import com.pingcap.tispark.datasource.BaseBatchWriteTest
+
+class InsertSuite extends BaseBatchWriteTest("test_delete_compatibility") {
```
shiyuhang0, line 21: add test for `enableUpdateTableStatistics=true` and `rowFormatVersion=1`

spark-wrapper/spark-3.0/src/main/scala/org/apache/spark/sql/connector/write/TiDBWriteBuilder.scala
```diff
@@ -0,0 +1,44 @@
...
+      override def insert(data: DataFrame, overwrite: Boolean): Unit = {
+        val schema = info.schema()
+        val df = sqlContext.sparkSession.createDataFrame(data.toJavaRDD, schema)
+        df.write
+          .format("tidb")
+          .options(tiDBOptions.parameters)
+          .option(TiDBOptions.TIDB_DATABASE, tiDBOptions.database)
+          .option(TiDBOptions.TIDB_TABLE, tiDBOptions.table)
+          .mode(SaveMode.Append)
+          .save()
```
shiyuhang0, line 40: deduplicate needs to be set to false here, or it will have a different behavior from tidb

spark-wrapper/spark-3.1/src/main/scala/org/apache/spark/sql/connector/write/TiDBWriteBuilder.scala
```diff
@@ -0,0 +1,43 @@
...
+      override def insert(data: DataFrame, overwrite: Boolean): Unit = {
+        val schema = info.schema()
+        val df = sqlContext.sparkSession.createDataFrame(data.toJavaRDD, schema)
+        df.write
+          .format("tidb")
+          .options(tiDBOptions.parameters)
+          .option(TiDBOptions.TIDB_DATABASE, tiDBOptions.database)
+          .option(TiDBOptions.TIDB_TABLE, tiDBOptions.table)
+          .mode(SaveMode.Append)
+          .save()
```
shiyuhang0, line 39: ditto

spark-wrapper/spark-3.2/src/main/scala/org/apache/spark/sql/connector/write/TiDBWriteBuilder.scala
```diff
@@ -0,0 +1,47 @@
...
+            val schema = info.schema()
+            println("Do write")
+            val df = sqlContext.sparkSession.createDataFrame(data.toJavaRDD, schema)
+            df.write
+              .format("tidb")
+              .options(tiDBOptions.parameters)
+              .option(TiDBOptions.TIDB_DATABASE, tiDBOptions.database)
+              .option(TiDBOptions.TIDB_TABLE, tiDBOptions.table)
+              .mode(SaveMode.Append)
+              .save()
```
shiyuhang0, line 42: ditto

docs/insert_sql_userguide.md
```diff
@@ -0,0 +1,125 @@
...
+          "insert into tpch_test.target_table_orders values" +
+            "(1,1,'0',11.11,date'2022-1-1','first','spark',12345,'no comment')")
+
+    // select data to write
+    val df = spark.sql("select * from tpch_test.ORDERS")
+   ```
+
+## TiDB Options
+
+The following table shows the TiDB-specific options, which can be passed in through `TiDBOptions` or `SparkConf`.
```
shiyuhang0, line 50: Spark SQL can't accept TiDBOptions. So, we need not mention about`TiDBOptions` here?

docs/insert_sql_userguide.md
```diff
@@ -0,0 +1,125 @@
...
+| tidb.password               | false    | -                      | TiDB Password, needed when enableUpdateTableStatistics is true                                                           |
+| enableRegionSplit           | false    | true                   | To split Region to avoid hot Region during insertion                                                                     |
+| scatterWaitMS               | false    | 300000                 | Max time to wait scatter region                                                                                          |
+| writeThreadPerTask          | false    | 1                      | Thread number each spark task use to write data to TiKV                                                                  |
+| bytesPerRegion              | false    | 100663296 (96M)        | Decrease this parameter to split more regions (increase write concurrency)                                               |
+| enableUpdateTableStatistics | false    | false                  | Update statistics in table `mysql.stats_meta` (`tidb.user` must own update privilege to table `mysql.stats_meta` if set true) |
```
shiyuhang0, line 65: - I think you can't pass enableRegionSplit,writeThreadPerTask,bytesPerRegion,scatterWaitMS beacause you don't get them in `newWriteBuilder`.
- database , table and pd.addresses is needn't.
- lack TiDB_ROW_FORMAT_VERSION

docs/insert_sql_userguide.md
shiyuhang0, line 86: I think DataSource API may not follow TiDB

docs/insert_sql_userguide.md
```diff
@@ -0,0 +1,125 @@
...
+## Type Conversion for Write
+
+The following types of SparkSQL Data are currently not supported for writing to TiDB:
+
+- ArrayType
+- MapType
+- StructType
+
+Spark SQL and TIDB have different conversion rules. We follow the rule of TIDB when using DataSource API,
+but follow the rule of Spark SQL when using it.
```
shiyuhang0, line 87:
```suggestion
but follow the rule of Spark SQL when using insert.
```

core/src/main/scala/com/pingcap/tispark/write/TiDBOptions.scala
TrafalgarRicardoLu: users need to set it by `spark.conf.set()`. I'll add it  in `newWriteBuilder `

docs/insert_sql_userguide.md
TrafalgarRicardoLu: I have fix point two and three. For point one, we will get them in  newWriteBuilder now.

core/src/main/scala/com/pingcap/tispark/write/TiDBOptions.scala
shiyuhang0: yes. you need to convert any configs users may set to tidboption in newWriteBuilder. And these configs should be in your doc

docs/insert_sql_userguide.md
```diff
@@ -0,0 +1,127 @@
...
+      `O_ORDERDATE` date NOT NULL,
+      `O_ORDERPRIORITY` char(15) NOT NULL,
+      `O_CLERK` char(15) NOT NULL,
+      `O_SHIPPRIORITY` int(11) NOT NULL,
+      `O_COMMENT` varchar(79) NOT NULL
+    )
+    */
+    
+   // write data to tidb
+    spark.sql(
```
zhangyangyu, line 40: we don't need catalog?

docs/insert_sql_userguide.md
```diff
@@ -0,0 +1,127 @@
...
+| tidb.addr                   | false    | -               | TiDB address, needed when enableUpdateTableStatistics is true                                                                 |
+| tidb.port                   | false    | -               | TiDB Port, needed when enableUpdateTableStatistics is true                                                                    |
+| tidb.user                   | false    | -               | TiDB User, needed when enableUpdateTableStatistics is true                                                                    |
+| tidb.password               | false    | -               | TiDB Password, needed when enableUpdateTableStatistics is true                                                                |
+| enableRegionSplit           | false    | true            | To split Region to avoid hot Region during insertion                                                                          |
+| scatterWaitMS               | false    | 300000          | Max time to wait scatter region                                                                                               |
```
zhangyangyu, line 59: Why delete doesn't have so many options?

TrafalgarRicardoLu: This options are used in write. Some of them are used to create jdbc client to update statistics, some of them are used to do region splits. 

Delete doesn't do these things. So it doesn't need.

docs/insert_sql_userguide.md
TrafalgarRicardoLu: I‘ll  add it.

docs/insert_sql_userguide.md
```diff
@@ -0,0 +1,130 @@
...
+   //use catalog
+   spark.sql("use tidb_catalog")
+    
+   // write data to tidb
+    spark.sql(
+          "insert into tpch_test.target_table_orders values" +
+            "(1,1,'0',11.11,date'2022-1-1','first','spark',12345,'no comment')")
+
+    // select data
+    val df = spark.sql("select * from tpch_test.target_table_orders")
```
shiyuhang0, line 48:
```suggestion
    spark.sql("select * from tpch_test.target_table_orders").show
```
use show to trigger an action,so that user can see the result

docs/insert_sql_userguide.md
```diff
@@ -0,0 +1,130 @@
...
+   ```
+
+## TiDB Options
+
+The following table shows the TiDB-specific options, which can be passed in through `SparkConf`.
```
shiyuhang0, line 53: password can't be passed by SparkConf, it can only be passed by `spark.conf.set`. they are different for the latter one is session level config, and it will include the configs in SparkConf (`SparkContext.getOrCreate().getConf`)
you need to remind user not set password in SparkConf at least. about other configs, they can set by both the SparkConf and `spark.conf.set`

docs/insert_sql_userguide.md
```diff
@@ -0,0 +1,131 @@
...
+          "insert into tpch_test.target_table_orders values" +
+            "(1,1,'0',11.11,date'2022-1-1','first','spark',12345,'no comment')")
+
+    // select data
+    spark.sql("select * from tpch_test.target_table_orders").show
+   ```
+
+## TiDB Options
+
+The following table shows the TiDB-specific options, which can be passed in through `SparkConf` or `spark.conf.set`.
```
shiyuhang0, line 53:
```suggestion
The following table shows the insert configurations, which can be passed in through `SparkConf` or `spark.conf.set`.
```

docs/insert_sql_userguide.md
```diff
@@ -0,0 +1,131 @@
...
+   // write data to tidb
+    spark.sql(
+          "insert into tpch_test.target_table_orders values" +
+            "(1,1,'0',11.11,date'2022-1-1','first','spark',12345,'no comment')")
+
+    // select data
+    spark.sql("select * from tpch_test.target_table_orders").show
+   ```
+
+## TiDB Options
```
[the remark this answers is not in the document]
shiyuhang0: I think options refer to datasource options. it is better to use configs or Configurations  in spark sql

shiyuhang0 (approved):

core/src/main/scala/com/pingcap/tispark/write/TiDBOptions.scala
qidi1, line 289: Why a new mechanism was introduced instead of following the past practice?https://github.com/pingcap/tispark/blob/cc9c121dab6dff68fbb7ca1cae4c6b662845c7b8/core/src/main/scala/com/pingcap/tispark/write/TiBatchWriteTable.scala#L55-L56

TrafalgarRicardoLu: we decide to remove jdbc client and only use it to update statistics. This option was gotten by jdbc client, so we are not able to get it now. Instead, we put it in TiDBOption and users can set it by `spark.conf.set`.

spark-wrapper/spark-3.0/src/main/scala/org/apache/spark/sql/connector/write/TiDBWriteBuilder.scala
```diff
@@ -36,6 +36,7 @@ case class TiDBWriteBuilder(
           .options(tiDBOptions.parameters)
           .option(TiDBOptions.TIDB_DATABASE, tiDBOptions.database)
           .option(TiDBOptions.TIDB_TABLE, tiDBOptions.table)
+          .option(TiDBOptions.TIDB_DEDUPLICATE, "false")
```
xuanyu66, line 39: Why add this option, is it a bug fix?

TrafalgarRicardoLu: We want

TrafalgarRicardoLu: We want to have same behavior with TiDB.  TiDB will throw an error when there are duplicate keys. If it's `true`, we will handle it instead of throw an error.

qidi1 (approved):

xuanyu66 (approved):

shiyuhang0 (approved):

xuanyu66 (approved):

core/src/main/java/com/pingcap/tikv/datatype/TypeMapping.java
```diff
@@ -90,6 +80,9 @@ public static org.apache.spark.sql.types.DataType toSparkType(DataType type) {
       if (type.isUnsigned() && type.getType() == TypeLonglong) {
         return DataTypes.createDecimalType(20, 0);
       }
+      if (type.getType() == TypeTiny && type.getLength() == 1) {
```
boneanxs, line 83: Hey @TrafalgarRicardoLu It appeals this is a breaking change, can explain more why this change is required?

zhangyangyu: I think no. IIRC MySQL doesn't have boolean type before, so everyone use 1 bit tinyint to represent a boolean type. Many MySQL ecosystem softwares still streat this as bool.

boneanxs: Yea, understand this is trying to align with others databases, but for someone who rely on this before, after upgrading to the new version, they possibly meet issues. Just an advice, maybe we can communicate such changes in the changelog to alert users who relied on the previous behavior.

zhangyangyu: Yeah, reasonable. We could do better. Does this change affect you?

boneanxs: Yea, we actually use this to ingest data from TiDB to lakehouse through HUDI, and because of this schema change, the target table could meet schema incompatible issues.

PR #2489, opened 2022-08-04
Only do auth check for tables in TiDB

TrafalgarRicardoLu:
Only do auth check for tables in TiDB.

We will check the catalog name of operation. If it doesn't use `tidb_catalog`, we won't do auth check for it.

### Check List 

Tests 
 - Unit test
 - Integration test


shiyuhang0:
It's better to add some test

xuanyu66 (changes requested):
Add some IT. https://github.com/pingcap/tispark/blob/master/core/src/test/scala/com/pingcap/tispark/auth/TiAuthIntegrationSuite.scala

core/src/test/scala/com/pingcap/tispark/auth/TiAuthIntegrationSuite.scala
```diff
@@ -79,10 +78,20 @@ class TiAuthIntegrationSuite extends SharedSQLContext {
     tidbStmt.execute(s"DROP DATABASE IF EXISTS `$dummyDatabase`")
     super.afterAll()
     _isAuthEnabled = false
+    _isHiveEnabled = false
     TiAuthorization.enableAuth = false
   }
 
+  test("Operator on hive table should pass") {
```
xuanyu66, line 85: move methods before afterAll()

shiyuhang0 (approved):

Daemonxiao (approved):

xuanyu66 (approved):

PR #2532, opened 2022-09-01
Fix cacheInvalidateCallback does not work ref #1209

qidi1:
Signed-off-by: qidi1 <1083369179@qq.com>

The original process is shown as below.
```scala
  CacheInvalidateListener
    .initCacheListener(sparkSession.sparkContext, tiSession.getRegionManager)
  tiSession.injectCallBackFunc(CacheInvalidateListener.getInstance())
```
Firstly, call the `getRegionManager` method, `regionManager` is initialized in singleton mode. 
```java
  public synchronized RegionManager getRegionManager() {
    RegionManager res = regionManager;
    if (res == null) {
      synchronized (this) {
        if (regionManager == null) {
          regionManager = new RegionManager(getPDClient(), this.cacheInvalidateCallback);
        }
        res = regionManager;
      }
    }
    return res;
  }
```
Then, call the `injectCallBackFunc` method and initialize `cacheInvalidateCallback`,  but `regionManager` is already initialized. So the `cacheInvaildCallback` cannot be passed into the `regionManager`.
```java
  public void injectCallBackFunc(Function<CacheInvalidateEvent, Void> callBackFunc) {
    this.cacheInvalidateCallback = callBackFunc;
  }
```
Add `setCacheInvalidateCallback` method in `RegionManager` , and call the method in `injectCallBackFunc`.

### Check List 

Tests 

 - Integration test

### DIFF
--- a/core/src/test/scala/com/pingcap/tispark/accumulator/AccumulatorSuite.scala
+++ b/core/src/test/scala/com/pingcap/tispark/accumulator/AccumulatorSuite.scala
@@ -1,22 +1,38 @@
+/*
+ *
+ * Copyright 2022 PingCAP, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
 package com.pingcap.tispark.accumulator;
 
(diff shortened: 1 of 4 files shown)


core/src/test/scala/com/pingcap/tispark/accumulator/AccumulatorSuite.scala
```diff
@@ -32,17 +32,18 @@ class AccumulatorSuite extends BaseTiSparkTest {
...
+      tidbStmt.execute(
+        """
           |CREATE TABLE `t1` (
           |`a` BIGINT(20)  NOT NULL,
           |`b` varchar(255) NOT NULL,
           |`c` varchar(255) DEFAULT NULL
           |) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
               """.stripMargin)
-      spark.sql("SELECT * FROM t1 where a>0 and b > 'aa'").show()
+      spark.sql("SELECT * FROM t1").show()
```
Daemonxiao, line 42: Why delete the conditional query？

PR #2486, opened 2022-08-03
upgrade spark

shiyuhang0:
### What problem does this PR solve? 

upgrade spark version to avoid some vulnerabilities

image

PR #2544, opened 2022-09-15
Bump spark version to avoid vulnerabilities

shiyuhang0:
This pr is a supplement to https://github.com/pingcap/tispark/pull/2486. It bumps the Spark version：
- from 3.0.2 to 3.0.3
- from 3.1.1 to 3.1.3
- from 3.2.1 to 3.2.2

Note that spark 3.0.3 still has vulnerabilities


zhangyangyu (approved):

xuanyu66 (changes requested):
https://github.com/pingcap/tispark/blob/master/pom.xml#L183-L209
Bump these too

xuanyu66:
And the better solution is to define variables so that we won't miss the places need to be modified

Daemonxiao (approved):

shiyuhang0:
> https://github.com/pingcap/tispark/blob/master/pom.xml#L183-L209 Bump these too

has done in https://github.com/pingcap/tispark/pull/2486

pom.xml
```diff
@@ -73,6 +73,10 @@
         <spark.version.compile>3.0.3</spark.version.compile>
         <spark.version.test>3.0.3</spark.version.test>
         <spark.version.release>3.0</spark.version.release>
```
zhangyangyu, line 75: still need or not?

shiyuhang0: we need it for it is the default value

shiyuhang0: I have optimized it

xuanyu66 (approved):

PR #2545, opened 2022-09-16
Support Auto Random

qidi1:
Signed-off-by: qidi1 <1083369179@qq.com>

## What problem does this PR solve?
This pr is a backport feature from TiBigData about https://github.com/tidb-incubator/TiBigData/issues/235.

TiBigData support auto random feature list.

[Design](https://github.com/tidb-incubator/TiBigData/pull/236)
[Code](https://github.com/tidb-incubator/TiBigData/pull/234)

### Differences from TiBigData implementation
Unlike the TiBigData design, we do not use the configuration (`tikv.sink.ignore-autorandom-column-value`) to explicitly indicate whether to ignore the user-specified value of the auto random column, but follow the auto increment design, where only in update mode the auto random column can only be specified.

## Check List
Tests

Integration test


core/src/main/scala/com/pingcap/tispark/write/TiBatchWriteTable.scala
```diff
@@ -111,94 +114,12 @@ class TiBatchWriteTable(
...
-          }
-          Row.fromSeq(rowSep)
-        }
-      }
+    val rdd = if (tiTableInfo.hasAutoIncrementColumn && isNeedAllocateIdForAutoIncrementCol(df)) {
+      // auto increment
+      allocateIdForAutoIncrementCol(count, startTimeStamp)
+    } else if (tiTableInfo.hasAutoRandomColumn && isNeedAllocateIdForAutoRandomCol(df)) {
+      // auto random
+      allocateIdForAutoRandomCol(count, startTimeStamp)
```
shiyuhang0, line 122: Can we merged allocateIdForAutoIncrementCol and allocateIdForAutoRandomCol into one method with `type` parameter.

qidi1: I try to do so.PLAT

Daemonxiao (approved):

shiyuhang0 (approved):

xuanyu66 (approved):

PR #2524, opened 2022-08-27
Support new Collation

Daemonxiao:
### What problem does this PR solve? 
Support new collation refers to #2444.
fix #2236.
In this pr, TiSpark support all utf8 and utf8mb4's new collation, such as xx_bin, xx_general_ci and xx_unicode_ci.

1. When new collation is enabled, we turn the collation id to negative so that other cluster components (for example, TiKV) can be aware of it without any change to the protocol definition.
```
  private FieldType toPBFieldType(DataType fieldType) {
    return FieldType.newBuilder()
        .setTp(fieldType.getTypeCode())
        .setFlag(fieldType.getFlag())
        .setFlen((int) fieldType.getLength())
        .setDecimal(fieldType.getDecimal())
        .setCharset(fieldType.getCharset())
        .setCollate(Collation.rewriteNewCollationIDIfNeeded(fieldType.getCollationCode())) // we set new collation here
        .build();
  }
```

2. When new collation is enabled, reset the return type in some expression.  Now, we only change collation in LogicalBinaryExpression, ComparisonBinaryExpression, and StringRegExpression which are related to comparing.
```
  private Expr.Builder scalarToPartialProto(Expression node, Object context) {
...
    // Return type
    if (Collation.isNewCollationEnabled()) {
      node.setNewCollation();        // change return collation.
    }
    builder.setFieldType(toPBFieldType(getType(node)));
...
    return builder;
  }
```

3. When new collation is enabled, encode restore data to the index value. 

The column is a string type that uses the new collation. 
* For type char, blob and unspecified, the collation is neither binary nor with the suffix “_bin”.
* For type varchar, the collation is not binary.
```
private static byte[] genIndexValueForCommonHandleVersion1(
      Row row, Handle handle, boolean distinct, TiIndexInfo tiIndexInfo, TiTableInfo tiTableInfo) {
...
    // encode restore data if needed.
    List<TiColumnInfo> columnInfoList = new ArrayList<>();
    List<Object> valueList = new ArrayList<>();
    for (TiIndexInfo index : tiTableInfo.getIndices()) {
      for (TiIndexColumn tiIndexColumn : index.getIndexColumns()) {
        TiColumnInfo indexColumnInfo = tiTableInfo.getColumn(tiIndexColumn.getOffset());
        DataType indexType = indexColumnInfo.getType();
        int prefixLength = (int) tiIndexColumn.getLength();
        if (needRestoreData(indexType)) {
[...]

PR #2565, opened 2022-10-12
Partition write support new collations

xuanyu66:

In https://github.com/pingcap/tispark/pull/2524, TiSpark supports new collation.
According to [Range column partition](https://dev.mysql.com/doc/refman/5.7/en/partitioning-columns-range.html#:~:text=Range%20columns%20partitioning%20is%20similar,types%20other%20than%20integer%20types), partitioning also needs to obey the sort orders of collation.

- For range column partition,  using different sort orders according to collation

### Check List 

Tests 

 - Integration test

Related changes

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/RangePartitionLocator.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/RangePartitionLocator.java
@@ -26,9 +26,14 @@ import com.pingcap.tikv.expression.Expression;
 import com.pingcap.tikv.expression.FuncCallExpr;
 import com.pingcap.tikv.expression.LogicalBinaryExpression;
 import com.pingcap.tikv.expression.LogicalBinaryExpression.Type;
+import com.pingcap.tikv.meta.Collation;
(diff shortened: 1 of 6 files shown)


tikv-client/src/main/java/com/pingcap/tikv/expression/visitor/RangePartitionLocator.java
```diff
@@ -59,6 +66,7 @@ public Boolean visit(ComparisonBinaryExpression node, PartitionLocatorContext co
         Expression expression = left.getChildren().get(0);
         ColumnRef columnRef = (ColumnRef) expression;
         columnRef.resolve(tableInfo);
+        dataType = IntegerType.BIGINT;
```
Daemonxiao, line 69: Why set YEAR as IntergerType? Shall we add an annotation here?

xuanyu66: function year() return integer

shiyuhang0 (approved):

Daemonxiao (approved):

PR #2300, opened 2022-04-13
DELETE table statistics

xuanyu66:
### What problem does this PR solve? 

Delete unused configuration `spark.tispark.statistics.auto_load`
Delete related document

PR #2259, opened 2022-03-03
Move the dependency of API v1 in Catalyst

shiyuhang0:
### What problem does this PR solve? 
https://github.com/pingcap/tispark/issues/2242

### What is changed and how it works?
- Add TiDBTable
- Add TiDBTableProvider and delete TiDBDataSource
- Replace `TiDBRelation` to `TiDBTable`
- Delete some catalyst extension because of useless now
- fix some test
- change code structure

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

The following usage is no longer supported, because is a datasource v1 read/write path
```
sprak.sql("use spark_catalog")  
spark.sql("create table xxx using tidb")
spark.sql("select * from xxx") 
spark.sql("insert into xxx")  
```
- use tidb_catalog to select instead
- use write api to insert instead

PR #2578, opened 2022-11-03
fix statistics

### LINKED ISSUE #2573: tispark 3.x version read speed is slower than tispark 2.5
**tidb version is 5.4.2
spark version 3.0 to 3.2
tispark version 3.0 to 3.1**
I used tispark2.5 before, but when i upgrade tispark version to 3.x ,I find it is slow when read tikv than tispark 2.5 version。

table info :

image

spark conf ：

image

tispark 2.5:

image

tispark 3.x:

image

xuanyu66:

close #2573
In https://github.com/pingcap/tispark/pull/2259, TiResolutionRule was deleted, which cause statistics would not be collected.
Thus in https://github.com/pingcap/tispark/pull/2300/files, `ENABLE_AUTO_LOAD_STATISTICS` was considered to be useless and deprecated.

- add a new ResolutionRule to collect statistics if `ENABLE_AUTO_LOAD_STATISTICS` is true.
- add documents to explain `ENABLE_AUTO_LOAD_STATISTICS`.
- fix the logicalPlan result.

### Check List 

Tests 

 - Unit test

Related changes

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/TiContext.scala
+++ b/core/src/main/scala/org/apache/spark/sql/TiContext.scala
@@ -68,6 +68,8 @@ class TiContext(val sparkSession: SparkSession) extends Serializable with Loggin
   clientSession.injectCallBackFunc(CacheInvalidateListener.getInstance())
   val meta: MetaManager = new MetaManager(clientSession.getCatalog)
   val debug: DebugTool = new DebugTool
+  val autoLoad: Boolean =
+    conf.getBoolean(TiConfigConst.ENABLE_AUTO_LOAD_STATISTICS, defaultValue = true)
 
   // add backtick for table name in case it contains, e.g., a minus sign
   private def getViewName(dbName: String, tableName: String, dbNameAsPrefix: Boolean): String =
--- a/.github/workflows/alter-primary-key-false-test.yml
+++ b/.github/workflows/alter-primary-key-false-test.yml
@@ -39,4 +39,4 @@ jobs:
         run: mvn clean package -Dmaven.test.skip=true -B
 
       - name: test
-        run: mvn test -am -pl core -Dtest=moo -DwildcardSuites=org.apache.spark.sql.catalyst.plans.logical.LogicalPlanTestSuite,com.pingcap.tispark.datasource.AutoRandomSuite -DfailIfNoTests=false
+        run: mvn test -am -pl core -DtrimStackTrace=false -Dtest=moo -DwildcardSuites=org.apache.spark.sql.catalyst.plans.logical.LogicalPlanTestSuite,com.pingcap.tispark.datasource.AutoRandomSuite -DfailIfNoTests=false
--- a/.github/workflows/alter-primary-key-false-test.yml
+++ b/.github/workflows/alter-primary-key-false-test.yml
@@ -39,4 +39,4 @@ jobs:
         run: mvn clean package -Dmaven.test.skip=true -B
 
       - name: test
-        run: mvn test -am -pl core -DtrimStackTrace=false -Dtest=moo -DwildcardSuites=org.apache.spark.sql.catalyst.plans.logical.LogicalPlanTestSuite,com.pingcap.tispark.datasource.AutoRandomSuite -DfailIfNoTests=false
+        run: mvn test -am -pl core -Dtest=moo -DwildcardSuites=org.apache.spark.sql.catalyst.plans.logical.LogicalPlanTestSuite,com.pingcap.tispark.datasource.AutoRandomSuite -DfailIfNoTests=false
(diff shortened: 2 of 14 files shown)


shiyuhang0 (approved):

core/src/main/scala/org/apache/spark/sql/TiContext.scala
```diff
@@ -68,6 +68,8 @@ class TiContext(val sparkSession: SparkSession) extends Serializable with Loggin
   clientSession.injectCallBackFunc(CacheInvalidateListener.getInstance())
   val meta: MetaManager = new MetaManager(clientSession.getCatalog)
   val debug: DebugTool = new DebugTool
+  val autoLoad: Boolean =
+    conf.getBoolean(TiConfigConst.ENABLE_AUTO_LOAD_STATISTICS, defaultValue = true)
```
Daemonxiao, line 72: Shall we use `true` as the default value?

xuanyu66: Yes. For the former version, `true` is the default value.
And for TiDB, you can not disable statistics. https://docs.pingcap.com/tidb/dev/statistics

Daemonxiao (approved):

wfxxh:
Is there any cache in tispark ? When i use this PR to run my process,the tikv read task number is correct,same as table region number.If I only process one table it is success.But if I loop process several tables it cause executor OOM。

**loop process several tables:**

image

image


**process the oom table when the loop is processed separately**

image

**my process**

image

xuanyu66:
@wfxxh Yes, the cache is in com.pingcap.tispark.statistics.StatisticsManager
image

wfxxh:
@xuanyu66 I does not think the cache of statisticsMap can cause the executor OOM. I print the statisticsMap size .
image
And I do not know the reasion , the executor log say nothing

xuanyu66:
@wfxxh OK, I will merge this PR. If there is any update, you can open a new issue and we figure it out then

wfxxh:
@xuanyu66  But the reasion when loop read tikv cause OOM is not find,It is may be a bug

xuanyu66:
@wfxxh Can you check if OOM will happen when using the released version?

wfxxh:
I have checked the v2.5.1 ,v3.1.1 and the master version.
The v2.5.1 version read task number is the same as this PR and equal to the table region number,but will cause OOM when loop read several table.
The v3.1.1 and the master version read task number is not correct,and equal to the table index region number.
The v2.5.1 version and the PR version is fast than v3.1.1 and master

xuanyu66:
@wfxxh So 2.5.1 and this PR will OOM while 3.1.1 and master don't have the OOM problem, am I right?

wfxxh:
@xuanyu66  Yes,but even it cause OOM ,there will start a new executor pod,the spark job final status is success,and it was faster than the 3.1.1 version

xuanyu66:
@wfxxh Can we dump the memory snapshot to analyze it?

wfxxh:
@xuanyu66 Yes I am trying to do it ,but I am run spark on k8s , I should mount a storage

wfxxh:
@xuanyu66 I am so sorry , I can not do it ,because when the executor pod OOM it will be deleted,and the storage pvc will deleted too

xuanyu66:
@wfxxh Can you increase your memory, and hold on for 10 minutes when the task is running. And dump the memory to analyze

wfxxh:
@xuanyu66 How to hold on the process? sleep the thread? But the time when i sleep it may by not oom

xuanyu66:
@wfxxh use "-XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath="c:\temp\dump2.hprof" "
and use the (hostpaht)[https://kubernetes.io/docs/concepts/storage/volumes/#hostpath]

wfxxh:
@xuanyu66 Yes,I have try this method ,but when the executor pod OOM and deleted,it will start a new pod ,and this will clear the hostpath dir.
image

wfxxh:
@xuanyu66 I think you can merge this PR,because even it may be cause the executor oom ,but the final status is success,and it is faster

xuanyu66:
@wfxxh You can simply use  "-XX:HeapDumpPath="c:\temp-${randomstring}\dump2.hprof"

wfxxh:
@xuanyu66 I do not konw what is your meaning. I say when the pod oom deleted ,and restart a new one ,the hostpath in host machine will be deleted too.

xuanyu66:
@wfxxh it's strange, normally hostpath will not be deleted. I used to build a program to collect all oom dump in last company,  
and also use hostpath.

xuanyu66:
@wfxxh I have merged this PR and we will discuss the OOM problem here https://github.com/pingcap/tispark/issues/2587

PR #2593, opened 2022-11-24
Support partition pruning with to_days function

shiyuhang0:

TiSpark only supports partition pruning with
- column expression
- year (col)

To_DAYS(col) needs to be supported.

The DDL may like this:
```
CREATE TABLE `pt_todays_date` (
  `id` int(11) DEFAULT NULL,
  `name` varchar(50) DEFAULT NULL,
  `purchased` date DEFAULT NULL,
  index `idx_id`(`id`)
PARTITION BY RANGE (to_days(purchased)) (
  PARTITION p0 VALUES LESS THAN (to_days('1995-01-02')),
  PARTITION p1 VALUES LESS THAN (to_days('2000-01-02')),
  PARTITION p2 VALUES LESS THAN (732313),
  PARTITION p3 VALUES LESS THAN (MAXVALUE)
)
```

**Support To_DAYS(col)**
1. Alter `Astbuilder` to support convert partExprStr (contains TO_DAYS) to Expression with AstBuilder
2. Add `TO_DAYS` func in `FuncCallExpr` and add the method to deal with to_days in `FuncCallExprEval.java` (calculate the literal in where clause to the result of to_days)
3. Support `TO_DAYS` in `PartAndFilterExprRewriter.java`
4. Enhancement `Converter.java` to support converting int to date and DateTime. So, TiSpark can judge if the filter condition is included in the partition table. (Bad Design, why don't just use int to judge? I just follow the original design)

 
**Fix the bug that year(col) throw exception when col is datetime type**

- Enhancement `FuncCallExprEval.java` to support converting java.sql.Timestamptimestamp(datetime type in TiDB) to datetime
- Enhancement `Converter.java` to support converting int to dateTime

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/expression/FuncCallExprEval.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/expression/FuncCallExprEval.java
@@ -22,6 +22,10 @@ import com.pingcap.tikv.types.DateTimeType;
 import com.pingcap.tikv.types.DateType;
 import com.pingcap.tikv.types.IntegerType;
 import com.pingcap.tikv.types.StringType;
+
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.function.Function;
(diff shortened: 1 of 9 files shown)


tikv-client/src/main/java/com/pingcap/tikv/types/Converter.java
```diff
@@ -262,6 +262,14 @@ public static ExtendedDateTime convertToDateTime(Object val) {
       } catch (Exception e) {
         throw new TypeException(String.format("Error parsing string %s to datetime", val), e);
       }
+    } else if (val instanceof Integer) {
+      if ((int) val < 719528) {
+        return new ExtendedDateTime(new DateTime((int) val, 1, 1, 0, 0));
```
xuanyu66, line 267: add comment to explain what is this value

tikv-client/src/main/java/com/pingcap/tikv/expression/FuncCallExprEval.java
```diff
@@ -61,4 +80,12 @@ static Function<Constant, Constant> getEvalFn(Type tp) {
     }
     return null;
   }
+
+  private static int dateTime2ToDays(DateTime date) {
+    // the number of day from 0000-00-00 to 1970-01-01
+    int var1 = 719528;
```
xuanyu66, line 86: change to constant

xuanyu66 (changes requested):

zhangyangyu (approved):

xuanyu66 (approved):

PR #2619, opened 2023-02-02
Support probe tiflash status

shiyuhang0:

Support probe TiFlash status.

When you have two TiFlash and suspend one during the job, the suspended TiFlash can still be chosen and the request to it will wait for a timeout for grpc. It wastes resources and affected the whole job's speed.

When we choose the TiFlash. we call the isalive to probe the status of TiFlash. If it is not ready, drop the task.

### Alternatives

In this pr, a background thread poll will refresh the cache of TiFlash store status every time a region task has been generated. Then drop the task whose store is unavailable. 
The cons are if there are many tasks, the thread poll will be distributed too much probe tasks.

Another way is to use a corn job to refresh the cache:
1. Define a Storeclient/HTTP that need not the region. (RegionStoreClien need region which we can not provide)
2. Then use pdclient to query all the stores and choose the TiFlash stores.
3. Use Storeclient/HTTP  to call is probe method to detect the tiflash status.


tikv-client/src/main/java/com/pingcap/tikv/TiConfiguration.java
```diff
@@ -42,7 +42,7 @@
 @Accessors(chain = true)
 public class TiConfiguration implements Serializable {
   private static final DateTimeZone DEF_TIMEZONE = Converter.getLocalTimezone();
-  private static final int DEF_TIMEOUT = 10;
+  private static final int DEF_TIMEOUT = 3;
```
zhangyangyu, line 45: Fix the doc too plz.

tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
```diff
@@ -349,6 +349,24 @@ public ExecutorService getThreadPoolForDeleteRange() {
...
+                  20,
+                  0,
+                  TimeUnit.MILLISECONDS,
+                  new ArrayBlockingQueue<>(1),
+                  new ThreadPoolExecutor.DiscardPolicy());
+        }
+      }
+    }
+    return isAliveThreadPool;
+  }
```
zhangyangyu, line 368: Some questions:
1. one pool for one session?
2. is 20 threads necessary?
3. the blocking queue has only 1 slot, is it right?

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/DAGIterator.java
```diff
@@ -269,4 +279,29 @@ private Iterator<SelectResponse> processByStreaming(RangeSplitter.RegionTask reg
       throw new TiClientInternalException("Error Closing Store client.", e);
     }
   }
+
+  public Boolean isMppStoreAlive(long id, RegionStoreClient client) {
```
zhangyangyu, line 283: comment to add link to this PR. I think the PR description is valuable.

tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
shiyuhang0: 1. Yes. one pool for one session, just like other pool in Tisession `indexScanThreadPool`,tableScanThreadPool etc.
2. There may be too many probe tasks. If there is only one thread, the latest task may be discarded.
3. The task execution speed may be slower than the task generated speed. I use one slot so that the poll can discard the tasks. If we use too many slots, The task may be outdated. We just need the latest task.
4. The same as 3. Newfixedthreadpool's blocking queue has too many slots.

all in all. it may be difficult to find the best parameters for the poll. If this change can not meet our need. I will consider using a corn job. (may use newScheduledThreadPool  in java)

docs/userguide_3.0.md
```diff
@@ -234,49 +234,49 @@ spark.sql("select t1.id,t2.id from spark_catalog.default.t t1 left join tidb_cat
...
+| `spark.tispark.grpc.framesize`                 | `2147483647`     | The maximum frame size of gRPC response in bytes (default 2G)                                                                                                                                                                                                                                                                                                                                                                                      |
+| `spark.tispark.grpc.timeout_in_sec`            | ``               | The gRPC timeout time in seconds. We will use 180s if it not set.                                                                                                                                                                                                                                                                                                                                                                                  |
```
zhangyangyu, line 241: Why not set the default field?

zhangyangyu (dismissed):

zhangyangyu (dismissed):

tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
```diff
@@ -39,8 +39,7 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
+import java.util.concurrent.*;
```
xuanyu66, line 42: https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports
Disable wildcard imports

tikv-client/src/main/java/com/pingcap/tikv/TiSession.java
```diff
@@ -349,6 +349,28 @@ public ExecutorService getThreadPoolForDeleteRange() {
     return res;
   }
 
+  public ExecutorService getThreadPoolForIsAlive() {
+    if (isAliveThreadPool == null) {
+      synchronized (this) {
+        if (isAliveThreadPool == null) {
+          isAliveThreadPool =
+              new ThreadPoolExecutor(
+                  20,
```
xuanyu66, line 358: Why use 20?
I think it should be associated with core numbers.
Runtime.getRuntime().availableProcessors();

tikv-client/src/main/java/com/pingcap/tikv/operation/iterator/DAGIterator.java
```diff
@@ -203,6 +202,8 @@ private SelectResponse process(RangeSplitter.RegionTask regionTask) {
     HashSet<Long> resolvedLocks = new HashSet<>();
     // In case of one region task spilt into several others, we ues a queue to properly handle all
     // the remaining tasks.
+    // Do we need to set the timeout for this loop? If all the stores are suspended, it will not
```
xuanyu66, line 205: What's the conclusion

shiyuhang0: we can open another pr to solve if we need. This problem has always existed before, not introduced by this pr.

zhangyangyu (approved):

zhangyangyu (approved):

zhangyangyu (approved):

xuanyu66 (approved):

PR #2632, opened 2023-02-24
Check transaction start_ts < gc_safe_point

shiyuhang0:

https://github.com/tikv/client-java/issues/721

Add a check after all the transaction requests, but before they handle the response.
In this way: once transaction start_ts < gc_safe_point, it will throw the ` start_ts < gc_safe_point` exception first even if it has other error returned by TiKV.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

### DIFF
--- a/tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
+++ b/tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
@@ -1213,7 +1213,7 @@ public class RegionStoreClient extends AbstractRegionStoreClient {
     Long safePoint = pdClient.getGCSafePoint(bo);
     if (startTs < safePoint) {
       throw new GrpcException(
-          "start_ts < gc_safe_point. Please ensure the execute time is less than the tidb_gc_life_time https://docs.pingcap.com/tidb/stable/system-variables#tidb_gc_life_time-new-in-v50. Or you may get the unexpected data.");
+          "start_ts < gc_safe_point. To avoid this, you can set the tidb_gc_life_time https://docs.pingcap.com/tidb/stable/system-variables#tidb_gc_life_time-new-in-v50 bigger than TiSpark execute time");
     }
   }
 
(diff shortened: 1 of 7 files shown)


xuanyu66 (dismissed):

tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
```diff
@@ -1202,6 +1207,16 @@ private void rawDeleteRangeHelper(RawDeleteRangeResponse resp) {
     }
   }
 
+  // checkStartTs will check the start_ts > gc_safe_point to ensure read correctly
```
zhangyangyu, line 1210:
```suggestion
  // checkStartTs will check the start_ts >= gc_safe_point to ensure read correctly
```

tikv-client/src/main/java/com/pingcap/tikv/region/RegionStoreClient.java
```diff
@@ -1202,6 +1207,16 @@ private void rawDeleteRangeHelper(RawDeleteRangeResponse resp) {
...
+  public void checkStartTs(long startTs) {
+    BackOffer bo = ConcreteBackOffer.newCustomBackOff(BackOffer.PD_INFO_BACKOFF);
+    Long safePoint = pdClient.getGCSafePoint(bo);
+    if (startTs < safePoint) {
+      throw new GrpcException(
```
zhangyangyu, line 1215: I think "execute time is less than gc life time" is not accurate. I don't have to be.

shiyuhang0: But it is hard for users to know the true safe point. One thing that can ensure is:
1. now- safe point >= gc_life_time
2. transaction execute time = now-start_ts
3. transaction execute time < spark execute time
if  gc_life_time > spark execute time  we can get:
now- safe point >= gc_life_time > spark execute time >transaction execute time=now-start_ts
=> now- safe point > now-start_ts => start_ts > safe point

shiyuhang0: any better error message for this exception?

zhangyangyu (dismissed):

shiyuhang0:
close this PR because asking PD for every KV request may increase the pressure of PD. 
We'd like to implement this when PD has some mechanisms to avoid this, like a lease

PR #2634, opened 2023-02-27
Support server safe point

shiyuhang0:

TiSpark does not have a mechanism to check GC. So, you may get the wrong data if your transaction's start_ts behind safe_point

This PR introduced the server safe point in TiSpark. TiKV will use the min server safe point among all servers (including TiDB).

TiSpark will generate a new serverId for every spark session. Every spark session will register the min start_ts as its' server safe point to PD. Thus, the safe point will not exceed the min start_ts among all spark applications.

### Test

It is hard to write IT. Here are the results I test locally.

1. PD logs show TiSpark register server safe point successfully

image

2. TiKV metric shows TiSpark server safe point can pause the safe point: TiSpark register 439743565061947392 (17:02:55) as safe point. This makes the GC safe point turns from 16:57:39 to 17:02:55 rather than 17:07:39

image

3.  TiKV metric shows safe point will recovery after the TiSpark finished: The GC safe point turns from 17:02:55 to 17:18:39. which means TiSpark will not block it anymore.

image

4. When the start_ts have already less than the current safe point. TiSpark will throw exceptions directly

3MUDaPC8Vq

### DIFF
--- a/core/src/main/scala/com/pingcap/tispark/savepoint/ServerSavePoint.scala
+++ b/core/src/main/scala/com/pingcap/tispark/savepoint/ServerSavePoint.scala
@@ -1,0 +1,50 @@
+package com.pingcap.tispark.savepoint
+
+import com.pingcap.tikv.TiSession
+import com.pingcap.tikv.exception.TiInternalException
+import com.pingcap.tikv.util.{BackOffer, ConcreteBackOffer}
+import org.slf4j.LoggerFactory
+
+import java.util.concurrent.{Executors, ScheduledExecutorService, TimeUnit}
+
+case class ServerSavePoint (serverId: String, ttl: Long, tiSession: TiSession){
+
+  private final val logger = LoggerFactory.getLogger(getClass.getName)
+  private var minStartTs = Long.MaxValue
+  val service: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor()
+  service.scheduleAtFixedRate(
+    ()=>{
+      if (minStartTs != Long.MaxValue){
+        val savePoint = tiSession.getPDClient.UpdateServiceGCSafePoint(serverId,ttl,minStartTs,ConcreteBackOffer.newCustomBackOff(BackOffer.PD_INFO_BACKOFF))
+        if (savePoint > minStartTs){
+          println("x")
+          logger.error(s"Failed to register server GC safe point because the current minimum safe point $savePoint is newer than what we assume $minStartTs. Maybe you delete the TiSpark safe point in PD")
+        }else{
+          println("x")
+          logger.info(s"register server GC safe point $minStartTs success.")
+        }
+      }
+    },
+    0,
+    1,
+    TimeUnit.MINUTES
+  )
+
+  def updateStartTs(starTs:Long): Unit = {
+    if (starTs < minStartTs){
+      minStartTs = starTs
+      checkServerSafePoint()
+    }
+  }
+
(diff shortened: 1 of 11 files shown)


core/src/main/scala/com/pingcap/tispark/savepoint/ServerSavePoint.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  private final val logger = LoggerFactory.getLogger(getClass.getName)
+  private var minStartTs = Long.MaxValue
+  val service: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor()
+  service.scheduleAtFixedRate(
+    ()=>{
+      if (minStartTs != Long.MaxValue){
+        val savePoint = tiSession.getPDClient.UpdateServiceGCSafePoint(serverId,ttl,minStartTs,ConcreteBackOffer.newCustomBackOff(BackOffer.PD_INFO_BACKOFF))
+        if (savePoint > minStartTs){
+          println("x")
```
Daemonxiao, line 20: Is it a hint for yourself?

shiyuhang0: forget to delete it.

core/src/main/scala/org/apache/spark/sql/TiContext.scala
```diff
@@ -49,11 +52,15 @@ class TiContext(val sparkSession: SparkSession) extends Serializable with Loggin
...
   lazy val sqlContext: SQLContext = sparkSession.sqlContext
+  val defaultTiSparkGCSafePointTTL: Int = 5 * 60
+  val serviceSavePoint: ServiceSavePoint =
+    ServiceSavePoint("tispark_" + UUID.randomUUID, defaultTiSparkGCSafePointTTL, tiSession)
 
   sparkSession.sparkContext.addSparkListener(new SparkListener() {
     override def onApplicationEnd(applicationEnd: SparkListenerApplicationEnd): Unit = {
       if (tiSession != null) {
         try {
+          serviceSavePoint.stopRegisterSavePoint()
```
zhangyangyu, line 63: if the application end, why not remove the minstartts instantly?

core/src/main/scala/com/pingcap/tispark/savepoint/ServiceSavePoint.scala
```diff
@@ -0,0 +1,58 @@
...
+          // will not happen unless someone delete the TiSpark service safe point in PD compulsively
+          logger.error(
+            s"Failed to register service GC safe point because the current minimum safe point $savePoint is newer than what we assume $minStartTs. Maybe you delete the TiSpark safe point in PD")
+        } else {
+          logger.info(s"register service GC safe point $minStartTs success.")
+        }
+      }
+    },
+    0,
+    1,
```
zhangyangyu, line 33: ttl should be large than this interval

core/src/main/scala/org/apache/spark/sql/TiContext.scala
shiyuhang0: good suggestion. But we can not remove it for there is no API. we can set a Long.MAXVALUE

zhangyangyu (dismissed):

zhangyangyu (dismissed):

zhangyangyu (approved):

xuanyu66 (approved):

PR #2639, opened 2023-03-06
Fix read from range column partition table error 

### LINKED ISSUE #2638: [BUG] Read from range column partition table error
**Describe the bug**

Read from range column partition table error.

**What did you do**
1. DDL
```
create table range_column_test (id varchar(10)) partition by RANGE COLUMNS(`id`) (PARTITION `p1` VALUES LESS THAN ('CN001'),PARTITION `p2` VALUES LESS THAN ('CN002'))
```

2. use tispark to read

```
select * from range_column_test where id = 'CN001'
```

3. error occurs
image

**Spark and TiSpark version info**
TiSpark : <= v3.2.0

shiyuhang0:

close #2638

1. Fix the output of explain: TiSpark no more throw `NoSuchElementException` when there is no partition after partition prune.

2. Fix the part of the bug that no partitions will be selected in the range column partition table.

 TiSpark will contain `''` when generates partition exprs. For example, if your DDL is `PARTITION p1 VALUES LESS THAN ('CN001')`, you will get something like `${column_name} less than "'CN001'"`. Then, your filter can not match the partition exprs.

After Fix, TiSpark will convert
-  'CN001' -> "CN001"
- '''CN001''' -> "'CN001'"


tikv-client/src/main/java/com/pingcap/tikv/expression/PartitionPruner.java
```diff
@@ -132,4 +139,35 @@ private static String wrapColumnName(String columnName) {
       return String.format("`%s`", columnName);
     }
   }
+
+  /**
+   * Spark SQL will parse string literal without escape, So we need to parse partition definition
+   * without escape too. wrapValue will replace the first '' to "", so that antlr will not regard
+   * the first '' as a part of string literal. wrapValue will also delete the escape character in
+   * string literal. e.g. 'string' -> "string" '''string''' -> "'string'" 'string''' -> "string'"
```
xuanyu66, line 147: `'string''' -> "string'"`  looks strange

xuanyu66 (changes requested):

Daemonxiao (approved):

xuanyu66 (approved):

PR #1623, opened 2020-09-17
Fix encoder v2 with type convertion and set default batch write version to v2

birdstorm:
Fix #1622 

Also fixes similar problems in decimal/string/enum conversions.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to be included in the release note

PR #2667, opened 2023-03-23
Do not reload table schema when update catalog cache

shiyuhang0:

TiSpark will reload all table schemas when init catalogCache. It will also reload all table schemas once we call listTable or getTable with the invalid cache.

One user reports that TiSpark hangs with more than 10000+ table schemas.

Add a configuration `spark.tispark.load_tables` , default is true.
when set it to false, we will not load all tables when we reload the cache. table cache will only be loaded when the table is used.

### Check List 

Tests 

 - I have tested it locally and it works.
 - User has tested and it works.


shiyuhang0:
@birdstorm It is a simple PR. 

I think we can just reload db cache when the schema version has changed. GetTable or ListTable method will find the table cache is empty, then just load all tables in its' own db rather than all dbs. These will reduce the pressure on users with a large number of tables.

I find you use relaod(true) which will reload all table scehmas in the following PR. I am afraid I miss something which needs it.
https://github.com/pingcap/tispark/pull/1623
https://github.com/pingcap/tispark/pull/2106

could you please take a look at this PR:

1. Is there anything I miss? Can I just reload without reloading the table cache?
2. Do you think we need a configuration to determine whether load tables cache?

birdstorm:
> @birdstorm It is a simple PR. 
> 
> I think we can just reload db cache when the schema version has changed. GetTable or ListTable method will find the table cache is empty, then just load all tables in its' own db rather than all dbs. These will reduce the pressure on users with a large number of tables.
> 
> I find you use relaod(true) which will reload all table scehmas in the following PR. I am afraid I miss something which needs it.
> https://github.com/pingcap/tispark/pull/1623
> https://github.com/pingcap/tispark/pull/2106
> 
> could you please take a look at this PR:
> 
> 1. Is there anything I miss? Can I just reload without reloading the table cache?
> 2. Do you think we need a configuration to determine whether load tables cache?

@shiyuhang0 have you tested the case where there is rapid change in the schema version? e.g., Many tables are created and deleted at the same time, while the queries randomly occur on those tables? We added the function to reload catalog on tables because when we tested previously, it is possible for some of the queries to return “table not exist“ error even if the tables exist.

birdstorm (approved):
LGTM

shiyuhang0:
> > @birdstorm It is a simple PR.
> > I think we can just reload db cache when the schema version has changed. GetTable or ListTable method will find the table cache is empty, then just load all tables in its' own db rather than all dbs. These will reduce the pressure on users with a large number of tables.
> > I find you use relaod(true) which will reload all table scehmas in the following PR. I am afraid I miss something which needs it.
> > #1623
> > #2106
> > could you please take a look at this PR:
> > 
> > 1. Is there anything I miss? Can I just reload without reloading the table cache?
> > 2. Do you think we need a configuration to determine whether load tables cache?
> 
> @shiyuhang0 have you tested the case where there is rapid change in the schema version? e.g., Many tables are created and deleted at the same time, while the queries randomly occur on those tables? We added the function to reload catalog on tables because when we tested previously, it is possible for some of the queries to return “table not exist“ error even if the tables exist.

So, these PRs is aiming to fix the table cache miss. But, I can not find where is the race in the code :(
I think it is necessary to have a test. Is the  “table not exist“ easy to reproduce? I need to know if the race will not happen anymore or  it is just not be reproduced

birdstorm:
> > > @birdstorm It is a simple PR.
> > > I think we can just reload db cache when the schema version has changed. GetTable or ListTable method will find the table cache is empty, then just load all tables in its' own db rather than all dbs. These will reduce the pressure on users with a large number of tables.
> > > I find you use relaod(true) which will reload all table scehmas in the following PR. I am afraid I miss something which needs it.
> > > #1623
> > > #2106
> > > could you please take a look at this PR:
> > > 
> > > 1. Is there anything I miss? Can I just reload without reloading the table cache?
> > > 2. Do you think we need a configuration to determine whether load tables cache?
> > 
> > @shiyuhang0 have you tested the case where there is rapid change in the schema version? e.g., Many tables are created and deleted at the same time, while the queries randomly occur on those tables? We added the function to reload catalog on tables because when we tested previously, it is possible for some of the queries to return “table not exist“ error even if the tables exist.
> 
> So, these PRs is aiming to fix the table cache miss. But, I can not find where is the race in the code :(
> I think it is necessary to have a test. Is the  “table not exist“ easy to reproduce? I need to know if the race will not happen anymore or  it is just not be reproduced

It should be easy to reproduce if the race still exists. After we have fixed the region cache issue, it is possible that the problem has been resolved since then.
In either case, I believe the case should be re-tested in this PR to ensure that.

zhangyangyu (approved):

xuanyu66 (approved):

PR #2761, opened 2023-08-25
Remove JDK 1.8 specific functions from tikv-client

frew:
Removes JDK 1.8 specific code.

Mostly the change doesn't affect functionality - we pull in javax.annotation-api from Maven which should work for both JDK 1.8 and later versions and remove unused code in MemoryUtil.

The one functionality change is removing MemoryUtil.free calls from TiBlockColumnVector and AutoGrowByteBuffer. This will defer buffer removal from the close() call to the finalize() call on the relevant object, but there doesn't appear to be a clean way independent of JDK versions to do this. I've tested some with Spark and it doesn't appear to have a major impact on memory usage.

### Check List 

Tests 

 Most changes are verified by compilation passing. Memory changes have been verified by manually running some Spark jobs.

Code changes

Some MemoryUtil methods have been removed, but that's intended to be internal.

Side effects

Possible that memory usage will change due to free() call removal.

Related changes

None of the above

### DIFF
--- a/tikv-client/pom.xml
+++ b/tikv-client/pom.xml
@@ -38,6 +38,11 @@
 
     <dependencies>
         <dependency>
+            <groupId>javax.annotation</groupId>
+            <artifactId>javax.annotation-api</artifactId>
+            <version>1.3.2</version>
+        </dependency>
+        <dependency>
             <groupId>org.antlr</groupId>
             <artifactId>antlr4-runtime</artifactId>
             <version>4.7.1</version>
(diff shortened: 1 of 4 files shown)


xuanyu66:
Currently, we don't have a plan to support JDK(>1.8), it's a huge work.

frew:
@xuanyu66 Apologies if I'm missing something, but it seems like this PR makes it work? I've at least compiled TiSpark on JDK 11 with this PR and used it to export >10 TB of data successfully. Am I missing a use case that requires the unsafe free code?

xuanyu66:
@frew Thanks for contributing. We will keep an eye on this and find time to do a thorough test

shiyuhang0:
tests are green
image

